PR #248, opened 2014-11-20
Extend content api test helpers to support options

pedrosmmoreira:
- Extend `#content_api_has_artefacts_with_a_tag` to take a options hash.
  This allows the interface to be flexible enough for a wider range of
  testing situations, i.e. artefacts in a specific format.

### DIFF
--- a/lib/gds_api/test_helpers/content_api.rb
+++ b/lib/gds_api/test_helpers/content_api.rb
@@ -181,7 +181,7 @@ module GdsApi
 
       def content_api_has_artefacts_with_a_tag(tag_type, slug, artefact_slugs=[], options={tag: {}, artefact: {}})
         if options.has_key?(:draft)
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/content_api.rb
```diff
@@ -179,12 +179,16 @@ def content_api_has_sorted_child_tags(tag_type, parent_slug_or_hash, sort_order,
...
+        if options.has_key?(:draft)
+          puts "Passing a key of :draft outside of the 'tag' options hash is being deprecated. Please use tag: { draft: bool }"
+        end
+
+        draft = options.fetch(:draft) { options[:tag][:draft] || false }
 
         body = plural_response_base.merge(
           "results" => artefact_slugs.map do |artefact_slug|
-            artefact_for_slug(artefact_slug)
+            artefact_for_slug(artefact_slug, options[:artefact])
```
alext: This breaks things if options doesn't include an artefact - it passes in nil, overriding the default of empty hash.

PR #264, opened 2015-01-09
Adapter for ContentRegister

tommyp:
This commit adds an API adapter for content register. A way of seeing what content item sits at a certain base_path. This follows the example of the Content Store. 

I'm opening this before adding tests because I want to discuss with someone what I should be testing against this class and also what test_helpers I should be providing. 

I get that it makes sense to provide a test helper with the Content Store to stub a request and return some data with the `content_store_has_item` helper but I'm not sure what I should be doing with this. Should I provide a helper to return `n` of a certain format? Or should it provide similar functionality to `content_store_has_item` in that it should allow you to stub an array of Content Registrations?

### DIFF
--- a/lib/gds_api/content_register.rb
+++ b/lib/gds_api/content_register.rb
@@ -1,0 +1,15 @@
+require_relative 'base'
+require_relative 'exceptions'
+
+class GdsApi::ContentRegister < GdsApi::Base
+
+  def entries(format)
+    get_json!(entries_url(format))
+  end
+
+  private
+
+  def entries_url(format)
+    "#{endpoint}/entries?format=#{format}"
+  end
+end
--- a/lib/gds_api/helpers.rb
+++ b/lib/gds_api/helpers.rb
@@ -2,6 +2,7 @@ require 'gds_api/asset_manager'
 require 'gds_api/business_support_api'
 require 'gds_api/collections_api'
 require 'gds_api/content_api'
+require 'gds_api/content_register'
 require 'gds_api/content_store'
 require 'gds_api/fact_cave'
 require 'gds_api/imminence'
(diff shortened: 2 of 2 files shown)


lib/gds_api/content_register.rb
```diff
@@ -0,0 +1,19 @@
+require_relative 'base'
+require_relative 'exceptions'
+
+class GdsApi::ContentRegister < GdsApi::Base
+
+  def entries(format)
+    get_json(entries_url(format))
+  end
+
+  def entries!(format)
```
alext: I don't think we'd want both variants here.  For all other adapters that access an index endpoint we just have a non-bang method that calls `get_json!`.  It should be an exceptional case for the endpoint to return a 404.

tommyp: Amended the commit to remove `entries!` and use `get_json!`.

alext:
Looks good.

Re: test helpers, it's probably worth looking at the imminence ones.  The best thing to do would be to start writing the tests you want in the client application without helpers. It should pretty quickly become clear what test helpers you need.

tommyp:
Re: test helpers - So after some chat in emails it appears I don't actually even need to use the ContentRegister, so I won't need to write any tests for this. Feel free to merge if you're happy to have an adapter without test helpers. 

¯\\\_(ツ)_/¯

lib/gds_api/helpers.rb
```diff
@@ -31,6 +32,10 @@ def content_api(options = {})
       @content_api ||= GdsApi::ContentApi.new(Plek.current.find("contentapi"), options)
     end
 
+    def content_register(options = {})
+      @content_register ||= GdsApi::ContentApi.new(Plek.current.find("content-register"), options)
```
jamiecobbett: I think you meant to write `ContentRegister` not `ContentApi` here.

jamiecobbett:
I agree with @alext about test helpers - they're not essential, and I would generally add them when I've written one in an app that needs it.

lib/gds_api/helpers.rb
tommyp: Whoops. Totally did. Fixed.

elliotcm:
I'm merging this as we now need the adapter for the elections work.

I'll be adding test helpers as needed for the tickets I'm doing.

PR #271, opened 2015-02-12
Bump to 17.5.0

tommyp:
Major version bump to get `n` times for Publishing API test helpers.


alext:
To be pedantic, this is a minor bump.  A major bump would be from 17.4.0 to 18.0.0 (it's major.minor.patch).

PR #273, opened 2015-02-12
Fix gem publishing by building gem under ruby 2.1

alext:
A recent update at rubygems.org results in errors when pushing gems built under 1.9.3 - something related to YAML serialisation.

This also includes a change to run the tests under several ruby versions.


jamiecobbett:
Possible candidate for the suspected cause in rubygems.org: https://github.com/rubygems/rubygems.org/pull/867

PR #276, opened 2015-02-18
Remove base_path from Publishing API request bodies

jamiecobbett:
Publishing API currently ignores this field and is about to start rejecting it.
It uses the version in the request URL instead.

Separating out the content item helpers into Publishing API and Content Store
also makes sense as these formats are not the same - for example the "links"
supplied are an array of content IDs when sent to Publishing API, but are
expanded out when fetched from Content Store.

For more context see: https://github.com/alphagov/content-store/pull/104


jamiecobbett:
Closing to start over.

PR #281, opened 2015-03-02
Support API: adapters for fetching problem reports, orgs with problem reports

benilovj:
These are endpoints added in alphagov/support-api#32


benilovj:
closing this for now because there aren't any clients consuming this information and the API may be changing soon.

PR #282, opened 2015-03-04
Publishing API responses no longer include the entity

jamiecobbett:
This updates the test helpers to match.

See: https://github.com/alphagov/content-store/pull/107

@heathd and I checked every app that uses Publishing API and weren't able to find any uses of the response body, let alone the entity.


heathd:
looks good to me

PR #284, opened 2015-03-06
Omit base_path from publishing_api request bodies

jamiecobbett:
This field is deprecated, soon to be rejected.

I couldn't see a clean alternative to duplicating the implementation.

This should be a major version bump as the API for the test helpers has changed in an incompatible way.

### DIFF
--- a/lib/gds_api/test_helpers/content_item_helpers.rb
+++ b/lib/gds_api/test_helpers/content_item_helpers.rb
@@ -10,7 +10,8 @@ module GdsApi
           "format" => "guide",
           "need_ids" => ["100001"],
           "public_updated_at" => "2014-05-06T12:01:00+00:00",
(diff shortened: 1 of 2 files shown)


vinayvinay:
@jamiecobbett: i'd prefer we continue including `ContentItemHelpers`, so that future modifications to content item JSON schema remain easy to incorporate:

- `ContentItemHelpers#content_item_for_base_path` would not return a `base_path` in the hash,
- `GdsApi::TestHelpers::PublishingApi` would include that behaviour as is, and
- `GdsApi::TestHelpers::ContentStore` would override that method:
```
def content_item_for_base_path(base_path)
  super.merge({ base_path: base_path })
end
```

jamiecobbett:
@vinayvinay top idea! Rebased.

PR #309, opened 2015-05-22
Fix a bug in the organisations_list test helper

boffbowsh:
Response wasn't being converted to JSON before being used in the stub


jamiecobbett:
Our styleguide says that version bumps happen in a commit and Pull Request of their own: https://github.com/alphagov/styleguides/blob/master/rubygems.md#versioning

boffbowsh:
Ah, sorry, will know for next time.

PR #312, opened 2015-06-04
Rewrite Response#method_missing

jackscotti:
If called passing parameters and/or a block, the method returned a
'wrong number of arguments' error.

It is an unwanted behaviour and this PR removes it.

### DIFF
--- a/lib/gds_api/response.rb
+++ b/lib/gds_api/response.rb
@@ -74,8 +74,8 @@ module GdsApi
       @ostruct ||= self.class.build_ostruct_recursively(to_hash)
     end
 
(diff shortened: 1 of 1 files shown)


lib/gds_api/response.rb
```diff
@@ -2,6 +2,7 @@
 require 'ostruct'
 require 'forwardable'
 require 'rack/cache'
+require 'pry'
```
issyl0: Did you mean to leave this line in here?

jackscotti: No :/

PR #317, opened 2015-06-23
Pass on authenticated_user header in API requests

danielroseman:
In the draft stack, requests to frontends will contain an `X_GOVUK_AUTHENTICATED_USER` header containing the authenticated user ID from signon. This needs to be passed on to any requests to content-store via gds-api-adapters in order to enable access limiting.

This PR modifies the existing GovukRequestIdSniffer to store multiple headers and pass them on via the json client.

https://trello.com/c/cxG0gj4N


lib/gds_api/govuk_headers.rb
```diff
@@ -0,0 +1,14 @@
+module GdsApi
+  class GovukHeaders
+    class << self
+      def headers
+        Thread.current[:headers] ||= {}
+      end
+
+      def values
```
tekin: I don't think this method name adequately communicate what the method returns. Maybe the interface for this class would be more user-friendly if it was something like this:

```ruby
GdsApi::GovukHeaders.set_header(header_name, value) # sets the header header_name with value
GdsApi::GovukHeaders.headers # returns the headers that have been set
```

i.e. an explicit setter method for setting a given header rather than accessing the thread-local variable directly and a `headers` method that returns the cleaned up headers that have actual values.

lib/gds_api/govuk_headers.rb
```diff
@@ -0,0 +1,14 @@
...
+        headers.select {|k, v| !(v.nil? || v.empty?) }
```
tekin: This can be shortened to `headers.select { |k,v| v.present? }`

danielroseman: I think `present?` is a Rails-specific method, the adapters need to work for non-Rails apps too.

README.md
```diff
@@ -42,18 +55,9 @@ the form:
...
-We set a unique header at the cache level called `GOVUK-Request-Id`. In order
-to serve a user's request, if apps make API requests they should pass on this
-header, so that we can trace a request across the entire GOV.UK stack.
-
-`GdsApi::GovukRequestIdSniffer` middleware takes care of this. This gem contains
-a railtie that configures this middleware for Rails apps without extra effort.
-Other Rack-based apps should opt-in by adding this line to your `config.ru`:
-
-```use GdsApi::GovukRequestIdSniffer```
+In addition, the adapters can pass on a header containing the authenticated user
```
tekin: I think it would be worth expanding on this a bit - it could be made clearer that these are two similar but separate forms of authorisation. i.e. the `bearer_token` method of authorisation is generally used to identify applications' API users to other endpoints, e.g. signon sends a bearer token for its "Signonotron API Client" user when it makes auth-related callbacks to apps to authenticate that it can make these requests. It may be that the opening paragraph above needs to be rewritten slightly to make the distinction between app-level identification with a bearer_token, and individual user identification with `HTTP_X_GOVUK_AUTHENTICATED_USER`.

lib/gds_api/govuk_headers.rb
bradleywright: I would expect this to just return the values side of the key/value equation, not the full hash. You know, like Python.

lib/gds_api/railtie.rb
```diff
@@ -1,10 +1,15 @@
-require_relative 'middleware/govuk_request_id_sniffer'
+require_relative 'middleware/govuk_header_sniffer'
 
 module GdsApi
   class Railtie < Rails::Railtie
     initializer "gds_api.initialize_govuk_request_id_sniffer" do |app|
-      Rails.logger.info "Using middleware GdsApi::GovukRequestIdSniffer to sniff for GOVUK-Request-Id header"
-      app.middleware.use GdsApi::GovukRequestIdSniffer
+      Rails.logger.info "Using middleware GdsApi::GovukHeaderSniffer to sniff for GOVUK-Request-Id header"
+      app.middleware.use GdsApi::GovukHeaderSniffer, 'HTTP_GOVUK_REQUEST_ID', 'GOVUK-Request-Id'
```
tekin: I don't think the capitalisation is correct here. Shouldn't it be `Govuk-Request-Id`? Maybe it would be safer if the conversion of the Rack-specification-formatted header name (`HTTP_VARIABLE_NAME`) to the version that the API adapter sets  should happen in `GovukHeaderSniffer` instead of being passed in. Looking at the way headers were being [merged in `JsonClient`](https://github.com/alphagov/gds-api-adapters/pull/317/files#diff-36a775ddd5ecd0c84b6e434ce0b174a1L175) before, it looks like it could simply be converted into a symbolic key, e.g. `header_name.sub(/^HTTP_/, "").downcase.to_sym`.

lib/gds_api/govuk_headers.rb
tekin: Good point!

lib/gds_api/govuk_headers.rb
tekin: Also like [Ruby](http://ruby-doc.org/core-2.2.2/Hash.html#method-i-values)

README.md
tekin: I've tweaked my comment above slightly. The real distinction is that the `bearer_token` is used for gds-sso-based **authentication**, and the authenticated user header is used to identify an authenticated user so that the app is able to perform additional **authorisation** if required.

danielroseman:
Thanks for the comments. I've reworked the interface as suggested, and tweaked the readme to make the authentication/authorisation distinction more obvious.

lib/gds_api/middleware/govuk_header_sniffer.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def call(env)
+      GdsApi::GovukHeaders.set_header(readable_name, env[@header_name])
+      @app.call(env)
+    end
+
+    private
+
+    def readable_name
+      @header_name.sub(/^HTTP_/, "").gsub(/_/, "-").downcase.to_sym
```
tekin: Is the `gsub(/_/, "-")` call required? Doesn't the subsequent call to `downcase` make it redundant?

tekin: Oh, ignore me, I'm being dim...

tekin: Actually, maybe it is unwanted... We want `"HTTP_GOVUK_REQUEST_ID"` to be transformed to `:govuk_request_id`, no?

PR #321, opened 2015-06-26
Don't cache cache-control "private" and "no-store" responses

tekin:
This updates gds-api-adapters to **not** cache responses that have cache-control "private" and "no-store" directives. It also updates `GdsApi::Response` with a `cache_control` method that exposes a `Rack::Cache::CacheControl` instance and, for further convenience, a  `cache_control_private?` method, making it easier for clients to know when a response is private and behave appropriately.

Trello: https://trello.com/c/oQvY66lk/239-better-cache-handling-in-gds-api-adapters

### DIFF
--- a/lib/gds_api/response.rb
+++ b/lib/gds_api/response.rb
@@ -46,9 +46,9 @@ module GdsApi
     end
 
     def expires_at
-      if headers[:date] && cache_control['max-age']
+      if headers[:date] && cache_control.max_age
         response_date = Time.parse(headers[:date])
-        response_date + cache_control['max-age'].to_i
+        response_date + cache_control.max_age
       elsif headers[:expires]
         Time.parse(headers[:expires])
       end
@@ -59,8 +59,8 @@ module GdsApi
 
       age = Time.now.utc - Time.parse(headers[:date])
 
-      if cache_control['max-age']
-        cache_control['max-age'].to_i - age.to_i
+      if cache_control.max_age
+        cache_control.max_age - age.to_i
       elsif headers[:expires]
         Time.parse(headers[:expires]).to_i - Time.now.utc.to_i
       end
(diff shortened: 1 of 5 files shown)


lib/gds_api/response.rb
```diff
@@ -46,24 +46,32 @@ def headers
...
+      elsif headers[:expires]
+        Time.parse(headers[:expires]).to_i - Time.now.utc.to_i
+      end
+    end
+
+    def cache_control
+      @cache_control ||= Rack::Cache::CacheControl.new(headers[:cache_control])
+    end
+
+    def cache_control_private?
```
danielroseman: CacheControl already defines a `private?` method.

tekin: Oh, so it does! Turns out it also has a `max_age` method, which makes it possible to simplify the code some more. I've adjusted the code accordingly - no more `cache_control_private?` method; clients can call `cache_control.private?` instead.

tekin:
Note I've made a further change here: I've adjusted the content-store test helpers so that the cache-control headers can be better controlled.

lib/gds_api/test_helpers/content_store.rb
```diff
@@ -9,10 +9,26 @@ module ContentStore
...
       CONTENT_STORE_ENDPOINT = Plek.current.find('content-store')
 
-      def content_store_has_item(base_path, body = content_item_for_base_path(base_path), expires_in = 900)
+      # Stubs a content item in the content store.
+      # The following options can be passed in:
+      #
+      #   :max_age  will set the max-age of the Cache-Control header in the response. Defaults to 900
+      #   :private  if true, the Cache-Control header will include the "private" directive. By default it
+      #             will include "public"
+      def content_store_has_item(base_path, body = content_item_for_base_path(base_path), options = {})
```
alext: This is a breaking change, so technically should require a major version bump.  (Good timing because they's already another breaking change lined up - #322).

tekin: Yep, as noted in the commit message.

PR #325, opened 2015-07-02
Update User-Agent header

elliotcm:
Previously included the version, now includes the
version and the originating app name.

The app name ENV var is set by Puppet for all
applications.

https://trello.com/c/zVGpH4oP/246-add-user-agent-tracking-to-gds-api-adapters

### DIFF
--- a/test/json_client_test.rb
+++ b/test/json_client_test.rb
@@ -625,7 +625,7 @@ class JsonClientTest < MiniTest::Spec
 
   def test_client_can_use_bearer_token
     client = GdsApi::JsonClient.new(bearer_token: 'SOME_BEARER_TOKEN')
-    expected_headers = GdsApi::JsonClient::DEFAULT_REQUEST_HEADERS.
+    expected_headers = GdsApi::JsonClient.default_request_headers.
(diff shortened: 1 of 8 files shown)


test/json_client_test.rb
```diff
@@ -789,4 +789,20 @@ def test_should_raise_error_if_attempting_to_disable_timeout
...
+    url = "http://some.other.endpoint/some.json"
+    stub_request(:get, url).to_return(:status => 200)
+
+    GdsApi::JsonClient.new.get_json(url)
+
+    assert_requested(:get, %r{/some.json}) do |request|
+      request.headers["User-Agent"] == "gds-api-adapters/#{GdsApi::VERSION} (api-tests)"
+    end
+
+    ENV['GOVUK_APP_NAME'] = previous_govuk_app_name
```
alext: If this test fails, this value won't get reset, and therefore could trigger a failure in other tests.  It would be good to reset this in an `ensure` clause just to be sure.

alext:
Minor comment about the tests, otherwise looks good to me.

test/json_client_test.rb
elliotcm: Good call, updated.

PR #326, opened 2015-07-02
Release 20.1.0

elliotcm:
Releases https://github.com/alphagov/gds-api-adapters/pull/325


alext:
I'm not sure this classes as a breaking change...  I don't think it's reasonable to consider the format of the User-Agent string as part of the public API.

elliotcm:
@alext I was iffy but was told "if in doubt".  If you're not concerned I'll switch it to a minor.

PR #331, opened 2015-07-09
Remove support for deprecated contentapi path

jamiecobbett:
This has been deprecated since:
https://github.com/alphagov/govuk_content_api/commit/5361729cd9adb4fe8ba4d94d6e428b6fc8fc8465

Using grep and looking at the user agent in the logs for contentapi in
production, the only place I was able to find this being used was in Whitehall
in the code for building breadcrumbs. We should be able to switch this to
specify the tag_type as it will always be looking for tags of type "section".

### DIFF
--- a/lib/gds_api/content_api.rb
+++ b/lib/gds_api/content_api.rb
@@ -52,13 +52,12 @@ class GdsApi::ContentApi < GdsApi::Base
   end
 
   def tag(tag, tag_type=nil)
-    url = "#{base_url}/tags"
-
-    if tag_type
-      url << "/#{CGI.escape(tag_type)}"
+    if tag_type.nil?
+      raise "Requests for tags without a tag_type are no longer supported. You probably want a tag_type of 'section'. See https://github.com/alphagov/govuk_content_api/blob/f4c0102a1ae4970be6a440707b89798442f768b9/govuk_content_api.rb#L241-L250"
     end
 
-    get_json("#{url}/#{CGI.escape(tag)}.json")
+    url = [base_url, "tags", CGI.escape(tag_type), CGI.escape(tag)].join("/") + ".json"
+    get_json(url)
   end
 
   def with_tag(tag, tag_type=nil, options={})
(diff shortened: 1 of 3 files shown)


lib/gds_api/content_api.rb
```diff
@@ -52,13 +52,12 @@ def child_tags(tag_type, parent_tag, options={})
...
-
-    if tag_type
-      url << "/#{CGI.escape(tag_type)}"
+    if tag_type.nil?
+      raise "Requests for tags without a tag_type are no longer supported. You probably want a tag_type of 'section'. See https://github.com/alphagov/govuk_content_api/blob/f4c0102a1ae4970be6a440707b89798442f768b9/govuk_content_api.rb#L241-L250"
     end
 
-    get_json("#{url}/#{CGI.escape(tag)}.json")
+    url = [base_url, "tags", CGI.escape(tag_type), CGI.escape(tag)].join("/") + ".json"
+    get_json("#{url}")
```
boffbowsh: `get_json(url)`

lib/gds_api/content_api.rb
```diff
@@ -52,13 +52,12 @@ def child_tags(tag_type, parent_tag, options={})
   end
 
   def tag(tag, tag_type=nil)
-    url = "#{base_url}/tags"
-
-    if tag_type
-      url << "/#{CGI.escape(tag_type)}"
+    if tag_type.nil?
```
boffbowsh: I do wonder if it's worth just defaulting the type to `"section"`, as that's the current behaviour.

alext: Not strictly true.  The current behaviour is to match a tag of any type with the given slug.  If 2 tags of different types have the same slug, one of them will be returned. (I think we relaxed the uniqueness constraint to allow tags of different types to have matching slugs)

alext: Sorry, I'm wrong there.  I misread the code in contentapi.

alext: In which case it may be worth defaulting to section, but outputting a deprecation warning (eg https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/router.rb#L48-L51)

lib/gds_api/content_api.rb
jamiecobbett: "Rebased"

lib/gds_api/content_api.rb
jamiecobbett: I felt that it wasn't worth the extra back-and-forth of emitting a deprecation warning, especially if I'm correct about the only client being Whitehall, which I'm migrating.

boffbowsh:
Fine by me :+1:

PR #337, opened 2015-07-30
Generalise request matching in PublishingApi assertions

heathd:
The assertions provided by the PublishingApi test helpers are useful.
The existing implementation allows an assertion to require that a
request was made to a publishing api endpoint with a hash containing a
certain list of required attributes.

However the underlying matching process would only take into account the
outermost level of the hash when performing the comparison. This means
that you could make an flexible assertion about the required elements at
the top level of a document sent to the publishing api, but if you
wanted to make assertions about the details section or elements nested
within the details section, you could only perform an exact match on the
details hash.

This would lead to brittle, difficult to maintain tests.

This commit allows passing custom matcher predicates to the assertions
and defines a new matcher which can match nested structures more
loosely.

The original strict matching behaviour is retained as a default to avoid
possible unintended side-effects on the test suites of other apps.

We considered making the looser behaviour the default, but after
reviewing the usage of some other apps I had enough doubt about that I
thought it would be better to be conservative and avoid changing the
default matching behaviour.

To use the assertions with the looser matcher you can do the following:

```
assert_publishing_api_put_item(
  base_path,
  request_json_including(details: {title: "My title"})
)
```

Note that both matchers will convert symbols to strings so the
hash can use either symbol or string keys.

## Feedback suggestions

* Do the predicate names make sense? (`request_json_matching`, `request_json_strictly_matching`)
  * names changed
* [x] Do you agree with the conservative approach?

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api.rb
+++ b/lib/gds_api/test_helpers/publishing_api.rb
@@ -43,36 +43,71 @@ module GdsApi
         stub_request(:put, %r{\A#{PUBLISHING_API_ENDPOINT}/publish-intent})
       end
 
-      def assert_publishing_api_put_item(base_path, attributes = {}, times = 1)
+      def assert_publishing_api_put_item(base_path, attributes_or_matcher = {}, times = 1)
         url = PUBLISHING_API_ENDPOINT + "/content" + base_path
-        assert_publishing_api_put(url, attributes, times)
+        assert_publishing_api_put(url, attributes_or_matcher, times)
       end
 
-      def assert_publishing_api_put_draft_item(base_path, attributes = {}, times = 1)
+      def assert_publishing_api_put_draft_item(base_path, attributes_or_matcher = {}, times = 1)
         url = PUBLISHING_API_ENDPOINT + "/draft-content" + base_path
-        assert_publishing_api_put(url, attributes, times)
+        assert_publishing_api_put(url, attributes_or_matcher, times)
       end
 
-      def assert_publishing_api_put_intent(base_path, attributes = {}, times = 1)
+      def assert_publishing_api_put_intent(base_path, attributes_or_matcher = {}, times = 1)
         url = PUBLISHING_API_ENDPOINT + "/publish-intent" + base_path
-        assert_publishing_api_put(url, attributes, times)
+        assert_publishing_api_put(url, attributes_or_matcher, times)
       end
 
-      def assert_publishing_api_put(url, attributes = {}, times = 1)
-        if attributes.empty?
-          assert_requested(:put, url, times: times)
+      def assert_publishing_api_put(url, attributes_or_matcher = {}, times = 1)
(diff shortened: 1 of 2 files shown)


lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -43,37 +43,76 @@ def stub_default_publishing_api_put_intent()
...
         stub_request(:any, /#{PUBLISHING_API_ENDPOINT}\/.*/).to_return(:status => 503)
       end
+
+    private
+      def values_match_recursively(expected_value, actual_value)
+        case expected_value
+        when Hash
+          return false unless actual_value.is_a?(Hash)
+          expected_value.all? do |expected_sub_key, expected_sub_value|
+            key = symbols_as_strings(expected_sub_key)
```
elliotcm: Why not just call `expected_sub_key.to_s`?

test/test_helpers/publishing_api_test.rb
```diff
@@ -0,0 +1,159 @@
...
+      let(:matcher) { request_json_matching({"a" => [{"b" => 1}, 2]}) }
+
+      it "matches a body with exact same inner array" do
+        assert matcher.call(stub("request", body: '{"a": [{"b": 1}, 2]}'))
+      end
+
+      it "matches a body with an inner hash with extra elements" do
+        assert matcher.call(stub("request", body: '{"a": [{"b": 1, "c": 3}, 2]}'))
+      end
+
```
elliotcm: Whitespace.

elliotcm:
General approach seems fine.  Specifics seem mostly fine.

Regarding your questions, I'd personally assume that `request_json_matching` performed an exact match rather than a partial match.  Should we follow the RSpec format where they use `hash_including`?

I'm a little unsure about extending the partial matching to hashes within arrays.  Do we have a concrete use case for that?

lib/gds_api/test_helpers/publishing_api.rb
heathd: keys could in theory be integers or booleans

elliotcm: OK

heathd:
thanks for your comments

re: naming, how about

* change `request_json_strictly_matching` to `request_json_matching` 
* change `request_json_matching` to `request_json_including`

re: use case for matching hashes within arrays, yes that's needed for matching the manuals table of contents which look like this:

```json
    details: {
      child_section_groups: [
        {
          title: "Contents",
          child_sections: [
            {
              title: "Child title",
              description: "Child description",
              base_path: "/guidance/manual/child-title",
            }
          ]
        }
      ]
    }
```

lib/gds_api/test_helpers/publishing_api.rb
heathd: just read the JSON spec, I was mistaken. Object keys must be strings.

elliotcm:
Name change sounds good to me.

PR #340, opened 2015-08-11
Add special route publisher under PublishingApi.

elliotcm:
This will be used in several apps for registering
"special" routes like `/government` or `/robots.txt`

See https://trello.com/c/blLdEZN5/292-make-apps-register-special-routes-on-deploy

### DIFF
--- a/lib/gds_api/publishing_api/special_route_publisher.rb
+++ b/lib/gds_api/publishing_api/special_route_publisher.rb
@@ -1,0 +1,36 @@
+require "gds_api/publishing_api"
+
+module GdsApi
+  class PublishingApi < GdsApi::Base
+    class SpecialRoutePublisher
+      def initialize(options = {})
+        @logger = options[:logger] || GdsApi::Base.logger
+        @publishing_api = options[:publishing_api] || GdsApi::PublishingApi.new(Plek.find("publishing-api"))
+      end
+
+      def publish(options)
+        logger.info("Publishing #{options.fetch(:type)} route #{options.fetch(:base_path)}, routing to #{options.fetch(:rendering_app)}")
+
+        publishing_api.put_content_item(options.fetch(:base_path), {
+          content_id: options.fetch(:content_id),
+          format: "special_route",
+          title: options.fetch(:title),
+          description: options[:description] || "",
+          routes: [
+            {
+              path: options.fetch(:base_path),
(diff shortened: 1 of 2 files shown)


test/publishing_api/special_route_publisher_test.rb
```diff
@@ -0,0 +1,46 @@
...
+          routes: [
+            {
+              path: "/favicon.ico",
+              type: "exact",
+            }
+          ],
+          publishing_app: "static-publisher",
+          rendering_app: "static-frontend",
+          update_type: "major",
+          public_updated_at: Time.now.iso8601,
```
heathd: i wonder if it's worth allowing the `public_updated_at` timestamp to be passed in as well. Let it default to 'now' but let it be overridden as well?

elliotcm: We've so far not had a use for that, and I can't see a use for it in the work we're doing.

heathd:
looks great, just one small suggestion.

I was unable to view the test failure due to my dns being broken at the moment.

elliotcm:
@heathd the test failure was due to ruby versioning, it's now passing.

lib/gds_api/publishing_api/special_route_publisher.rb
```diff
@@ -0,0 +1,36 @@
...
+          routes: [
+            {
+              path: options.fetch(:base_path),
+              type: options.fetch(:type),
+            }
+          ],
+          publishing_app: options.fetch(:publishing_app),
+          rendering_app: options.fetch(:rendering_app),
+          update_type: "major",
+          public_updated_at: (Time.respond_to?(:zone) ? Time.zone.try(:now) : Time.now).iso8601,
```
alext: Why the `Time.zone.try(:now)` here.  That could in theory return nil, which would be invalid.

elliotcm: I intended to default it to `Time.now` if that was the case but didn't.

test/publishing_api/special_route_publisher_test.rb
alext: This line has the potential to cause random test failures if a second ticks over between the assertion and the actual call below.  We should use Timecop to freeze the time to avoid this.

PR #342, opened 2015-08-12
Fix `helpers.rb` after finder api removal

heathd:
The removal of `finder_api` in 2715160c539 did not remove the `require`
line from `helpers.rb`, so the file raises an error when required.

There were no tests which load this file, so this error was not picked
up.

I've removed the offending require line and added a very simple test of
the `helpers.rb` so that such errors are picked up in the future.


alext:
:+1:

I do wonder whether we should deprecate the `Helpers` module entirely.

PR #344, opened 2015-08-13
Fix timezone handling

heathd:
if `Time.zone` returned `nil`, which unexpectedly does happen in some
circumstances, then the registration would crash.

This makes explicit the expected behaviour.

I've also changed the tests to use `Timecop` to freeze the system clock to
avoid possible intermit false failures in the tests (due to the clock
happening to roll over the 1 second boundary between the setting of the
expectation and the publishing of the special route).

### DIFF
--- a/lib/gds_api/publishing_api/special_route_publisher.rb
+++ b/lib/gds_api/publishing_api/special_route_publisher.rb
@@ -1,4 +1,5 @@
 require "gds_api/publishing_api"
+require "time"
 
 module GdsApi
   class PublishingApi < GdsApi::Base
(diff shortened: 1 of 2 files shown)


lib/gds_api/publishing_api/special_route_publisher.rb
```diff
@@ -25,10 +26,14 @@ def publish(options)
...
           rendering_app: options.fetch(:rendering_app),
           update_type: "major",
-          public_updated_at: (Time.respond_to?(:zone) ? Time.zone.try(:now) : Time.now).iso8601,
+          public_updated_at: time.now.iso8601,
         })
       end
 
+      def time
+        (Time.respond_to?(:zone) && Time.zone) || Time
+      end
```
alext: Nitpick: this should probably be a private method.

alext:
Minor comment, otherwise this looks good to me.

lib/gds_api/publishing_api/special_route_publisher.rb
heathd: :+1:

heathd:
@alext ok addressed your comment, thanks

PR #348, opened 2015-08-24
`content_register_isnt_available` test helper

mikejustdoit:
Stubs requests to Content Register, returning 503.

Entirely copied from the `publishing_api_isnt_available` test helper (https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api.rb#L89-L91).

### DIFF
--- a/lib/gds_api/test_helpers/content_register.rb
+++ b/lib/gds_api/test_helpers/content_register.rb
@@ -19,6 +19,11 @@ module GdsApi
           to_return(body: entries.to_json, status: 200)
       end
 
+      def content_register_isnt_available
+        stub_request(:any, /#{CONTENT_REGISTER_ENDPOINT}\/.*/).
+          to_return(status: 503)
+      end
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/content_register.rb
```diff
@@ -19,6 +19,11 @@ def stub_content_register_entries(format, entries)
           to_return(body: entries.to_json, status: 200)
       end
 
+      def content_register_isnt_available
+        stub_request(:any, /#{CONTENT_REGISTER_ENDPOINT}\/.*/)
+          .to_return(status: 503)
```
jennyd: The other methods in this module are split across lines with the `.` at the end of the first line - could you make this one consistent with that style?

mikejustdoit: Fixed, @jennyd

PR #350, opened 2015-08-28
Add a special route unpublisher

tommyp:
As we have special routes on GOV.UK published with the
`SpecialRoutePublisher`, we may also want to unpublish those special
routes.

This commit adds a `SpecialRouteUnpublisher` which publishes a `gone`
route for a `base_path`. This follows the same logic as the
`SpecialRoutePublisher`.

### DIFF
--- a/lib/gds_api/publishing_api/special_route_unpublisher.rb
+++ b/lib/gds_api/publishing_api/special_route_unpublisher.rb
@@ -1,0 +1,33 @@
+require "gds_api/publishing_api"
+
+module GdsApi
+  class PublishingApi < GdsApi::Base
+    class SpecialRouteUnpublisher
+      def initialize(options = {})
+        @logger = options[:logger] || GdsApi::Base.logger
+        @publishing_api = options[:publishing_api] || GdsApi::PublishingApi.new(Plek.find("publishing-api"))
+      end
+
+      def unpublish(options)
+        logger.info("Unpublishing #{options.fetch(:type)} route #{options.fetch(:base_path)}")
+
+        publishing_api.put_content_item(options.fetch(:base_path), {
+          content_id: options.fetch(:content_id),
+          format: "gone",
+          update_type: "major",
+          publishing_app: options.fetch(:publishing_app),
+          routes: [
+            {
+              path: options.fetch(:base_path),
+              type: options.fetch(:type),
+            }
+          ],
+        })
+      end
+
+    private
+      attr_reader :logger, :publishing_api
+
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


elliotcm:
Is this going to be run on every deploy like the publishings?

lib/gds_api/publishing_api/special_route_unpublisher.rb
```diff
@@ -0,0 +1,33 @@
...
+module GdsApi
+  class PublishingApi < GdsApi::Base
+    class SpecialRouteUnpublisher
+      def initialize(options = {})
+        @logger = options[:logger] || GdsApi::Base.logger
+        @publishing_api = options[:publishing_api] || GdsApi::PublishingApi.new(Plek.find("publishing-api"))
+      end
+
+      def unpublish(options)
+        logger.info("Unublishing #{options.fetch(:type)} route #{options.fetch(:base_path)}")
```
elliotcm: Typo.

elliotcm:
Is there enough shared with the publisher to make it one class?

tommyp:
@elliotcm: 
1. It will in Manuals Frontend, but once it's deployed I'll remove it from the App. 
2. The only thing really shared between them is `initialize`. The bulk of the difference is within the hash differing the two types of routes. I can refactor these classes to share the similar code between them if that is a blocker.

elliotcm:
@tommyp I'm just a little wary of the adapters getting bigger.  It's probably fine.  @alext could you give it your blessing?

alext:
I'm not sure I understand the need for this in API adapters. Is this something we're doing a lot? (if so we should probably discuss why this is happening...).

Secondly, there's nothing here that's specific to special routes, it's merely a `gone` item publisher, so I think the naming isn't right.

In general, replacing an item with a gone item is something that we should only be doing rarely, so I'd be reluctant to add this to api adapters.

tommyp:
The reason for this is that Manuals Frontend currently [publishes a special route](https://github.com/alphagov/manuals-frontend/blob/master/lib/tasks/publishing_api.rake#L41-L56) for a placeholder Manual which has since cease to be needed. I was thinking I would add this, update Manuals Frontend and then remove the publishing of the special route once it's done. 

I was following the `SpecialRoutePublisher`. Alternative naming suggestions welcome.

Totally agree, but I'm not sure where the removal of this should be happening. Should I build the `gone` content item myself and then remove it when it's been published?

tommyp:
After a chat with @alext I'm going to close this. Publishing a gone content item for Special Routes isn't something we should be doing regularly so instead of adding this I'm going to do it directly from the App involved.

PR #355, opened 2015-09-17
Set both kinds of timeout in RestClient

boffbowsh:
Setting `timeout` only sets the read timeout, which is the period
between a request being issued and receiving the response. We’ve had
issues with licensify in staging and new production where connecting to
the remote host waits for over a minute before erroring, so setting the
connection timeout will allow the request to fail faster.

The default timeout is 4 seconds, which seems reasonable for the
initial SYN/ACK cycle for a remote host.

In version 2.0.0 of RestClient (currently unreleased), both of these
timeouts will be set via the `timeout` variable, but can also be
individually set with `read_timeout` and `open_timeout`. Therefore,
setting `timeout` and `open_timeout` here is forward-compatible. The
change can be reverted once we upgrade to RestClient 2.0.0.

/cc @mattbostock @jennyd

### DIFF
--- a/test/json_client_test.rb
+++ b/test/json_client_test.rb
@@ -33,6 +33,14 @@ class JsonClientTest < MiniTest::Spec
     end
   end
 
+  def test_long_connections_timeout
+    url = "http://www.example.com/timeout.json"
+    stub_request(:get, url).to_raise(Net::OpenTimeout)
+    assert_raises GdsApi::TimedOutException do
+      @client.get_json(url)
+    end
+  end
+
   def test_get_should_raise_endpoint_not_found_if_connection_refused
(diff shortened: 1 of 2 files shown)


test/json_client_test.rb
```diff
@@ -33,6 +33,14 @@ def test_long_get_requests_timeout
     end
   end
 
+  def test_long_connections_timeout
+    url = "http://www.example.com/timeout.json"
+    stub_request(:get, url).to_raise(Net::OpenTimeout)
```
alext: I don't think `Net::OpenTimeout` exists in ruby 1.9.3 (hence the test failures).

boffbowsh: Fixed.

PR #357, opened 2015-09-17
[FOR DISCUSSION:] Log API call times to Statsd

mattbostock:
Log the duration of API calls to statsd for each application, using
their hostname to distinguish between each application.

Although we already track request times in Nginx, it's hard to see where
those API calls originated from.

By logging call times in this way, we'll be able to see:

- call times by machine they originated from
- call times by application they originated from
- call times by application the API call is targetting

By combining these three dimensions, we should be able to build a useful
picture for troubleshooting misbehaving APIs, networking partitions and
back-pressure.

__Note: I haven't tested this yet as I'd like feedback on the implementation and as to how useful this is__.

### DIFF
--- a/gds-api-adapters.gemspec
+++ b/gds-api-adapters.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'plek'
   s.add_dependency 'rack-cache'
   s.add_dependency 'rest-client', '~> 1.8.0'
+  s.add_dependency 'statsd-ruby', '1.2.1'
 
   s.add_development_dependency "minitest", "~> 3.4.0"
   s.add_development_dependency 'gem_publisher', '~> 1.5.0'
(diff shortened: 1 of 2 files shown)


boffbowsh:
Might want the ability to turn this off, or not have it configured by default. AFAIK this gem is supposed to work for non-GDS clients of our APIs, and they may not want these calls polluting their graphites. Perhaps use the presence or not of `GOVUK_STATSD_PREFIX` as a switch?

boffbowsh:
:+1: for intent

mattbostock:
@tijmenb Good call, I'll amend.

mattbostock:
@boffbowsh: You're right, we should turn this off by default. I'm not sure that using the prefix variable is the best way as it gives that variable unexpected meaning... I'll have a think.

bradwright:
A nice way might be to use dependency injection to inject a stats class, which by default is a noop. It would mean we'd need to update all the apps, but it'd be more explicit.

tijmenb:
I don't know much about these things, but would it be nicer to make `gds-api-adapters` emit [ActiveSupport Instrumentation events](http://edgeguides.rubyonrails.org/active_support_instrumentation.html), and let the apps itself forward those to statsd?

mattbostock:
Closing this while I think about reworking it.

PR #360, opened 2015-09-24
[publishing-api] contract and endpoints for V2 content API

alext:
Add contract definition, and api endpoints for the new V2 content API
(for sending content without links).

Trello: https://trello.com/c/X02OtoYt

### DIFF
--- a/test/publishing_api_v2_test.rb
+++ b/test/publishing_api_v2_test.rb
@@ -30,7 +30,7 @@ describe GdsApi::PublishingApiV2 do
       content_item = content_item_for_content_id(@content_id)
 
       publishing_api
-        .given("both content stores and url-arbiter empty")
+        .given("both content stores and the url-arbiter are empty")
         .upon_receiving("a request to create a content item without links")
         .with(
           method: :put,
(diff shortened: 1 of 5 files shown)


tijmenb:
Wouldn't it be preferable to have a `GdsApi::PublishingApiV2` or something like that? 

I'm afraid I will forget what versions `put_content` or `put_content_item` refer to.

alext:
> Wouldn't it be preferable to have a GdsApi::PublishingApiV2 or something like that?

There's been some back and forth on this.  It's probably worth having a discussion about this when people are back in next week.

alext:
After discussion, I've updated this to use `GdsApi::PublishingApiV2`.

test/publishing_api_v2_test.rb
```diff
@@ -0,0 +1,170 @@
...
+      response = @api_client.put_content(@content_id, content_item)
+      assert_equal 200, response.code
+    end
+
+    it "responds with 409 Conflict if the path is reserved by a different app" do
+      content_item = content_item_for_content_id(@content_id, "base_path" => "/test-item", "publishing_app" => "whitehall")
+
+      publishing_api
+        .given("/test-item has been reserved in url-arbiter by the publisher application")
+        .upon_receiving("a request to create a content item at /test-item")
```
danielroseman: would be clearer to say "a request from whitehall to create..."

alext: Good call, updated.

danielroseman:
:+1:  although I suppose we can't merge this until we have the matching endpoints (or we set up pact-broker).

elliotcm:
Bunch of extraneous newlines in this.

PR #364, opened 2015-10-09
[DISCUSS] Enable doc generation

heathd:
Pact can auto-generate documentation in markdown format. I had thought that this was only possible via the pact-broker, but it can also be done when running tests.

This might be useful?


alext:
Closing this because this will be handled by pact-broker.

PR #365, opened 2015-10-12
GET/PUT links endpoints

elliotcm:
Adds the helper methods and pact tests for the GET and PUT publishing API endpoints for managing links independently of content items.

Part of https://trello.com/c/SPLM0ANz/318-support-links-endpoints-in-publishing-api and related to https://github.com/alphagov/publishing-api/pull/86

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -20,6 +20,11 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     get_json(links_url(content_id))
   end
 
+  def put_links(content_id, payload)
+    links = payload.fetch(:links)
+    put_json!(links_url(content_id), links: links)
+  end
+
 private
 
   def content_url(content_id)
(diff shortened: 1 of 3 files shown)


benlovell:
Ruby 1.9.3 doesn't like kwargs.

lib/gds_api/publishing_api_v2.rb
```diff
@@ -16,9 +16,21 @@ def publish(content_id, update_type)
     })
   end
 
-  private
+  def get_links(content_id)
+    get_json(links_url(content_id))
+  end
+
+  def put_links(content_id, links:)
```
danielroseman: The gem still needs to support Ruby 1.9 so we can't use this syntax.

test/publishing_api_v2_test.rb
```diff
@@ -334,4 +334,230 @@ def content_item_for_content_id(content_id, attrs = {})
       assert_equal "Cannot publish an already published content item", error.error_details["error"]["message"]
     end
   end
+
+  describe "#get_links" do
+    describe "when there's a links entry with links" do
```
danielroseman: :+1:, I like this split of pact setup into before and verification in the actual spec. For the sake of consistency, we should probably move the other tests in this file to this pattern.

danielroseman:
:+1: apart from one issue.

elliotcm:
@benlovell @danielroseman Updated.

elliotcm:
@danielroseman Tests refactored.

PR #366, opened 2015-10-13
Remove business_support_schemes from Imminence API

h-lame:
Imminence hasn't supported this API since https://github.com/alphagov/imminence/pull/88
so it makes sense to finally remove this API from the adapters.


h-lame:
Did a search and it looks like only business-support-finder still used the test helpers so there's a PR to stop it using them. https://github.com/alphagov/business-support-finder/pull/107

h-lame:
Searching doesn't reveal any repos that still call the actual API method either.  I may not have access to all repos, so someone else may need to confirm this.

mikejustdoit:
"0 additions" :+1:

PR #371, opened 2015-10-21
Publish pacts to the pact broker.

alext:
This adds hooks to publish the generated pacts to the pact-broker after
the test run.

If `PACT_TARGET_BRANCH` is specified, it'll push the pacts to that target.
This will be used to push to branch targets (master, or branch-foo) so
that this project can then trigger builds of other projects using these
pacts before marking the build as green.

This additionally adds a hook to publish the pacts corresponding to the
released version when the gem is published.

**Note:** This depends on https://github.com/alphagov/ci-puppet/pull/343, and also CI being configured with necessary env variables, so **DO NOT MERGE** until those are in place.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,5 @@
 # -*- encoding: utf-8 -*-
 
-require "bundler/gem_tasks"
 require "rdoc/task"
 require 'rake/testtask'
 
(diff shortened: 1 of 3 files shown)


alext:
This has successfully published the pact to the pact-broker - https://pact-broker.dev.publishing.service.gov.uk/pacts/provider/Publishing%20API/consumer/GDS%20API%20Adapters/versions/branch-publish_pacts

alext:
CI has now been configured, so this is good to merge.

PR #373, opened 2015-10-22
Add the capability to publish a specified locale

tuzz:
Do not merge this until this has been merged: https://github.com/alphagov/publishing-api/pull/99

Please see the other pull request for more information.

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -10,10 +10,11 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     get_json(content_url(content_id))
   end
 
-  def publish(content_id, update_type)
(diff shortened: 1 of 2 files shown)


tijmenb:
Would it be possible to use a hash for the arguments? That will make it clear what the second and third arguments are.

Something like:


```ruby
response = @api_client.publish(@content_id, update_type: "major", locale: "fr")
```

tuzz:
Hi @tijmenb - we've changed it so that optional arguments can be specified with a hash. The `update_type` is mandatory, so we kept it as a mandatory argument to the method for now. We're also not using named params yet because this gem still supports pre Ruby 2 applications.

Are you happy with this change?

tijmenb:
Yep, awesome! :+1:

PR #376, opened 2015-10-26
Mapit raises error on 404

tommyp:
Previously, the Mapit adapter was only raising the error when the API
return a 400. This commit updates it to also return on a 404, which
happens when a User enters a valid postcode but gets back no matches.
Most commonly, this happens with new builds.


tommyp:
This change wasn't doing what I thought it was doing. Closing in favour of another PR.

PR #377, opened 2015-10-26
Remove mention of url-arbiter

steventux:
Url arbiter has been removed from publishing API, update provider state
titles accordingly.

### DIFF
--- a/test/publishing_api_test.rb
+++ b/test/publishing_api_test.rb
@@ -17,7 +17,7 @@ describe GdsApi::PublishingApi do
       content_item = content_item_for_publishing_api(base_path).merge("update_type" => "major")
 
       publishing_api
-        .given("both content stores are empty")
+        .given("no content exists")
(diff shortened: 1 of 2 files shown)


alext:
I wonder whether we should reconsider the wording of these states so as to avoid having to make similar changes to them in the future.

Ideally we would phrase these in a way that doesn't mention any implementation details at all - ie it would be good if they didn't reference the content-store at all.

steventux:
@alext I agree, I think @tuzz also made this point that we should remove mention of downstream applications. I'll add this to the PR.

steventux:
@alext see https://github.com/alphagov/gds-api-adapters/commit/8aae1580a877fe21664fdba9e533e189edae4750 for rephrased descriptions

alext:
https://github.com/alphagov/publishing-api/pull/118 will fix the tests on this end without breaking the tests in publishing-api.

PR #378, opened 2015-10-27
Mapit raises error on 404

tommyp:
Previously, the Mapit adapter was only raising the error when the API
return a 400. This commit updates it to also return on a 404, which
happens when a User enters a valid postcode but gets back no matches.
Most commonly, this happens with new build housing developments which
haven't been added to the MapIt DB.

### DIFF
--- a/lib/gds_api/mapit.rb
+++ b/lib/gds_api/mapit.rb
@@ -4,11 +4,15 @@ require_relative 'exceptions'
 class GdsApi::Mapit < GdsApi::Base
 
   def location_for_postcode(postcode)
-    response = get_json("#{base_url}/postcode/#{CGI.escape postcode}.json")
-    return Location.new(response) unless response.nil?
+    response = get_json!("#{base_url}/postcode/#{CGI.escape postcode}.json")
+    Location.new(response)
+
+  rescue GdsApi::HTTPNotFound => e
+    # allow 404 errors, as these will be valid postcodes with no match in Mapit
+    e
   rescue GdsApi::HTTPErrorResponse => e
     # allow 400 errors, as they can be invalid postcodes people have entered
-    raise e unless e.code == 400
+    e
   end
 
   def areas_for_type(type)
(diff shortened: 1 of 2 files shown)


lib/gds_api/mapit.rb
```diff
@@ -4,11 +4,13 @@
 class GdsApi::Mapit < GdsApi::Base
 
   def location_for_postcode(postcode)
```
benlovell: All of these explicit returns are superfluous, given the scoping.

tommyp: Removed the extra returns.

alext:
This is a breaking change. Currently this returns nil if nothing is found for a postcode, and apps deal with that.  This is also consistent with similar endpoints in other clients in this gem.

What's the problem with the current behaviour?  Can you not just check for `nil`?

tommyp:
Checking for `nil` doesn't tell us if the Postcode entered is invalid and would return a 400 (Partial: `WC2B` or nonsense: `FOO`) or is an actual postcode with no results (`WC1E 9ZZ` is an example). We need to know the difference between this two states in order to provide better error messages back to the User.

Right now if a User enters what to them is a valid postcode like `WC1E 9ZZ`, we tell them it isn't valid because we just check for nil. The user has a similar experience when partial postcode like `WC2B`.

Is a better idea for this to leave the existing behaviour and add a bang method with my changes?

Out of curiosity, what Apps use the Mapit API other than Frontend?

alext:
If the goal is to differentiate between invalid postcodes and non-existent postcodes, we're probably better off doing our own validation using something like the [uk postcode gem](https://rubygems.org/gems/uk_postcode) before we even call Mapit.

> Out of curiosity, what Apps use the Mapit API other than Frontend?

Imminence and content API both call out to Mapit, and there may be others as well.

tommyp:
If I'm reading correctly, Postcode Gem is checking to see if a Postcode _looks_ like a Postcode. It doesn't check to see if it actually exists. For the 404 state Mapit is actually saying "This is a valid Postcode but I can't find a match in Mapit". Our Mapit database can get out of date, especially when it comes to new builds. This is something the Gem can't help us with.

tommyp:
After chatting in meatspace, we've decided not to return a 404 and instead return nil. This is the same behaviour as other APIs. I'm going to close this PR and open another one.

PR #379, opened 2015-10-27
Add publishing api path reservation test helpers...

steventux:
Part of https://trello.com/c/bf4nuPIc/340-move-url-arbitration-to-publishing-api

Moves [url arbiter test helpers](https://github.com/alphagov/govuk-client-url_arbiter/blob/master/lib/govuk/client/test_helpers/url_arbiter.rb).

Note I've omitted the `:get` stubs in a couple of the helpers as these weren't used in Panopticon.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api.rb
+++ b/lib/gds_api/test_helpers/publishing_api.rb
@@ -90,6 +90,39 @@ module GdsApi
         stub_request(:any, /#{PUBLISHING_API_ENDPOINT}\/.*/).to_return(:status => 503)
       end
 
+      def stub_default_publishing_api_path_reservation
+        stub_request(:put, %r[\A#{PUBLISHING_API_ENDPOINT}/paths/]).to_return { |request|
+           base_path = request.uri.path.sub(%r{\A/paths/}, "")
+           { :status => 200,  :headers => {:content_type => "application/json"},
+             :body => publishing_api_path_data_for(base_path).to_json }
+        }
+      end
+
+      def publishing_api_has_path_reservation_for(path, publishing_app)
+        data = publishing_api_path_data_for(path, "publishing_app" => publishing_app)
+        error_data = data.merge({
+          "errors" => {"path" => ["is already reserved by the #{publishing_app} application"]},
+        })
+
+        stub_request(:put, "#{PUBLISHING_API_ENDPOINT}/paths#{path}").
+                  to_return(:status => 422, :body => error_data.to_json,
+                            :headers => {:content_type => "application/json"})
+
+        stub_request(:put, "#{PUBLISHING_API_ENDPOINT}/paths#{path}").
+          with(:body => {"publishing_app" => publishing_app}).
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -90,6 +90,39 @@ def publishing_api_isnt_available
...
+                            :headers => {:content_type => "application/json"})
+
+        stub_request(:put, "#{PUBLISHING_API_ENDPOINT}/paths#{path}").
+          with(:body => {"publishing_app" => publishing_app}).
+          to_return(:status => 200,
+                    :headers => {:content_type => "application/json"},
+                    :body => data.to_json)
+      end
+
+      def publishing_api_returns_validation_error_for(path, error_details = nil)
```
alext: I think this method needs naming in a way that makes it clear it's for path reservations.

lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -90,6 +90,39 @@ def publishing_api_isnt_available
         stub_request(:any, /#{PUBLISHING_API_ENDPOINT}\/.*/).to_return(:status => 503)
       end
 
+      def stub_default_publishing_api_put_path
```
alext: I think this also needs to be named in a way that makes it clear it's for the path reservations. Maybe something like `stub_default_publishing_api_path_reservation`

lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -90,6 +90,39 @@ def publishing_api_isnt_available
...
+      end
+
+      def publishing_api_has_path_reservation_for(path, publishing_app)
+        data = publishing_api_path_data_for(path, "publishing_app" => publishing_app)
+        error_data = data.merge({
+          "errors" => {"path" => ["is already reserved by the #{publishing_app} application"]},
+        })
+
+        stub_request(:put, "#{PUBLISHING_API_ENDPOINT}/paths#{path}").
+                  to_return(:status => 409, :body => error_data.to_json,
```
danielroseman: We're returning 422 now.

lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -90,6 +90,39 @@ def publishing_api_isnt_available
         stub_request(:any, /#{PUBLISHING_API_ENDPOINT}\/.*/).to_return(:status => 503)
       end
 
+      def stub_default_publishing_api_put_path
+        stub_request(:put, %r[\A#{PUBLISHING_API_ENDPOINT}/paths/]).to_return { |request|
+           base_path = request.uri.path.sub(%r{\A/paths/}, "")
+           { :status => 201,  :headers => {:content_type => "application/json"},
```
danielroseman: I think all our put endpoints return 200.

steventux:
@alext, @danielroseman thanks for review, updated.

PR #381, opened 2015-10-28
Mapit API returns error on 400

tommyp:
This change allows the API adapter for Mapit to return an error when the
User enters in an invalid postcode. This will allow us to log in
Frontend when this happens.

### DIFF
--- a/lib/gds_api/mapit.rb
+++ b/lib/gds_api/mapit.rb
@@ -5,10 +5,7 @@ class GdsApi::Mapit < GdsApi::Base
 
   def location_for_postcode(postcode)
     response = get_json("#{base_url}/postcode/#{CGI.escape postcode}.json")
-    return Location.new(response) unless response.nil?
-  rescue GdsApi::HTTPErrorResponse => e
-    # allow 400 errors, as they can be invalid postcodes people have entered
-    raise e unless e.code == 400
+    Location.new(response) unless response.nil?
   end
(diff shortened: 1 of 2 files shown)


lib/gds_api/mapit.rb
```diff
@@ -5,10 +5,11 @@ class GdsApi::Mapit < GdsApi::Base
 
   def location_for_postcode(postcode)
     response = get_json("#{base_url}/postcode/#{CGI.escape postcode}.json")
-    return Location.new(response) unless response.nil?
+    Location.new(response) unless response.nil?
+
   rescue GdsApi::HTTPErrorResponse => e
     # allow 400 errors, as they can be invalid postcodes people have entered
```
tijmenb: The comment doesn't apply anymore.

lib/gds_api/mapit.rb
```diff
@@ -5,10 +5,11 @@ class GdsApi::Mapit < GdsApi::Base
...
-    raise e unless e.code == 400
+    e
```
alext: It's a little odd to have this method return the exception object instead of raising the exception. It would be more consistent with other adapters for this to simple raise pass the exception up (ie just get rid of the `rescue` clause entirely.

tommyp:
Removed the `rescue` and changed the test to check that a bad postcode will raise an exception.

PR #383, opened 2015-11-02
Add support for optimistic locking to the v2 publishing API endpoints.

elliotcm:
This allows publisher applications to detect conflicts
when (for example) a user has left their edit page
open for a long time and someone else has updated
the item in the meantime.

If the version the app thinks it's updating does
not match the version in the Publishing API, a 409
Conflict is returned.

If the `previous_version` key is not provided, no
checking is performed.  This makes this locking
opt-in on the part of the publisher application.

https://trello.com/c/s3dN4eDP/336-optimistic-locking

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -14,8 +14,16 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
   end
 
   def publish(content_id, update_type, options = {})
-    params = { update_type: update_type }
-    params = params.merge(locale: options[:locale]) if options[:locale]
+    params = {
+      update_type: update_type
+    }
+
+    optional_keys = [
+      :locale,
+      :previous_version,
+    ]
+
+    params = merge_optional_keys(params, options, optional_keys)
 
     post_json!(publish_url(content_id), params)
   end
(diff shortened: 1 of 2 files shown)


test/publishing_api_v2_test.rb
```diff
@@ -139,6 +139,75 @@ def content_item_for_content_id(content_id, attrs = {})
...
       end
     end
+
+    describe "optimistic locking" do
+      describe "if the content item has not change since it was requested" do
+        before do
+          @content_item = content_item_for_content_id(@content_id, "previous_version" => 3)
+
+          publishing_api
+            .given("the content item is at version 3")
```
alext: The given will need to include a specific content_id, which will need to be specified in the paths below.  When setting up the provider state in the producer, there's no way for it to access `@content_id`.  Including it in the state description keeps this obvious, and is also consistent with some of the states below.

lib/gds_api/publishing_api_v2.rb
```diff
@@ -43,4 +56,14 @@ def links_url(content_id)
   def publish_url(content_id)
     "#{endpoint}/v2/content/#{content_id}/publish"
   end
+
+  def merge_optional_keys(params, options, optional_keys)
+    optional_keys.inject(params) do |hash, optional_key|
```
alext: This could be simpler using `each_with_object`. You wouldn't need the `else` clause that way.

test/publishing_api_v2_test.rb
elliotcm: Updated.

lib/gds_api/publishing_api_v2.rb
elliotcm: Updated, though I find them both equally awkward.

tuzz:
Looks good to me. Happy for this to be merged.

alext:
Looks good, but I'm not sure this can be merged until the corresponding change in publishing-api has been merged, otherwise this will break both the master build here, and in publishing-api.

elliotcm:
Pubishing API changes here: https://github.com/alphagov/publishing-api/pull/113

PR #384, opened 2015-11-04
Autocorrect obvious typos in postcodes before sending to MapIt

issyl0:
- We were seeing lots of errors for postcodes entered with obvious
  typos, eg. hyphens instead of spaces, 0 instead of O, I instead of 1,
  etc, so this is an attempt at fixing most common ones. Strip most
  non-alphanumerics and trailing whitespace. Also use the uk_postcode
  gem to fix transposition of I/1, O/0.
- This was originally done in Imminence (https://github.com/alphagov/imminence/pull/116), but then it turned
  out that it wasn't working in the way I thought it did, with multiple
  apps (frontend included - vital for postcode lookups on local
  transaction pages) going directly to MapIt via the MapIt GDS API
  Adapter, hence doing the munging in here instead. Even Imminence uses
  this API Adapter, so the behaviour will be the same without the PR on
  Imminence.

(Thanks to @h-lame and @davidbasalla for helping me make sense of what was going on!)

### DIFF
--- a/gds-api-adapters.gemspec
+++ b/gds-api-adapters.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'lrucache', '~> 0.1.1'
   s.add_dependency 'rest-client', '~> 1.8.0'
   s.add_dependency 'rack-cache'
+  s.add_dependency 'uk_postcode', '2.1.0'
 
   s.add_development_dependency 'gem_publisher', '~> 1.5.0'
   s.add_development_dependency 'mocha', "> 1.0.0"
--- a/lib/gds_api/mapit.rb
+++ b/lib/gds_api/mapit.rb
@@ -1,9 +1,12 @@
 require_relative 'base'
 require_relative 'exceptions'
+require_relative 'postcode_sanitizer'
+
 
 class GdsApi::Mapit < GdsApi::Base
 
   def location_for_postcode(postcode)
+    postcode = GdsApi::PostcodeSanitizer.sanitize(postcode)
     response = get_json("#{base_url}/postcode/#{CGI.escape postcode}.json")
     Location.new(response) unless response.nil?
   end
--- a/lib/gds_api/postcode_sanitizer.rb
+++ b/lib/gds_api/postcode_sanitizer.rb
@@ -1,0 +1,11 @@
+require 'uk_postcode'
+
+module GdsApi
+  class PostcodeSanitizer
+    def self.sanitize(postcode)
+      # Strip trailing whitespace, non-alphanumerics, and use the
+      # uk_postcode gem to potentially transpose O/0 and I/1.
+      UKPostcode.parse(postcode.gsub(/[^a-z0-9 ]/i, '').strip).to_s
+    end
+  end
+end
--- a/test/mapit_test.rb
+++ b/test/mapit_test.rb
@@ -45,6 +45,14 @@ describe GdsApi::Mapit do
       assert_equal "30UN", response.areas.last.codes['ons']
     end
 
+    it "should strip non-alphanumeric characters apart from a single space from the postcode, and potentially transpose letters for numbers" do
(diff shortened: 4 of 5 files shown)


gds-api-adapters.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'lrucache', '~> 0.1.1'
   s.add_dependency 'rest-client', '~> 1.8.0'
   s.add_dependency 'rack-cache'
+  s.add_dependency 'uk_postcode', '2.1.0'
```
tijmenb: Would there be a way to make this dependency opt-in, so that apps not-using `mapit` don't have to download/load this gem?

issyl0: I was about to suggest we just `require` it at the top of the `mapit` file, but it would still need installing somehow. I'll look into if we can do it your way...

tijmenb: In `govuk_message_queue_consumer` we are using Airbrake without requiring it. Perhaps we could  raise an error `unless defined?(UKPostcode)` in the postcode sanitizer?

h-lame: Rubygems doesn't support optional dependencies so to do this we'd have to remove the dependency entirely.  We could make requiring `gds_api/mapit` fail if `uk_postcode` isn't available (it probably would anyway because the `require 'uk_postcode'` would raise a `LoadError` - but we might want to trap that and raise a nicer error.  We'd need to  put something in the readme explaining this though - it's not the best user experience.  If you're really making the mapit adapter optional, perhaps the solution is to extract the mapit adaptor into its own gem?

alext:
I question whether this is the correct place to do this auto-correction. It seems that this is a user-interaction concern, not part of the API interaction, and as such, it should probably be in the app that handles the interaction with the user (ie frontend)

gds-api-adapters.gemspec
h-lame: The approach taken by `govuk_message_queue_consumer` is that if Airbrake is available, cool, we'll tell it about an error, if not though that's cool too we just won't do anything.  We could take the same approach - if the uk_postcode gem is available we'll sanitize the postcode, if not, we won't, but I'm not sure that gives us much benefit.

The ultimate problem here is that no matter what we do about making the dependency optional (required to use mapit at all or required only if you want to sanitize postcodes) we have to put  `gem 'uk_postcode'` into the Gemfile of every app that wants to talk to mapit about postcodes, otherwise we don't get the benefit that the `PostcodeSanitizer` is there to provide - and that seems a bit onerous.

issyl0:
@alext: I originally tried in Imminence, thinking that Imminence wrapped MapIt for all our apps. Then I realised that Frontend talks to MapIt directly, and that to autocorrect postcodes everywhere in here would be better than doing the same thing in the _n_ different apps that take user-inputted postcodes? I suppose we're already doing a subset of this correction on its own in Business Support Finder ('cause of an error I spotted and fixed a few months ago that was completely unrelated to any of these local transaction stories), but should we not be completely consistent and reduce duplication? Why would autocorrecting postcodes everywhere be a bad thing - if users typo their postcode, we should be nice?

boffbowsh:
I also don't really like the feel of this. Can you enumerate the apps that this would apply to please @issyl0? This gem isn't supposed to do too much in the way of logic.

issyl0:
Actually, the point I'm labouring is slightly invalid because, having enumerated, _n_ = 2. Imminence and Frontend. So I suppose we could do it in both apps and not have this.

boffbowsh:
Ah that's ok then. Thanks :)

issyl0:
Right. Closing this in favour of a separate approach: https://github.com/alphagov/imminence/pull/116 and https://github.com/alphagov/frontend/pull/893.

alext:
I'm not sure why we would need this in Imminence. It doesn't seem correct for an API to be doing this autocorrection.  IMO, that should be done by the component that interacts with the end-user, not an API further down the chain.

PR #388, opened 2015-11-16
Add test helpers for Publishing API v2

tommyp:
This commit adds test helpers for the Publishing API v2. Most of these have
been copied over from the v1 test helpers and modified.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -1,0 +1,28 @@
+require 'gds_api/test_helpers/json_client_helper'
+require 'gds_api/test_helpers/content_item_helpers'
+require 'gds_api/test_helpers/intent_helpers'
+require 'json'
+
+module GdsApi
+  module TestHelpers
+    module PublishingApiV2
+      include ContentItemHelpers
+
+      PUBLISHING_API_V2_ENDPOINT = Plek.current.find('publishing-api') + '/v2'
+
+      def stub_publishing_api_put_content(content_id, body)
+        stub_publishing_api_put(content_id, body, '/content')
+      end
+
+      def stub_publishing_api_put_links(content_id, body)
+        stub_publishing_api_put(content_id, body, '/links')
+      end
+
+    private
+      def stub_publishing_api_put(content_id, body, resource_path)
+        url = PUBLISHING_API_V2_ENDPOINT + resource_path + "/" + content_id
+        stub_request(:put, url).with(body: body).to_return(status: 200, body: '{}', headers: {"Content-Type" => "application/json; charset=utf-8"})
+      end
+    end
+  end
+end
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -18,6 +18,11 @@ module GdsApi
         stub_publishing_api_put(content_id, body, '/links')
       end
 
+      def stub_publishing_api_publish(content_id, body)
+        url = PUBLISHING_API_V2_ENDPOINT + "/content/#{content_id}/publish"
+        stub_request(:post, url).with(body: body).to_return(status: 200, body: '{}', headers: {"Content-Type" => "application/json; charset=utf-8"})
+      end
+
       def stub_default_publishing_api_put
         stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/content})
       end
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -18,6 +18,43 @@ module GdsApi
         stub_publishing_api_put(content_id, body, '/links')
       end
 
+      def stub_default_publishing_api_put
+        stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/content})
+      end
+
+      def assert_publishing_api_put_item(content_id, attributes_or_matcher = {}, times = 1)
+        url = PUBLISHING_API_V2_ENDPOINT + "/content/" + content_id
+        assert_publishing_api_put(url, attributes_or_matcher, times)
+      end
+
+      def assert_publishing_api_put(url, attributes_or_matcher = {}, times = 1)
+        if attributes_or_matcher.is_a?(Hash)
+          matcher = attributes_or_matcher.empty? ? nil : request_json_matching(attributes_or_matcher)
+        else
+          matcher = attributes_or_matcher
+        end
+
+        if matcher
+          assert_requested(:put, url, times: times, &matcher)
+        else
+          assert_requested(:put, url, times: times)
+        end
+      end
+
+      def request_json_matching(required_attributes)
+        ->(request) do
+          data = JSON.parse(request.body)
+          required_attributes.to_a.all? { |key, value| data[key.to_s] == value }
+        end
+      end
+
+      def request_json_including(required_attributes)
+        ->(request) do
+          data = JSON.parse(request.body)
+          required_attributes == data
+        end
+      end
+
     private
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -0,0 +1,93 @@
...
+      def stub_default_publishing_api_v2_put()
+        stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/content})
+      end
+
+      def stub_default_publishing_api_v2_put_draft()
+        stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/draft-content})
+      end
+
+      def stub_default_publishing_api_v2_put_intent()
+        stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/publish-intent})
```
benilovj: does this endpoint exist yet?

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -0,0 +1,93 @@
...
+      def stub_publishing_api_v2_put_links(content_id, body)
+        stub_publishing_api_v2_put(content_id, body, '/links')
+      end
+
+      def stub_default_publishing_api_v2_put()
+        stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/content})
+      end
+
+      def stub_default_publishing_api_v2_put_draft()
+        stub_request(:put, %r{\A#{PUBLISHING_API_V2_ENDPOINT}/draft-content})
```
benilovj: this won't work like that in v2...

lib/gds_api/test_helpers/publishing_api_v2.rb
tommyp: I'm not sure. Maybe someone can weight in? @tuzz, @steventux or @elliotcm?

jamiecobbett: I believe that publish-intent isn't part of v2. Some apps might need to use the unscoped/unversioned publish-intent and the v2 endpoints until publishing-api can provide the behaviour to make it redundant.

tuzz: Publish Intents are v1 only: https://github.com/alphagov/publishing-api/blob/master/config/routes.rb#L6-L8

That method can be removed.

tuzz: Also, those stubbed routes aren't correct. We don't have a /draft-content endpoint anymore. Instead, there's a single endpoint for content items and they are transitioned into a published state by hitting the publish endpoint.

We can help with this pull request if you put something on Slack tomorrow.

benilovj:
@tuzz I also need these stubs as I'm working on a v2 story, I'll sync with @tommyp tomorrow to correct these...

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+      end
+
+    private
+      def stub_publishing_api_v2_put(content_id, body, resource_path)
+        url = PUBLISHING_API_V2_ENDPOINT + resource_path + "/" + content_id
+        stub_request(:put, url).with(body: body).to_return(status: 200, body: '{}', headers: {"Content-Type" => "application/json; charset=utf-8"})
+      end
+
+      def values_match_recursively(expected_value, actual_value)
```
elliotcm: I'm not sure what you're achieving here that isn't achieved by `==`.

```
2.1.6 :001 > hash = {a: 1, b: {c: 3, d: [{e: 4, f: 5}]}}
 => {:a=>1, :b=>{:c=>3, :d=>[{:e=>4, :f=>5}]}} 
2.1.6 :002 > hash == hash
 => true 
2.1.6 :003 > hash === hash
 => true 
```

tommyp: There were copied from Publishing API V1 test helpers. I'll change it if following a simpler approach is better.

elliotcm: If we don't know why we need it then we might not need it.

elliotcm: Also, if it turns out it's there to provide clearer diffing when they don't match, we could use something like https://github.com/liufengyun/hashdiff

jamiecobbett: @elliotcm that hypothesis rings a bell. Using that library seems like a good idea anyway.

tommyp: I'm gonna get rid of this method since I'm not using it.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -0,0 +1,74 @@
...
+      def stub_publishing_api_v2_put_links(content_id, body)
+        stub_publishing_api_v2_put(content_id, body, '/links')
+      end
+
+      def stub_publishing_api_v2_publish(content_id, body)
+        url = PUBLISHING_API_V2_ENDPOINT + "/content/#{content_id}/publish"
+        stub_request(:post, url).with(body: body).to_return(status: 200, body: '{}', headers: {"Content-Type" => "application/json; charset=utf-8"})
+      end
+
+      def stub_default_publishing_api_v2_put()
```
jamiecobbett: I think a much better name for this would be `stub_any_publishing_api_v2_put_content`. If we want it to stub PUTs to links (which this doesn't do currently), then I suppose it should lose my proposed `_content`.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+    private
+      def stub_publishing_api_v2_put(content_id, body, resource_path)
+        url = PUBLISHING_API_V2_ENDPOINT + resource_path + "/" + content_id
+        stub_request(:put, url).with(body: body).to_return(status: 200, body: '{}', headers: {"Content-Type" => "application/json; charset=utf-8"})
+      end
+
+      def values_match(expected_value, actual_value)
+        expected_value == actual_value
```
jamiecobbett: Now that this implementation is trivial and has a single use, I think it should be inlined.

tijmenb:
Would it be clearer to skip the `v2` in the method names? They won't be used at the same time as v1, and v1 is going to be removed at one point anyway.

jamiecobbett:
@tijmenb I think that would make it very confusing/impossible to use the helpers for both API versions - which will be needed by apps that need to use publish-intents, for example.

benilovj:
@tijmenb yep, as @jamiecobbett said, we will need to run against both API versions for a while in Whitehall...

tijmenb:
@jamiecobbett @benilovj 

We could solve that by adding the publish-intents methods to `PublishingApiV2`.

I was under the impression that V2 is the replacement of V1. If intents are a current thing, don't they belong in V2 too?

jamiecobbett:
@tijmenb the rationale is that they are only required until the publishing platform handles scheduled publishing itself - ie phase 2.

tijmenb:
@jamiecobbett at which time they can be removed from the API, like normally happens, right?

I'm worried that references to versions like "v2" will be a thing in all apps years from now, even once `v1` is retired.

tijmenb:
If Whitehall is the only one using the two adapters at the same time (which it really shouldn't), we could `alias` the helper methods there. That way other apps can keep using helpers without the version.

lib/gds_api/test_helpers/publishing_api_v2.rb
benilovj: done

benilovj:
@jamiecobbett @tijmenb As discussed irl, I've removed `_v2` from the method names.

PR #391, opened 2015-11-20
Add helpers to stub Pub API v2 show and index

tommyp:
These helpers can be used to stub out the `show` and `index` endpoints
for the Publishing API V2.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -70,6 +70,26 @@ module GdsApi
         end
       end
 
+      def publishing_api_has_fields_for_format(format, items, fields)
+        body = items.map { |item|
+          item.with_indifferent_access.slice(*fields)
+        }
+
+        query_params = fields.map { |f|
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -70,6 +70,26 @@ def request_json_including(required_attributes)
         end
       end
 
+      def publishing_api_has_fields_for_format(format, items, fields)
+        body = items.map { |item|
+          item.with_indifferent_access.slice(*fields)
+        }
+
+        query_params = fields.map { |f|
+          "&fields%5B%5D=#{f}"
```
tuzz: Could there be cases where fields need to be URI escaped? I can't think of any, so this is probably fine.

tuzz:
Looks fine. :+1: from me

lib/gds_api/test_helpers/publishing_api_v2.rb
tommyp: Me neither, but we can always change it later if need be.

PR #401, opened 2015-12-01
Fix publishing api v2 publish stub

jamiecobbett:
It was nesting the keys inside an update_type key, but shouldn't have been.

Code tested with this helper will be sending the update_type and locale keys in a way that will be ignored by publishing-api. This is currently true of Whitehall.

I think this would be a PATCH release.

/cc @elliotcm @benilovj


tuzz:
:+1: nice catch

jamiecobbett:
Please hold off merging while I investigate the Whitehall failures (yay massive test suites).

benilovj:
:+1:

PR #404, opened 2015-12-07
Migrate SpecialRoutePublisher to use v2 of publishing-api

stephen-richards:
All clients of publishing-api should use v2 of the API.  This library contains one class, SpecialRoutePublisher which talks to the publishing-api (rather than just provide convenience adapter methods to do so).  This change modifies that class to use v2 of the API.

Trello: https://trello.com/c/yaUJdIRH/456-migrate-gds-api-adapters-gdsapi-publishingapi-specialroutepublisher-to-use-v2-of-publishing-api

### DIFF
--- a/lib/gds_api/publishing_api/special_route_publisher.rb
+++ b/lib/gds_api/publishing_api/special_route_publisher.rb
@@ -1,4 +1,4 @@
-require "gds_api/publishing_api"
+require "gds_api/publishing_api_v2"
 require "time"
 
 module GdsApi
@@ -6,14 +6,14 @@ module GdsApi
     class SpecialRoutePublisher
       def initialize(options = {})
         @logger = options[:logger] || GdsApi::Base.logger
-        @publishing_api = options[:publishing_api] || GdsApi::PublishingApi.new(Plek.find("publishing-api"))
+        @publishing_api = options[:publishing_api] || GdsApi::PublishingApiV2.new(Plek.find("publishing-api"))
       end
 
       def publish(options)
         logger.info("Publishing #{options.fetch(:type)} route #{options.fetch(:base_path)}, routing to #{options.fetch(:rendering_app)}")
 
-        publishing_api.put_content_item(options.fetch(:base_path), {
-          content_id: options.fetch(:content_id),
+        put_content_response = publishing_api.put_content(options.fetch(:content_id), {
+          base_path: options.fetch(:base_path),
           format: "special_route",
           title: options.fetch(:title),
           description: options[:description] || "",
@@ -25,9 +25,10 @@ module GdsApi
           ],
           publishing_app: options.fetch(:publishing_app),
           rendering_app: options.fetch(:rendering_app),
-          update_type: "major",
           public_updated_at: time.now.iso8601,
         })
+        publishing_api.publish(options.fetch(:content_id), 'major')
+        put_content_response
       end
 
     private
--- a/test/publishing_api/special_route_publisher_test.rb
+++ b/test/publishing_api/special_route_publisher_test.rb
@@ -1,18 +1,14 @@
 require 'test_helper'
 require "gds_api/publishing_api/special_route_publisher"
+require File.dirname(__FILE__) + '/../../lib/gds_api/test_helpers/publishing_api_v2'
 
 describe GdsApi::PublishingApi::SpecialRoutePublisher do
-  let(:publishing_api) {
-    stub(:publishing_api, put_content_item: nil)
-  }
+  include ::GdsApi::TestHelpers::PublishingApiV2
 
-  let(:publisher) {
-    GdsApi::PublishingApi::SpecialRoutePublisher.new(publishing_api: publishing_api)
-  }
-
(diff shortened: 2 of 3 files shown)


lib/gds_api/publishing_api/special_route_publisher.rb
```diff
@@ -12,7 +12,8 @@ def initialize(options = {})
       def publish(options)
         logger.info("Publishing #{options.fetch(:type)} route #{options.fetch(:base_path)}, routing to #{options.fetch(:rendering_app)}")
 
-        publishing_api.put_content_item(options.fetch(:base_path), {
+        put_content_response = publishing_api.put_content(options.fetch(:content_id), {
+          base_path: options.fetch(:base_path),
           content_id: options.fetch(:content_id),
```
tijmenb: `content_id` is no longer required in V2 (because it's inferred from the URL params)

test/publishing_api/special_route_publisher_test.rb
```diff
@@ -22,13 +23,12 @@
...
-  describe ".publish" do
-    it "publishes the special routes" do
-      Timecop.freeze(Time.now) do
-        publishing_api.expects(:put_content_item).with(
-          special_route[:base_path],
-          {
-            content_id: special_route[:content_id],
+  let(:content_id) { special_route[:content_id] }
+
+  let(:payload) {
```
tijmenb: The indenting here is a bit odd now. It might be worth it to inline `payload` into the `publishes the special routes` test, as it's only used there.

test/publishing_api/special_route_publisher_test.rb
```diff
@@ -40,12 +40,18 @@
...
             publishing_app: special_route[:publishing_app],
             rendering_app: special_route[:rendering_app],
-            update_type: "major",
             public_updated_at: Time.now.iso8601,
           }
-        )
+  }
 
-        publisher.publish(special_route)
+  let(:ok_response) { OpenStruct.new({code: 200}) }
```
tijmenb: Perhaps we can user the api test helpers to replace the expectations on the `publishing_api` object:

```ruby
it "publishes the special routes" do
  stub_any_publishing_api_call

  publisher.publish(special_route)

  assert_publishing_api_put_content(....)
  assert_publishing_api_publish(...)
end
```

This feels more in line with how normal calls are tested, and would remove the need for returning mock OpenStruct objects.

lib/gds_api/publishing_api/special_route_publisher.rb
stephen-richards: content_id is removed - but we need to get the documentation changed which suggests that it is required: https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#put-v2contentcontent_id  (click Request/Response detail)

test/publishing_api/special_route_publisher_test.rb
stephen-richards: Agreed, Now that we are not testing for the response code

tijmenb:
I've tried running this:

```
vagrant@development:~/govuk/rummager (master)$ be rake publishing_api:publish_special_routes
Publishing exact route /sitemap.xml, routing to rummager
rake aborted!
undefined method `put_content' for #<GdsApi::PublishingApi:0x007f4583beaa08>
```

This is because `SpecialRoutePublisher` still uses the `PublishingApi` V1 class. Because we've stubbed the `put_content` and `publish` calls, the tests pass while they shouldn't.

I would suggest getting rid of the stubs first and test with the `assert_X` test helpers, so that the tests fail (like https://github.com/alphagov/gds-api-adapters/pull/404#discussion_r47073218).

stephen-richards:
Have changed in line with Tijmen's comments

test/publishing_api/special_route_publisher_test.rb
```diff
@@ -24,55 +21,71 @@
...
+          rendering_app: special_route[:rendering_app],
+          public_updated_at: Time.now.iso8601,
+        }
 
+        stub_any_publishing_api_call
+        
         publisher.publish(special_route)
+
+        base_path = "#{endpoint}/v2/content/#{content_id}"
+        assert_requested(:put, base_path, body: payload)
```
tijmenb: Any reason not to use `assert_publishing_api_put_content` here?

lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '26.3.1'
+  VERSION = '26.4.1'
```
tijmenb: Shouldn't update the version here. This is probably why this PR has merge conflicts.

boffbowsh:
Any objection to me picking this up and getting it across the line? We want to update the feedback app to use V2 for its special route.

lib/gds_api/version.rb
stephen-richards: Reverted

stephen-richards:
boffbowsh - yep, please do. I think I've done all that I was going to do anyway - would be interested to see if you want to make any changes.

PR #407, opened 2015-12-10
Add additional publishing-api-v2 test helpers

stephen-richards:
Additional test helper methods for publishing-api-v2 to added so that all clients can make use of them:

- specify a response when stubbing PUT content
- specify a version when stubbing POST content/content_id/version
- specify a put_content, publish, put_links sequence with any body
- assert that put_links has been called

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -10,8 +10,20 @@ module GdsApi
 
       PUBLISHING_API_V2_ENDPOINT = Plek.current.find('publishing-api') + '/v2'
 
-      def stub_publishing_api_put_content(content_id, body)
-        stub_publishing_api_put(content_id, body, '/content')
+      # stubs a PUT /v2/content/:content_id request with the given content id and request body.
+      # if no response_hash is given, a default response as follows is created:
+      # {status: 200, body: '{}', headers: {"Content-Type" => "application/json; charset=utf-8"}}
+      #
+      # if a response is given, then it will be merged with the default response.
+      # if the given parameter for the response body is a Hash, it will be converted to JSON.
+      #
+      # e.g. The following two examples are equivalent: 
+      #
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -14,6 +14,15 @@ def stub_publishing_api_put_content(content_id, body)
         stub_publishing_api_put(content_id, body, '/content')
       end
 
+      def stub_publishing_api_put_content_and_return_response(content_id, response)
```
tijmenb: What is the reason that you're choosing to add specific methods for a stub that returns something?

For example, this method could be created by adding a third argument to `#stub_publishing_api_put_content`.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -28,6 +37,16 @@ def stub_publishing_api_discard_draft(content_id)
         stub_request(:post, url).to_return(status: 200, headers: {"Content-Type" => "application/json; charset=utf-8"})
       end
 
+      def stub_publishing_api_publish_with_version(content_id, version, update_type = 'major')
+        stub_publishing_api_publish(content_id, {update_type: update_type, previous_version: version}.to_json)
+      end
+
+      def stub_any_publishing_api_put_content_publish_and_put_links(content_id)
```
tijmenb: I don't understand why this should be different from `stub_publishing_api_put_content_links_and_publish`.

lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '26.4.0'
+  VERSION = '26.4.2'
```
tijmenb: Do not increment the version here. Do this in a separate PR, examples: https://github.com/alphagov/gds-api-adapters/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+bump

tijmenb:
Note that the changelog also needs updating here.

PR #408, opened 2015-12-10
Add `PublishingApiV2#get_linked_items`

jackscotti:
This method is used to interact with the `/v2/linked` endpoint on the
publishing-api.

Endpoint PR:
https://github.com/alphagov/publishing-api/pull/157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Add `PublishingApiV2#get_linked_items`.
+
 # 26.5.0
 
 * Changed SpecialRoutePublisher to use v2 of publishing-api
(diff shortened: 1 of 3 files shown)


tijmenb:
Could you add something to the changelog?

jackscotti:
Done, thanks. Blocked for now, depending on [this](https://github.com/alphagov/publishing-api/compare/release_174...release_179) being deployed.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 26.5.0
```
tijmenb: You won't know if there will be a release before this is merged, so the convention is to leave the version number out or put it under a `Unreleased` heading.

tijmenb:
@jackscotti 

What you need to do is set up a "provider state" in publishing-api:

https://github.com/alphagov/publishing-api/blob/master/spec/service_consumers/pact_helper.rb

PR #409, opened 2015-12-11
Pass publishing_app in get_content_items request

danielroseman:
The value is sourced from the GOVUK_APP_NAME env var by default; it can be
overridden by passing in the publishing_app parameter explicitly to
get_content_items. Passing a value of "all" will return all items.

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -53,6 +53,8 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
   end
 
   def get_content_items(params)
+    params = { publishing_app: app_name }.merge(params)
+    params.delete(:publishing_app) if params[:publishing_app] == 'all'
     query = query_string(params)
     get_json("#{endpoint}/v2/content#{query}")
   end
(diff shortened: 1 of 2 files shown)


tijmenb:
It may not be a good idea to make this method return filtered items by default. It's not backwards compatible with the current usage, which is to return all items (used in [content-tagger](https://github.com/alphagov/content-tagger/blob/bd6483113e77e9a6dab8db2a49722a2e4c05c39b/app/services/tag_fetcher.rb) for example).

I also think it may be confusing to devs to have a client method that returns different values depending on the place it is called. 

Finally, `GOVUK_APP_NAME` is not available in development, which means that in development this call will return something different from production, which will certainly trip somebody up.

If we really want this, perhaps we could make it more explicit by adding a method `get_own_content_items`?

danielroseman:
I'm not sure about the compatibility argument; we will definitely bump the version and document the change, so content-tagger will only need to add the `publishing_app: 'all'` parameter when it upgrades its version of the gem.

The idea though is very definitely for this to be the default. We will shortly be introducing token-based authentication for publishing-api, which will enforce this: apps will only be able to see the content they own. (At that point we can perhaps consider a "see all" permission for the case of apps like content-tagger.)

PR #410, opened 2015-12-11
Add additional publishing-api-v2 test helpers

stephen-richards:
Additional test helper methods for publishing-api-v2 to added so that all clients can make use of them:

- allow a response to be specified when stubbing put_content
- stub the publishing_api as not available


stephen-richards:
stub_publishing_api_put now defaults to the standard response if override_response hash is not provided as third param

PR #413, opened 2015-12-22
Add support for Rummager's `delete_content` & `get_content`

tijmenb:
These endpoints were added some time ago, but didn't have support in this gem yet.

cc @AndrewVos


AndrewVos:
lgtm

PR #415, opened 2016-01-13
Special route publisher puts links

benilovj:
This allows linking special routes to organisations (eg [here](https://github.com/alphagov/contacts-admin/blob/master/lib/tasks/publishing_api.rake#L17)).

/cc @elliotcm


elliotcm:
@danielroseman does this clash with your work?

danielroseman:
@ecm not really, only in that it will require another bump for apps that want to use it.

I'm not entirely sure I understand what it means for a special route to have links, though.

benilovj:
As discussed in person, adding links to the special route associates that piece of content with that organisation, same as we would with other formats.

PR #420, opened 2016-01-25
Include the request in the error message

jamiecobbett:
This should make the default error reporting much more useful. At the moment,
it can be very difficult to figure out what the request was that triggered the
error. Whilst clients could explicitly capture and record the problem request
body, that requires lots of manual work in every place a gds-api-adapter is
used.

The downside is that the error message suddenly becomes much longer. I can't
really see how that is a problem.

### DIFF
--- a/lib/gds_api/exceptions.rb
+++ b/lib/gds_api/exceptions.rb
@@ -14,10 +14,11 @@ module GdsApi
   class HTTPErrorResponse < BaseError
     attr_accessor :code, :error_details
 
-    def initialize(code, message = nil, error_details = nil)
+    def initialize(code, message = nil, error_details = nil, request_body = nil)
       super(message)
       @code = code
       @error_details = error_details
+      @request_body = request_body
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/gds_api/exceptions.rb
```diff
@@ -52,8 +53,8 @@ def ignoring_missing(&block)
       ignoring([HTTPNotFound, HTTPGone], &block)
     end
 
-    def build_specific_http_error(error, url, details = nil)
-      message = "url: #{url}\n#{error.http_body}"
+    def build_specific_http_error(error, url, details = nil, request_body = nil)
+      message = "URL: #{url} Response body: #{error.http_body} Request body: #{request_body}"
```
elliotcm: Interpolating two multiline strings into a single line string is going to be real hard to read.

elliotcm:
Are we at any risk of cross site injection by sending back the request?  Particularly in logs and errbit etc.

lib/gds_api/exceptions.rb
jamiecobbett: I can reintroduce the line break, though Errbit stripped that out when it displays the error.

elliotcm: If it only winds up in errbit don't bother, but if it's going into the app's logs it's worth having 3 lines.

jamiecobbett:
@elliotcm I see where you're coming from, but if that were a problem, it would already be a problem.

lib/gds_api/exceptions.rb
jamiecobbett: Rebased with a sprinkling of newlines.

PR #423, opened 2016-01-26
Pin Plek version to >= 1.9.0

tuzz:
The Plek.find method was added in 1.9.0.

We should ensure we're using a version greater-equal to this
because some of our calls to Plek are Plek.find.


tuzz:
Thanks @benilovj, there's also #424 if you have a sec.

PR #428, opened 2016-02-03
Publishing api switch to patch links

jamiecobbett:
Currently the Pact tests are failing. I think this is because the master of publishing-api doesn't have the PATCH route yet. Once https://github.com/alphagov/publishing-api/pull/198 has been merged, I think those tests will pass.


boffbowsh:
LGTM :ship: that :hankey: when :green_heart:

PR #429, opened 2016-02-09
Publshing API v2: harden adapter against nil content_ids

benilovj:
This catches the case when the publisher is trying to misuse the adapter
without having the error fail downstream (eg [this error](https://errbit.integration.publishing.service.gov.uk/apps/53020c2d0da11590e70000a0/problems/56b8a65f65786326b45b0000)).

/cc @alphagov/team-gov-uk-core-formats-and-publishing

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -59,25 +59,30 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
 
   def get_linked_items(content_id, params = {})
     query = query_string(params)
+    validate_content_id(content_id)
     get_json("#{endpoint}/v2/linked/#{content_id}#{query}")
   end
 
 private
 
   def content_url(content_id, params = {})
+    validate_content_id(content_id)
     query = query_string(params)
     "#{endpoint}/v2/content/#{content_id}#{query}"
   end
 
   def links_url(content_id)
+    validate_content_id(content_id)
     "#{endpoint}/v2/links/#{content_id}"
   end
 
   def publish_url(content_id)
+    validate_content_id(content_id)
     "#{endpoint}/v2/content/#{content_id}/publish"
   end
 
   def discard_url(content_id)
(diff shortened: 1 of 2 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -86,4 +91,8 @@ def merge_optional_keys(params, options, optional_keys)
       hash.merge!(optional_key => options[optional_key]) if options[optional_key]
     end
   end
+
+  def validate_content_id(content_id)
+    raise ArgumentError, "content_id cannot be nil" if content_id.nil?
```
benlovell: `unless content_id` is slightly more concise, at least in my humble opinion.

test/publishing_api_v2_test.rb
```diff
@@ -1232,4 +1232,24 @@ def content_item_for_content_id(content_id, attrs = {})
       end
     end
   end
+
+  describe "content ID validation" do
+    [:get_content, :get_links, :get_linked_items, :discard_draft].each do |method|
```
benlovell: Do we favour `%i(...)` for literal symbol arrays as we do with literal string arrays?

benlovell:
As usual my comments are mainly trivial styling, don't let that be a blocker. :+1:

lib/gds_api/publishing_api_v2.rb
benilovj: fixed

test/publishing_api_v2_test.rb
benilovj: I think this was only [introduced in Ruby 2.0](http://dev.venntro.com/2013/03/ruby-2-0-0-in-detail/#i-and-i-symbol-array-literal) and we need to support down to 1.9.3 with `gds-api-adapters`, as far as I know.

benlovell: Ah yes, I should have checked first. :speak_no_evil:

PR #435, opened 2016-02-10
Raise `HTTPUnauthorized` for 401 responses

tijmenb:
Publishing API has started to return 401s for requests without a bearer token. Raising a specific exception for this will allow us to debug this more easily.

Includes some refactoring / cleanup.


bradwright:
:+1: to this, just ran into this myself.

PR #439, opened 2016-02-11
Add a publishing_api_has_links helper

tommyp:
This is to stub that the Publishing API V2 contains a links payload.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -154,6 +154,12 @@ module GdsApi
         stub_request(:get, url).to_return(status: 200, body: item.to_json, headers: {})
       end
 
+      def publishing_api_has_links(links)
+        links = links.with_indifferent_access
+        url = PUBLISHING_API_V2_ENDPOINT + "/links/" + links[:content_id]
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -154,6 +154,12 @@ def publishing_api_has_item(item)
         stub_request(:get, url).to_return(status: 200, body: item.to_json, headers: {})
       end
 
+      def publishing_api_has_links(links)
+        links = links.with_indifferent_access
+        url = PUBLISHING_API_V2_ENDPOINT + "/links/" + links[:content_id]
```
boffbowsh: Would `content_id` be in the links payload? Perhaps this should be passed in as an arg to the function.

tommyp: I discussed this with @danielroseman and he said to follow the approach I've taken. For reference, the `links` payload looks like this:

```json
{
  "content_id": "b1dc075f-d946-4bcb-a5eb-941f8c8188cf",
  "links": {
    "organisations": [
      "d39237a5-678b-4bb5-a372-eb2cb036933d"
    ]
  },
  "version": 193
}
```

PR #449, opened 2016-02-19
Pass Govuk-Original-URL header on to sub requests

rboulton:
We're adding tracing of the original URL which triggered requests to our logging, so that we can identify which pages depend on which backend services.

This PR adds support for passing the Govuk-Original-Url header on to calls made by gds-api-adapters, and updates the railtie to set this up to happen automatically for Rails apps.  Other apps will need to add an explicit call, as detailed in the README.

This PR follows on from alphagov/govuk-puppet#4065, which sends the header from varnish and logs it from nginx.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# unreleased
+
+* Pass the Govuk-Original-Url header on to requests made by gds-api-adapters,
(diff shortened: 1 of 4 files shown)


mobaig:
LGTM

PR #451, opened 2016-02-19
Add Pub API does not have links test helper

tommyp:
This also makes the fake 404 response generic so it can be used for both
content and links.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -156,13 +156,18 @@ module GdsApi
 
       def publishing_api_does_not_have_item(content_id)
         url = PUBLISHING_API_V2_ENDPOINT + "/content/" + content_id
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -165,6 +165,11 @@ def publishing_api_has_links(links)
         stub_request(:get, url).to_return(status: 200, body: links.to_json, headers: {})
       end
 
+      def publishing_api_does_not_have_links(content_id)
+        url = PUBLISHING_API_V2_ENDPOINT + "/links/" + content_id
+        stub_request(:get, url).to_return(status: 404, body: resource_not_found(content_id "link set").to_json, headers: {})
```
tijmenb: Is `resource_not_found` missing a comma in the arguments?

tommyp: Good catch. Fixed.

boffbowsh:
:ship: when :green_heart:

PR #452, opened 2016-02-19
Bump to 28.2.1

tommyp:
This includes the `publishing_api_does_not_have_links` test helper added
in bcf11b.


bradwright:
Not sure about this versioning. It should have been `28.3.0`.

PR #453, opened 2016-02-25
Add places for postcode helper for imminence

brenetic:
- Helper methods added to enable the stubbing of find my nearest place searches via Imminence
- Small refactor to make use of the new helper methods


sihugh:
:+1:

PR #455, opened 2016-02-26
Fix missing end

brenetic:
- Fix missing `end`


benlovell:
The tests passed previously too :scream:

PR #469, opened 2016-03-07
Support searching by a document_type in email-alert-api

tuzz:
https://trello.com/c/oqGCU2XP/525-add-email-signup-links-to-travel-advice-on-frontend

As part of the work to migrate Travel Advice over to using the email-alert-frontend workflow, we need to be able to support subscriber lists that have a document_type but no tags or links. The email-alert-api now supports this (see https://github.com/alphagov/email-alert-api/commit/a6301c403f8b44f2244606c1c1cfd3f5dfee38b2).

This work makes it so that the adapters can pass a document_type through without tags or links.


elliotcm:
Ticket?

tuzz:
Added some context.

PR #470, opened 2016-03-07
Gem bump 29.4.0

tuzz:
Merge after #469


elliotcm:
Usually we create these after the PR has been merged in case other things have been merged in the meantime, affecting the version.

elliotcm:
Actually now I'm looking at this PR more carefully, the version bump shouldn't be in here with code changes.

PR #474, opened 2016-03-10
Add support for the publishing-api /lookup endpoint

tijmenb:
This endpoint was added in alphagov/publishing-api#230.

Documentation:

screen shot 2016-03-10 at 15 05 55

(will appear at http://www.rubydoc.info/gems/gds-api-adapters/GdsApi/PublishingApiV2)


Davidslv:
:+1:

PR #477, opened 2016-03-16
Wrap resultant calls to GET /v2/content in meta data

dougdroper:
Includes pagination and HATEOAS


jackscotti:
Pacts tests were failing.

PR #480, opened 2016-04-04
Extend #stub_publishing_api_publish to take response hash

christophwong:
This allows stubbing a non-200 response for testing unhappy paths


benilovj:
seems reasonable to me :+1:

PR #481, opened 2016-04-05
Bump gem version 30.0.1

christophwong:


christophwong:
merged with @dougdroper

tijmenb:
We really shouldn't merge our own PRs, slippery slope etc. Also the tests seem to be failing, is that related?

christophwong:
@tijmenb I showed the PR to @dougdroper and it looked ok.
Sorry about the test, when we merged it, it was definitely green, but it might be because of this morning's Github's server issue that the webhook were showing wrong info. I will revert it and redo this.

christophwong:
Oh actually the failing test seems to be a problem on jenkins and not this commit. As a rerun has fixed it.

tijmenb:
:+1: great

PR #482, opened 2016-04-05
Revert "Bump gem version 30.0.1"

christophwong:
Reverts alphagov/gds-api-adapters#481


christophwong:
closing as it seems that it wasn't this commit that caused jenkins to fail.

PR #483, opened 2016-04-07
Stubbed request to support pagination

Rosa-Fox:
Stubbed request for testing pagination within `specialist-publisher`.

In order to paginate content items within `specialist-publisher`, a user can pass  `page` and `per_page` params into the url query string.

The change in this commit stubs a request to `publishing-api` which includes pagination params. This is then used in tests within `specialist-publisher`. These tests can be found on [this branch](https://github.com/alphagov/specialist-publisher/tree/pagination-phase-2) - probably best to scroll through 'files changed', most model specs and some feature specs!

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -139,6 +139,16 @@ module GdsApi
         end
       end
 
+
+      # Example of use:
+      # publishing_api_has_content(
+      #   vehicle_recalls_and_faults,   # this is a variable containing an array of content items
+      #   document_type: described_class.publishing_api_document_type,   #example of a document_type: "vehicle_recalls_and_faults_alert"
+      #   fields: fields,   #example: let(:fields) { %i[base_path content_id public_updated_at title publication_state] }
+      #   page: 1,
+      #   per_page: 50
+      #)
+
       def publishing_api_has_content(items, params = {})
         body = Array(items).map { |item|
           item.with_indifferent_access.slice(*params[:fields])
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -139,32 +139,15 @@ module GdsApi
         end
       end
 
-      def publishing_api_has_fields_for_document(format, items, fields)
+      def publishing_api_has_content(items, params = {})
         body = Array(items).map { |item|
-          item.with_indifferent_access.slice(*fields)
+          item.with_indifferent_access.slice(*params[:fields])
         }
 
-        query_params = fields.map { |f|
-          "&fields%5B%5D=#{f}"
-        }
+        query_string = params.to_query
 
-        url = PUBLISHING_API_V2_ENDPOINT + "/content?document_type=#{format}#{query_params.join('')}"
-
-        stub_request(:get, url).to_return(:status => 200, :body => { results: body }.to_json, :headers => {})
-      end
-
-      def publishing_api_has_fields_for_document_with_pagination(format, items, fields, page, per_page)
-        body = Array(items).map { |item|
-          item.with_indifferent_access.slice(*fields)
-        }
-
-        query_params = fields.map { |f|
-          "&fields%5B%5D=#{f}"
-        }
-
-        url = PUBLISHING_API_V2_ENDPOINT + "/content?document_type=#{format}#{query_params.join('')}&page=#{page}&per_page=#{per_page}"
-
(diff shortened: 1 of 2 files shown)


tijmenb:
I'm worried that we're adding too many special cases to the test helpers - the module has ~30 public methods already. 

Perhaps we could come up with a couple of methods that would allow simpler stubbing, but are more explicit?

@Rosa-Fox can you link to the code that would use this helper?

Rosa-Fox:
Hi @tijmenb,

I am not too sure how to make it any simpler. I suppose maybe because it is so similar to the method above it (`publishing_api_has_fields_for_document`) I could set default values for the `page` and `per_page` arguments on `publishing_api_has_fields_for_document` and then append the `page` and `per_page` to the query string in that method instead? I guess then that would mean less duplication and the defaults should prevent 'wrong number of argument errors' but I thought it seemed less confusing to separate out. 

Helper is used in many of the model and feature specs. Here is an [example](https://github.com/alphagov/specialist-publisher-rebuild/blob/pagination-phase-2/spec/models/vehicle_recalls_and_faults_alert_spec.rb#L88). We do need to improve the tests later on in the project because there is currently too much code duplication between the model specs.

Rosa-Fox:
Hi @tijmenb,

I have looked into this a bit further. Previously `publishing_api_has_fields_for_document` is used to stub requests to the publishing-api in a `before` block [here](https://github.com/alphagov/specialist-publisher-rebuild/blob/pagination-phase-2/spec/models/vehicle_recalls_and_faults_alert_spec.rb#L88). It was also use to stub the request for the data to be sent to Rummager [here](https://github.com/alphagov/specialist-publisher-rebuild/blob/pagination-phase-2/spec/models/vehicle_recalls_and_faults_alert_spec.rb#L152). Now that pagination has been added things have changed as we need the pagination params to be included in the request for pagination but `page` and `per_page` don't really need to be sent to Rummager.

I could set `page` and `per_page` default arguments and use `publishing_api_has_fields_for_document` in the 2 different places like they had before but that would also mean passing in pagination parameters to the SearchPresenter [here](https://github.com/alphagov/specialist-publisher-rebuild/blob/pagination-phase-2/app/presenters/search_presenter.rb#L24) which doesn’t seem necessary.

Do you think it would work to keep as it is in this PR, but rename `publishing_api_has_fields_for_document` to `publishing_api_has_fields_for_search` to make the functionality clearer?

tijmenb:
To summarise a conversation on Slack, I think a generic test helper like this would work:

```ruby
def publishing_api_has_content(items, params = {})
  items = Array(items).map { |item|
    item.with_indifferent_access.slice(*params[:fields])
  }

  url = PUBLISHING_API_V2_ENDPOINT + "/content?#{query_string(params)}"

  stub_request(:get, url).to_return(body: { results: items }.to_json)
end
```

This would enable you to stub any publishing-api `/content` call:

```ruby
publishing_api_has_content(
  my_array_of_content_items,
  document_type: "the_document_type",
  fields: %i[title content_id],
  page: 1,
  per_page: 20
)
```

This would be good in the future, because if you're ever going to add another argument to the `publishing_api.get_content` you don't have to update this gem.

tijmenb:
@christophwong pointed me to the fact that the first example I posted referred to `publishing_api_has_content` and `publishing_api.get_content()`. 

`get_content` which is the method to get one content item. `get_content_items` is the index-endpoint to get a list of content items. I've updated the example to use the correct terminology.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -139,18 +139,15 @@ def request_json_matches(required_attributes)
         end
       end
 
-      def publishing_api_has_fields_for_document(format, items, fields)
+      def publishing_api_has_content(items, params = {})
```
Davidslv: Can you provide documentation on how to use this method please?
Follow the convention [used here](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L183) so we can read the documentation here: http://www.rubydoc.info/gems/gds-api-adapters

Thank you

tijmenb:
Note that the test failure here is unrelated to the code. I've fixed it in https://github.com/alphagov/gds-api-adapters/pull/485. Once that's merged you can rebase on master and this'll be green :frog:.

Rosa-Fox:
Thanks @tijmenb, we fixed it too! 12 minutes later! but thank you!! I have made the changes, let me know what you think.

tijmenb:
Looks good!

It does need a little test. You'll probably notice that `params.to_query` won't work because `active_support` isn't included in this project - it'll only work on Rails applications.

tijmenb:
Excellent! :ship:

PR #486, opened 2016-04-13
Improve documentation for publishing-api and rummager

tijmenb:
This PR adds some extra documentation to the publishing-api and rummager clients and updates the readme. 

It doesn't cover all the methods in Publishing API. Where documentation is missing I've added links to the publishing-api documentation in the repo. This may help new starters find their way around GOV.UK a little better.

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -1,9 +1,12 @@
 require_relative 'base'
 
+# Adapter for the Publishing API.
+#
+# @see https://github.com/alphagov/publishing-api
+# @see https://github.com/alphagov/publishing-api/blob/master/doc/publishing-application-examples.md
+# @see https://github.com/alphagov/publishing-api/blob/master/doc/object-model-explanation.md
 class GdsApi::PublishingApiV2 < GdsApi::Base
   # Put a content item
-  #
-  # This creates a new draft item, which will be sent to the draft content store.
   #
   # @param content_id [UUID]
   # @param payload [Hash] A valid content item
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -1,6 +1,14 @@
 require_relative 'base'
 
 class GdsApi::PublishingApiV2 < GdsApi::Base
+  # Put a content item
+  #
+  # This creates a new draft item, which will be sent to the draft content store.
+  #
+  # @param content_id [UUID]
+  # @param payload [Hash] A valid content item
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#put-v2contentcontent_id
   def put_content(content_id, payload)
     put_json!(content_url(content_id), payload)
   end
@@ -14,6 +22,7 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
   # @option params [String] locale The language, defaults to 'en' in publishing-api.
   #
   # @return [GdsApi::Response] a content item
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#get-v2contentcontent_id
   def get_content(content_id, params = {})
     get_json(content_url(content_id, params))
   end
@@ -29,6 +38,7 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
   # @return [GdsApi::Response] a content item
   #
   # @raise [HTTPNotFound] when the content item is not found
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#get-v2contentcontent_id
   def get_content!(content_id, params = {})
     get_json!(content_url(content_id, params))
   end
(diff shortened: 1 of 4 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -1,6 +1,14 @@
 require_relative 'base'
 
 class GdsApi::PublishingApiV2 < GdsApi::Base
+  # Put a content item
+  #
+  # This creates a new draft item, which will be sent to the draft content store.
```
tuzz: This also updates an existing content item.

tuzz:
Thanks for this, @tijmenb!

lib/gds_api/publishing_api_v2.rb
```diff
@@ -1,6 +1,14 @@
...
+  #
+  # @param content_id [UUID]
+  # @param payload [Hash] A valid content item
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#put-v2contentcontent_id
```
MatMoore: This is good to link to. There's also https://github.com/alphagov/publishing-api/blob/master/doc/publishing-application-examples.md
and https://github.com/alphagov/publishing-api/blob/master/doc/object-model-explanation.md which I've found useful. Do you think it makes sense to link these anywhere here?

lib/gds_api/publishing_api_v2.rb
```diff
@@ -1,6 +1,14 @@
 require_relative 'base'
 
 class GdsApi::PublishingApiV2 < GdsApi::Base
+  # Put a content item
+  #
+  # This creates a new draft item, which will be sent to the draft content store.
+  #
+  # @param content_id [UUID]
+  # @param payload [Hash] A valid content item
```
MatMoore: `A valid x` assumes I already know a lot about how the publishing platform works. Can we refer people to the content schemas and/or some examples?

Davidslv:
Something I find useful is to always have an example of the result/output of the method, like Elixir and Python have with [DocTest](http://elixir-lang.org/docs/stable/ex_unit/ExUnit.DocTest.html)

lib/gds_api/publishing_api_v2.rb
```diff
@@ -92,10 +124,28 @@ def discard_draft(content_id, options = {})
     post_json!(discard_url(content_id), params)
   end
 
+  # FIXME: Add documentation
```
MatMoore: FIXME :construction:

lib/gds_api/publishing_api_v2.rb
```diff
@@ -92,10 +124,28 @@ def discard_draft(content_id, options = {})
...
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#get-v2linkscontent_id
   def get_links(content_id)
     get_json(links_url(content_id))
   end
 
+  # Patch the links of a content item
+  #
+  # @param content_id [UUID]
+  # @param payload [Hash] A "links hash"
```
MatMoore: We should cover semantics:

* A links hash is a hash of tag type -> array of links
* The common tag types are: ...
* The array of tags is unordered, we use the "primary/secondary" pattern for x y z tag types
* When you leave out a tag type, existing tags of that type are unchanged
* When you set a tag type to an empty array, it deletes existing tags of that type
* What happens when the content id doesn't exist?

lib/gds_api/publishing_api_v2.rb
tijmenb: These `FIXME`'s are for anyone to pick up, in exchange for internet points and lots of happy emoji.

lib/gds_api/publishing_api_v2.rb
tijmenb: Ok now you're just creating work :)

I can link out to https://github.com/alphagov/publishing-api/blob/master/doc/semantics-of-the-publishing-api.md#links. @tuzz do you know about other places where `links` are documented?

lib/gds_api/publishing_api_v2.rb
tijmenb: The link below that should cover that I think?

lib/gds_api/rummager.rb
```diff
@@ -4,17 +4,32 @@
 module GdsApi
   class Rummager < Base
 
+    # Unified search
+    #
```
MatMoore: The names "Unified search" and "Advanced search" are outdated now. We changed all the internal code to just call it search and prefixed all the old stuff with "legacy" so its obvious its deprecated. Maybe we should alias the endpoints as well so we have a common language. (We shouldn't remove or rename anything without versioning the API though).

lib/gds_api/publishing_api_v2.rb
MatMoore: Well we don't have to do everything in one PR :) Feel free to assign to me!

lib/gds_api/rummager.rb
tijmenb: Good point, added:

https://trello.com/c/eqtFozNE/281-replace-unified-search-with-search-in-the-rummager-api-client

MatMoore:
:+1: Lots of good stuff here. I commented on a few things inline but don't treat these as a blocker for merge.

lib/gds_api/publishing_api_v2.rb
tijmenb: Removed the extra explanation, as this is better described in the publishing-api docs.

lib/gds_api/publishing_api_v2.rb
tijmenb: Great idea! I've added them to the class-level docs.

tijmenb:
Updated PR.

PR #490, opened 2016-04-15
Add header clearing

elliotcm:


tijmenb:
Can we add a little more context to the PR? Why was this added, what's it going to solve?

PR #493, opened 2016-04-25
Pin pact-related gems to known working versions

tijmenb:
`pact-mock_service` version 0.8.2 is causing test failures related to webmock. This pins all pact-related gems to known working versions so that tests in CI will pass.

Fixes test failures in https://github.com/alphagov/gds-api-adapters/pull/492 & master.

https://ci.dev.publishing.service.gov.uk/job/govuk_gds_api_adapters/


jackscotti:
👍

PR #494, opened 2016-04-26
Remove content_register class and test helpers.

danielroseman:
The application is being removed.


tijmenb:
👋  content register!

PR #496, opened 2016-05-04
Add notifications and notification endpoints...

steventux:
Exposes the endpoints `/notifications` and `/notifications/:id` from the email-alert-api,
these list recent notifications (aka GovDelivery bulletins) and show the detail of a notification respectively.
No subscriber information is revealed via these methods.

[Email Alert API endpoints this PR exposes](https://github.com/alphagov/email-alert-api/commit/72e0f98cddeb8e9f9f1b0607b27bd645710bce9a)

PR #498, opened 2016-05-06
Bump version to 30.3.1

steventux:
Update version to 30.3.1 because #496


tijmenb:
This should probably be 30.3.0.

PR #504, opened 2016-05-16
Add locale arg to unpublish endpoint...

steventux:
[This is used by the publishing api unpublish endpoint to scope content lookup](https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/unpublish.rb#L118).


elliotcm:
Is defaulting to "en" something we've done for other endpoint adapters?  The endpoint itself defaults to "en" if nothing is provided.

steventux:
Now you mention it, [the publishing-api does this for us](https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/unpublish.rb#L109) so I'll make it nil by default.

steventux:
@ecm updated.

elliotcm:
(I'm not ecm on here)

PR #505, opened 2016-05-17
Add support for get_expanded_links endpoint

tijmenb:
This endpoint was added in alphagov/publishing-api#307.


tijmenb:
Pact tests will pass after https://github.com/alphagov/publishing-api/pull/312 is merged.

Davidslv:
👍

PR #507, opened 2016-05-18
Remove implicit dependency between JsonClient and GdsApi::Base

floehopper:
Previously `JsonClient` used `GdsApi::Base.logger` as its default logger. However, there was no corresponding `require 'base'` statement at the top of the file. This caused a [problem in Smart Answers][1] which (for better or worse) currently uses `JsonClient` directly.

It's not possible to fix this by adding such a `require` statement, because this would introduce a circular dependency. So as an alternative, I've used `NullLogger.instance` as the default value, but at the same time ensured that `GdsApi::Base` always specifies a logger in the options passed to `JsonClient#initialize`. I'm pretty confident this means that the behaviour hasn't really changed and yet there is no longer a dependency between `JsonClient` and `GdsApi::Base`.

I don't know whether I'm doing something wrong, but running the tests with the default Rake task seemed very flakey i.e. I kept seeing a lot of test failures even before making any changes. However, every couple of runs, the tests seem to work OK and they all passed.

[1]: https://github.com/alphagov/smart-answers/pull/2538

### DIFF
--- a/lib/gds_api/base.rb
+++ b/lib/gds_api/base.rb
@@ -41,7 +41,8 @@ class GdsApi::Base
   def initialize(endpoint_url, options={})
     options[:endpoint_url] = endpoint_url
     raise InvalidAPIURL unless endpoint_url =~ URI::regexp
-    default_options = GdsApi::Base.default_options || {}
+    base_options = { logger: self.class.logger }
+    default_options = base_options.merge(GdsApi::Base.default_options || {})
     @options = default_options.merge(options)
     self.endpoint = options[:endpoint_url]
   end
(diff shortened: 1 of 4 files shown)


MatMoore:
This seems sensible to me. I'm assuming it's only SmartAnswers that uses `JsonClient` directly or we'd know about it by now?

floehopper:
@MatMoore: I've just discovered that [I can't reproduce the original problem in Smart Answers](https://github.com/alphagov/smart-answers/pull/2538#issuecomment-220924247) and was about to comment on this PR to say that it might not be necessary. However, I still think the change in this PR makes sense on its own. Thanks for merging.

PR #509, opened 2016-05-23
Allow passing a locale to patch_links

boffbowsh:
The PATCH links publishing api endpoint accepts a locale parameter so
that it can publish the correct locale to the content store. This
change passes the parameter through from the calling client.


boffbowsh:
Actually publishing-api should send content items for all locales during patch_links

PR #510, opened 2016-05-24
Add grouped content and links

MatMoore:
Add an API adapter for the new grouped-content-and-links publishing API endpoint.


MatMoore:
Closing for now - we'll use a branch until we've settled on the final form of the endpoint (name and response structure)

PR #512, opened 2016-05-26
Passthru bulk_publishing for patch_links

boffbowsh:
Setting this optional flag controls which priority jobs are given
within the publishing-api.

Uses https://github.com/alphagov/publishing-api/pull/329


tijmenb:
Ooh, that's very useful 👍

PR #514, opened 2016-05-26
Fix docs for patch_links

tijmenb:
The documentation for the arguments were wrong. Also adds docs for the `bulk_publishing` param. (https://github.com/alphagov/gds-api-adapters/pull/512).


boffbowsh:
👍

PR #515, opened 2016-05-31
[WIP] Update `areas` param handling

davidbasalla:
The `mapit_has_a_postcode_and_areas` helper allows to pass it a simplified `areas` array by converting it to the correct mapit data structure inside the helper method. The `mapit_has_areas` methods was not doing that, which made using the two helpers together awkward in tests. This brings the `mapit_has_areas` helper into line so that the two methods will convert the passed in `areas` arrays similarly.


davidbasalla:
Closed in favour of https://github.com/alphagov/gds-api-adapters/pull/524

PR #521, opened 2016-06-14
Document how to set a custom timeout

tuzz:
This may be controversial. I considered adding a better interface for setting the timeout rather than reaching inside the client's options hash. I can investigate this if people feel strongly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,24 @@ By default we log to a NullLogger since we don't want to pollute your test
 results or logs. To log output you'll want to set `GdsApi::Base.logger` to
 something that actually logs:
 
-    GdsApi::Base.logger = Logger.new("/path/to/file.log")
+```ruby
+GdsApi::Base.logger = Logger.new("/path/to/file.log")
+```
+
+## Setting the timeout
+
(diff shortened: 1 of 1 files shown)


boffbowsh:
Your documentation changes broke the build :trollface:

tuzz:
It's ok @boffbowsh, it passed on its fifth try. Seems legit.

README.md
```diff
@@ -32,7 +32,24 @@ By default we log to a NullLogger since we don't want to pollute your test
...
+
+## Setting the timeout
+
+By default the JsonClient timeout is set to 4 seconds. If this is exceeded a
+`GdsApi::TimedOutException` will be raised. Individual clients may decide to
+override this timeout. Alternatively, you can override this in the application
+that uses the adapter with:
+
+```ruby
+Services.publishing_api.client.options[:timeout] = number_of_seconds
```
tijmenb: You can also provide it to the client:

```ruby
publishing_api_client = GdsApi::PublishingApiV2.new(
  Plek.new.find('publishing-api'),
  timeout: 20,
)
```

Is that cleaner?

PR #524, opened 2016-06-20
Add a new MapIt method, `area_for_code_type`

davidbasalla:
- This is used in some work we're doing in `frontend`
  (https://github.com/alphagov/frontend/tree/use_authorities_from_mapit) to get
  areas data from MapIt directly:
  https://trello.com/c/YZEWeHwX/404-switch-frontend-to-get-la-slugs-from-mapit-instead-of-authorities-json-3.
- We expose the MapIt endpoint for `/code/<code_type>/<code>` as documented [in this merged MapIt PR](https://github.com/mysociety/mapit/pull/242)
- After [adding the CodeType 'govuk_slug' to MapIt](https://github.com/alphagov/mapit/pull/20), we need to update to `mapit_has_a_postcode_and_areas` test helper to include it

### DIFF
--- a/lib/gds_api/mapit.rb
+++ b/lib/gds_api/mapit.rb
@@ -12,6 +12,10 @@ class GdsApi::Mapit < GdsApi::Base
     get_json("#{base_url}/areas/#{type}.json")
   end
 
+  def area_for_code(code_type, code)
+    get_json("#{base_url}/code/#{code_type}/#{code}.json")
+  end
+
   class Location
     attr_reader :response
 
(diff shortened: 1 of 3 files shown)


lib/gds_api/mapit.rb
```diff
@@ -12,6 +12,10 @@ def areas_for_type(type)
     get_json("#{base_url}/areas/#{type}.json")
   end
 
+  def area_for_code_type(code_type, code)
```
jennyd: On naming, this is really getting the area for a code - `area_for_code_type` sounds as though it might get all areas for a code type as `areas_for_type` does, instead of the single area with the code of a given type. I think I'd find `area_for_code` clearer - what do you think?

davidbasalla: Good point @jennyd! Have updated and rebased

PR #525, opened 2016-06-23
Expand YARD docs for gds_api/asset_manager

rgarner:
* Add caveat about deriving content type from filename (some systems
  are passing a content type and seeing things work because the filename
  happens to match. Be explicit that we're only concerned with the
  filename)
* Add examples
* Change vanilla HttpResponse to GdsApi::Response for return types
* Add a list of attributes currently returned


tuzz:
This is much much better. Thank you for this.

PR #526, opened 2016-06-29
[Discuss, don't merge] Always raise when publishing-api & content-store return not found

tijmenb:
This commit makes sure that the publishing-api and content-store clients raise a `GdsApi::NotFound` error when the server returns a 404 (Not Found) or 410 (Gone). 

API calls won't return `nil` anymore.

This was previously only the case for some methods. In some cases a pattern was to have a `bang!` method and a non-`bang!` method, so that the client could opt-in to this behaviour. But clients like Rummager use the `get_json!` method exclusively.

Reasons to change:

- This makes the behaviour consistent across the most used clients (publishing-api, content-store and rummager)
- Less duplication between `bang!` and `non_bang` methods.
- Makes it less likely that we inadvertently introduce nils, which generate errors that are hard to spot and to debug.

**Don't merge**: I'd like feedback about this first, as it's quite a backwards-incompatible change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 Please use this place to add information of work that hasn't been released,
 and specify if that is backwards-compatible.
 
+* Publishing API client always raises when it encounters a 404. *NOT BACKWARDS COMPATIBLE*
+
 # 31.2.0
 
 * Add an `area_for_code` method to the MapIt API adapter.
--- a/lib/gds_api/content_store.rb
+++ b/lib/gds_api/content_store.rb
@@ -9,10 +9,6 @@ class GdsApi::ContentStore < GdsApi::Base
     end
   end
 
-  def content_item(base_path)
-    get_json(content_item_url(base_path))
-  end
-
   def incoming_links!(base_path, params = {})
     query = query_string(params)
     get_json!("#{endpoint}/incoming-links#{base_path}#{query}")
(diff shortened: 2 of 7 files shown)


tuzz:
Are there legitimate cases to make a request for a content item that doesn't exist?

I agree in principle that it's better to raise errors rather than return nils which can cause other bugs. I'm just concerned people might start using `begin ... rescue` as control flow if there's no alternative.

thehenster:
This change would be consistent with the non-banged `put_content` method too, which raises exceptions. So I approve :+1:

But if I were king of the world my preference would be to not raise exceptions for any of these calls for the reason @tuzz mentions.

MatMoore:
If I was king of the world nil wouldn't exist 👑 

I strongly prefer using exceptions by default, as if they go unhandled, they don't mask the source of the error. I think this PR makes the library more consistent and less error prone. 👍 

@tuzz that seems like a valid use case. I'm not sure if the `nil` methods actually handle this well though - does the non-bang version actually distinguish between 404s and other kinds of HTTP error? If it's something we still want to handle maybe we could have a method specifically for this, e.g.  `get_content_if_exists`?

thehenster:
You can be king :) Nil is bad.

It's a case of picking your poison isn't it.. Nil bugs are super annoying. But a lack of defensive error handling, which exceptions encourage, introduces a breed of bugs itself. :shrug: Golang went as far to remove them from the language..

I think I have just seen similar types of client libraries lull developers into a false security. Despite the endpoints being ours, sometimes they should be treated with as much suspicion as a payment gateway API or the like.

benhyland:
Given the choice of error handling mechanism, I rank as folows:
0. Error impossible by construction
1. Sum types
13. Checked exceptions
17. Unchecked (or untyped!) exceptions
99998. nil

👍

tuzz:
👍  from me.

danielroseman:
👍

tijmenb:
Closing this temporarily, I'd like to try something with deprecating the old behaviour so that the upgrade path is smoother.

PR #530, opened 2016-07-01
Add a param for 'allow_draft' unpublishing

tuzz:
This allows content items to transition from draft to
the unpublished state without being published. See:

https://github.com/alphagov/publishing-api/blob/master/doc/publishing-api-syntactic-usage.md#post-v2contentcontent_idunpublish


elliotcm:
Could you add something to the commit message explaining what this is for?

PR #533, opened 2016-07-06
Use rummager endpoint instead of search

36degrees:
Search is being deprecated as we move towards having vhosts matching application names.

The readme for gds-api-adapters suggests using `Plek.new.find('rummager')` when creating Rummager adapters, so the tests should support that case.

This will break tests for any applications who currently call `Plek.new.find('search')` when creating Rummager adapters. They will need to update their applications to use the rummager endpoint instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Unreleased
 
-- Provide a warning message for rummager deprecated stubs
-
 Please use this place to add information of work that hasn't been released,
 and specify if that is backwards-compatible.
+
+* Provide a warning message for rummager deprecated stubs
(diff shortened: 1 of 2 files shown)


Davidslv:
Can you add the change to the Unreleased section of the Changelog.md please?

36degrees:
@Davidslv Done!

Davidslv:
Thanks.
I'm afraid you had to rebase, and can you also add that is not backwards compatible please?
It's easier for the person who will bump the gem version to have that information to know what will be the next one.

36degrees:
Happy with this changelog entry?

PR #536, opened 2016-07-11
Allow base_path lookups for a single given state

rgarner:
* Gives us a way to see what existing item is causing a collision in
  dfid-transition
* Don't allow multiple state query, as some combinations (e.g.
  live/draft) are inadvisable

## Related PRs

* [ ] https://github.com/alphagov/publishing-api/pull/410


MatMoore:
👍  to merge as long as the corresponding publishing-api PR is merged.

PR #537, opened 2016-07-12
Fix flakey tests on Jenkins

tijmenb:
Depending on the order of tests, global state was being set
on the GovukHeaders singleton, which then interfered with
the pact tests that were checking the content type header.


tijmenb:
Merged my own PR because it's @tuzz's commit.

PR #539, opened 2016-07-13
Release 32.0.0

elliotcm:
Releases https://github.com/alphagov/gds-api-adapters/pull/538

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
-# Unreleased
+# 32.0.0
 
-Please use this place to add information of work that hasn't been released,
-and specify if that is backwards-compatible.
-
+* Allow publishing apps to request a specific version of content.
 * Provide a warning message for rummager deprecated stubs
 * Modify Rummager test helper to use `rummager` vhost instead of `search`.
   **This is a breaking change**, which means applications currently using the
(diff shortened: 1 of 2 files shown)


boffbowsh:
Why not 32.0.0?

elliotcm:
While probably the right thing to do it felt weird releasing someone else's breaking change as though it was part of my work, even though that's effectively what I'm doing anyway.

elliotcm:
@boffbowsh your opine?

36degrees:
Either way, might be worth having something documenting how releases work for this repo as I wasn't sure if a) I should create a release for my previous changes myself and b) how I'd even begin to do that.

elliotcm:
@36degrees Releasing gem versions is detailed here: https://github.com/alphagov/styleguides/blob/master/rubygems.md

Though it isn't particularly clear about when how etc.  There's nothing special about this gem vs our others so perhaps we should beef up that documentation, unless someone knows of other docs somewhere else (I couldn't find anything in the opsmanual).

36degrees:
It'd be good to get this release out soon as we have some downstream stuff waiting on the changes that I made (which I now know I should have just released myself – live and learn 😄)

FWIW, I'd expect the next release to be 32.0.0.

boffbowsh:
Yeah I'd expect 32.0.0, else I'd wonder where that release went.

elliotcm:
Done.

36degrees:
LGTM 👍

elliotcm:
@36degrees hit that butan

PR #540, opened 2016-07-26
new publishing_api_has_item_in_sequence stub method

Seraphiana:
added a method that stubs the publishing API to return several (different) results in order that they are queried


tuzz:
A bit more context:

We need this method in order to test Specialist Publisher's publish workflow:
https://github.com/alphagov/specialist-publisher-rebuild/blob/5ad752a4868c595c02904c1e95918e40e1d3b70d/spec/features/publishing_a_cma_case_spec.rb#L55

In the above example, we stub the Publishing API such that the first time we make a GET request we return the draft item and the second time we return the published item. This makes the test consistent with how the Publishing API behaves when the publish command is called and lets us test the correct data is sent to other downstream systems, e.g. rummager and email-alert-api.

PR #542, opened 2016-07-27
Add API adapter for Local Links Manager

davidbasalla:
Added #local_link for Local Links Manager

[Corresponding PR](https://github.com/alphagov/local-links-manager/pull/44)

[Trello card](https://trello.com/c/yLMi98ze/428-3-of-4-start-using-data-from-local-links-manager-for-local-transactions-in-production)

### DIFF
--- a/lib/gds_api/local_links_manager.rb
+++ b/lib/gds_api/local_links_manager.rb
@@ -1,0 +1,9 @@
+require_relative 'base'
+
+class GdsApi::LocalLinksManager < GdsApi::Base
+  def local_link(authority_slug, lgsl, lgil=nil)
+    url = "#{endpoint}/api/link?authority_slug=#{authority_slug}&lgsl=#{lgsl}"
+    url += "&lgil=#{lgil}" if lgil
+    get_json(url)
+  end
+end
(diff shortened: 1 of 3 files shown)


davidbasalla:
Flagging as DNM because we realise we're missing at least one useful test helper.

davidbasalla:
Added missing test helpers, ok to review/merge again.

sihugh:
Looks good!  Should we have something that covers the `400` / `404` response from Local-links-manager?  I know we're going to check in Frontend that the `authority_slug` exists in MapIt for example, but it might not be in Local-links-manager for some reason.

davidbasalla:
@sihugh We've added tests for the 404 case when LocalLinksManager can't find the LocalAuthority/Service/ServiceInteraction.

Do we think we need to add a case for the 400 error (missing parameters) as well, since we're unlikely to do this from Frontend? We've had difficulties testing with setting query params such as 'authority_slug' to `nil`. In our test we create the following stub:
```
http://locallinksmanager.dev.gov.uk/api/link?authority_slug&lgsl=2 
```
whereas the request URL that we generate looks like this (extra '=' sign):
```
http://locallinksmanager.dev.gov.uk/api/link?authority_slug=&lgsl=2
```

We haven't been able to get around it so far :/

h-lame:
@davidbasalla - does the stub create the url with the "missing equals sign" if you set the authority_slug to the empty string instead of `nil`?

davidbasalla:
@sihugh @h-lame We've added the tests for the 400/missing parameters case now, could you re-review :)

sihugh:
:+1:

PR #544, opened 2016-07-29
Add option to always raise for 404s

tijmenb:
This commit adds an option to ensure that the publishing-api and content-store clients raise a `GdsApi::NotFound` error when the server returns a 404 (Not Found) or 410 (Gone).

This was previously only the case for some methods. In some cases a pattern was to have a `bang!` method and a non-`bang!` method, so that the client could opt-in to this behaviour. Clients like Rummager use the `get_json!` method exclusively.

Reasons to change:

- This makes the behaviour consistent across the most used clients (publishing-api, content-store and rummager)
- Less duplication between `bang!` and `non_bang` methods.
- Makes it less likely that we inadvertently introduce nils, which generate errors that are hard to spot and to debug.

To keep backwards compatibility, we make this opt-in behaviour, providing a friendlier upgrade path.

Follow up from https://github.com/alphagov/gds-api-adapters/pull/526

## Usage notes

Add this to opt-in to the new behaviour:

```ruby
# config/initializers/gds_api_adapters.rb
GdsApi.configure do |config|
  config.always_raise_for_not_found = true
end
```

### DIFF
--- a/lib/gds_api/config.rb
+++ b/lib/gds_api/config.rb
@@ -1,0 +1,17 @@
+module GdsApi
+  def self.configure
+    yield(config)
+  end
+
+  def self.config
+    @config ||= Config.new
+  end
+
+  class Config
+    # Always raise a `HTTPNotFound` exception when the server returns 404 or
+    # 410. This avoids nil-errors in your code and makes debugging easier.
+    #
+    # Currently defaults to false, but will be enabled by default in the future.
+    attr_accessor :always_raise_for_not_found
+  end
+end
--- a/lib/gds_api/config.rb
+++ b/lib/gds_api/config.rb
@@ -11,7 +11,11 @@ module GdsApi
     # Always raise a `HTTPNotFound` exception when the server returns 404 or
     # 410. This avoids nil-errors in your code and makes debugging easier.
     #
-    # Currently defaults to false, but will be enabled by default in the future.
+    # Currently defaults to false.
+    #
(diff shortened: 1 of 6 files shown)


lib/gds_api/config.rb
```diff
@@ -0,0 +1,17 @@
+module GdsApi
+  def self.configure
+    yield(config)
+  end
+
+  def self.config
```
MatMoore: Why is this a global config rather than an option for the client instance?

tijmenb: Because then you would have to set it for all clients. Is that something you would prefer?

MatMoore: Generally yeah, because it's more obvious what's going on when inputs are explicitly declared. If it can be changed anywhere in the app then you have to look in multiple places to know how the client will behave.

Not a massive issue if it's temporary code though.

tijmenb: I don't think it's helpful to allow this setting per-client. That would only increase the complexity. You would configure this in an initializer, there's no reason to change this anywhere in the app.

tijmenb:
@MatMoore I've added a deadline for the deprecation and added a deprecation notice if the client is using the deprecated methods.

lib/gds_api/config.rb
tijmenb: Also added a note to the description on how to use this.

tijmenb:
https://github.com/alphagov/content-tagger/pull/99 is using this now.

PR #545, opened 2016-07-29
Add option to return hash from all requests

tijmenb:
The `hash_response_for_requests` config will configure `GdsApi::Response` so that it won't return OpenStruct objects, but will behave like a normal hash. This will prevent nil-errors.

Some background by @tuzz: https://trello.com/c/GHlFpkuw

Related: https://github.com/alphagov/gds-api-adapters/pull/544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 * Add option to always raise for 404 and 410.
+* Add option to make `GdsApi::Response` just behave like a hash, not an OpenStruct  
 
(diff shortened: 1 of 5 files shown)


h-lame:
There is not a 👍 big enough for this.

MatMoore:
👍 

Can we phase out the old behaviour too?

tijmenb:
https://github.com/alphagov/content-tagger/pull/99 is using this now.

PR #558, opened 2016-08-15
Make get raw symmetrical with get json

h-lame:
`get_raw` didn't have the same toggle for `always_raise_for_not_found` that the other `<verb>_json` methods had, so we add it to make them behave the same.


tijmenb:
👍

PR #559, opened 2016-08-17
Replace dead links with less dead links

MatMoore:
💀 🔗


tijmenb:
Note that this intentionally linked to the branch to keep the docs up to date: https://github.com/alphagov/gds-api-adapters/pull/497

I think we need to start hosting this thing ourselves, rubydoc.info is quite unpredictable.

PR #560, opened 2016-08-17
Revive and extend helpers for rummager /document/ API

MatMoore:
This PR is best reviewed commit by commit: the first one un-deprecates and documents adapter methods, and the second one changes the behaviour of the corresponding test helpers. Both relate to the rummager documents API.

## Why we're reviving `delete_document`

`delete_content!` is not a substitute for `delete_document`:

- this endpoint doesn't work for content without base paths
- it doesn't work for recommended-links (urls that may be external to gov.uk)
- it doesn't handle things stored in the metadata index, such as best bets

I considered creating more specific methods to replace `delete_document`, but I don't think this is worth doing. There are already multiple apps using `delete_document`:
https://github.com/search?utf8=%E2%9C%93&q=org%3Aalphagov+delete_document&type=Code&ref=searchresults

In my opinion `/document/` is a more useful API. `delete_document` complements `add_document`, which together cover the primary things a publishing app needs to do with rummager. The `/content/` API only supports GET and DELETE.

I've also taken the opportunity to expand the method documentation a bit.

It might also be worth aliasing methods like `add_content!` to `add_content`, given that the use of `!` in this class is inconsistent and they all raise exceptions in the same way. Thoughts?

This was required for https://github.com/alphagov/search-admin/pull/48
Trello: https://trello.com/c/WY4JGxKG/39-manage-external-recommended-links-in-search-admin

### DIFF
--- a/lib/gds_api/test_helpers/rummager.rb
+++ b/lib/gds_api/test_helpers/rummager.rb
@@ -6,18 +6,31 @@ module GdsApi
     module Rummager
       RUMMAGER_ENDPOINT = Plek.current.find('rummager')
 
-      def stub_any_rummager_post
-        stub_request(:post, %r{#{RUMMAGER_ENDPOINT}/documents})
+      def stub_any_rummager_post(index: nil)
+        if index
+          stub_request(:post, %r{#{RUMMAGER_ENDPOINT}/#{index}/documents})
+            .to_return(status: [202, "Accepted"])
+        else
+          stub_request(:post, %r{#{RUMMAGER_ENDPOINT}/documents})
+            .to_return(status: [202, "Accepted"])
+        end
       end
 
       def stub_any_rummager_post_with_queueing_enabled
+        warn "stub_any_rummager_post_with_queueing_enabled is deprecated: use stub_any_rummager_post instead"
+
         stub_request(:post, %r{#{RUMMAGER_ENDPOINT}/documents}) \
           .to_return(status: [202, "Accepted"])
       end
 
-      def assert_rummager_posted_item(attributes)
-        url = RUMMAGER_ENDPOINT + "/documents"
-        assert_requested(:post, url) do |req|
+      def assert_rummager_posted_item(attributes, index: nil, **options)
+        if index
+          url = RUMMAGER_ENDPOINT + "/#{index}/documents"
+        else
+          url = RUMMAGER_ENDPOINT + "/documents"
+        end
+
+        assert_requested(:post, url, **options) do |req|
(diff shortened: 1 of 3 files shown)


MatMoore:
I'm going to do some more work on the test helpers for rummager as well which I'll roll into this PR. 

- The existing stubs and asserts do not fully cover the `/documents/` API since they miss out the optional `index` part of the URL
- There are two stubs for posting a document: with and without queueing (200 vs 202 responses). Rummager always queues now so we shouldn't be pretending otherwise in our tests

MatMoore:
I've completed the above changes.

This PR is best reviewed commit by commit: the first one un-deprecates and documents adapter methods, and the second one changes the behaviour of the corresponding test helpers. Both relate to the [rummager documents API](https://github.com/alphagov/rummager/blob/master/docs/documents.md).

lib/gds_api/test_helpers/rummager.rb
```diff
@@ -6,47 +6,76 @@ module TestHelpers
...
 
-      def assert_rummager_posted_item(attributes)
-        url = RUMMAGER_ENDPOINT + "/documents"
-        assert_requested(:post, url) do |req|
+      def assert_rummager_posted_item(attributes, index: nil, **options)
+        if index
+          url = RUMMAGER_ENDPOINT + "/#{index}/documents"
+        else
+          url = RUMMAGER_ENDPOINT + "/documents"
+        end
```
tuzz: What sort of thing would you pass as an `index` argument? When would you explicitly set an index vs. posting to `/documents` ?

tuzz:
👍  makes sense.

lib/gds_api/test_helpers/rummager.rb
MatMoore: If you post to `/documents`, rummager will use the default index (mainstream I think).

But rummager has multiple indexes - for example, whitehall content goes to `government` (whitehall doesn't use gds-api-adapters for this yet) and best bets go to a special metadata index.

MatMoore:
@tuzz Are you happy with me raising the major version for this change? If so I'll rebase to fix the conflicts.

tuzz:
@MatMoore yep, happy! 👍

lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '33.2.2'
+  VERSION = '34.0.0'
```
tijmenb: (For future reference, versions should be bumped in a separate PR).

PR #561, opened 2016-08-22
More correct headers.

elliotcm:
According to HTTP RFC 2616, GET and DELETE request
bodies should be ignored, so we shouldn't be using
them.

Sending a Content-Type header with a bodyless
request makes no sense and causes some servers to
fail as they try to JSON parse an empty string
(which is invalid).

This change removes the Content-Type header from
GET and DELETE requests and deprecates sending
a body with a DELETE request (currently only used
by the need API where a POST would be more appropriate).


alexmuller:
This all looks correct to me but I'd like somebody with more experience of api-adapters to review it too.

deanwilson:
The ruby code looks fine to me.

PR #564, opened 2016-08-22
Release version 33.2.2

alexmuller:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 33.2.2
+
+* Require `config` in JsonClient
```
boffbowsh: This sounds like a new requirement that clients need to fulfil, which sounds like a breaking change. "Fix JsonClient not explicitly requireing the config" ?

alexmuller: Thanks, updated!

PR #565, opened 2016-08-22
Better headers in publishing API pacts.

elliotcm:
This sends a more accurate set of headers into
the publishing API during pact tests.

Because this adds a JSON "Accept" header to the
request, GDS SSO starts to correctly treat the
request as an API request and starts requiring a
bearer token.  So we provide that too.

Requires https://github.com/alphagov/publishing-api/pull/474 to be merged first.


alexmuller:
Makes our tests look more like production :+1:

PR #566, opened 2016-08-25
Add /search.json method

MatMoore:
This lets gds-api-adapters work with both external and internal endpoints.

Rummager now listens at both `unified_search.json` and `search.json`. The former can be removed once nothing is using it anymore. The public search endpoint has always been `search.json`.

https://trello.com/c/cj8UX2jX

### DIFF
--- a/lib/gds_api/rummager.rb
+++ b/lib/gds_api/rummager.rb
@@ -6,6 +6,7 @@ module GdsApi
 
     # Perform a search.
     #
+    # @deprecated Alias for `#search`.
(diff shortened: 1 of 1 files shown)


lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '34.0.0'
+  VERSION = '34.1.0'
```
tijmenb: Version updates need to go in separate PRs.

PR #569, opened 2016-09-01
Add test_helper for gone content item

davidbasalla:
Now that 'not found' and 'gone' content items raise exceptions (see https://github.com/alphagov/gds-api-adapters/commit/7e17dd8ae2bc41bf38671f7a68776440a51b0bc8), it is useful to have a test helper for gone items to mimic the correct response.


davidbasalla:
Added 'Do not merge' as I think the helper needs to return the content item (not `nil`) when  `always_raise_for_not_found` is set to `false`, so some work to do still

PR #572, opened 2016-09-06
Return pagination results from publishing_api_has_content

carvil:
A request to `/v2/content` returns pagination information besides the results.
The keys available on the response are:

```
["total", "pages", "current_page", "links", "results"]
```

This commit adds pagination-specific information to the response so we closely
match what a production request would look like.

The values are dummy for now, but we could potentially build logic to return
more appropriate results in the future in case that's important.

This information will be useful in content-tagger as we are interested in
pagination information.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -169,7 +169,20 @@ module GdsApi
       #)
       def publishing_api_has_content(items, params = {})
         url = PUBLISHING_API_V2_ENDPOINT + "/content"
-        stub_request(:get, url).with(:query => params).to_return(status: 200, body: { results: items }.to_json, headers: {})
+        per_page = params.fetch(:per_page, 50)
+        page = params.fetch(:page, 1)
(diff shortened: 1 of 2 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -169,7 +169,11 @@ def request_json_matches(required_attributes)
       #)
       def publishing_api_has_content(items, params = {})
         url = PUBLISHING_API_V2_ENDPOINT + "/content"
-        stub_request(:get, url).with(:query => params).to_return(status: 200, body: { results: items }.to_json, headers: {})
+        body = { results: items, total: 1, pages: 1, current_page: 1 }
```
tijmenb: Should we use `items.size` for the `total`?

carvil: We could do that, yes. Also, since we can send a specific `page` param, I can check for it and change `current_page` accordingly. Does that make sense?

carvil:
@tijmenb does this make more sense now?

PR #573, opened 2016-09-06
Change Publishing API's port in tests

carvil:
We were using port `3093` in the tests for a mocked version of the Publishing
API. That is also the port used in development mode. This means that if we run the
publishing api in development mode and run the gds-api-adapters' tests, we get a
bunch of failures, as the code hits the actual Publishing API endpoints.

This commit changes the port of the test Publishing API and makes it easier to
change it later on by using a method to return the localhost URL of the
publishing API.

### DIFF
--- a/test/test_helpers/pact_helper.rb
+++ b/test/test_helpers/pact_helper.rb
@@ -1,7 +1,13 @@
+PUBLISHING_API_PORT=3001
+
+def publishing_api_host
+  "http://localhost:#{PUBLISHING_API_PORT}"
+end
+
 Pact.service_consumer "GDS API Adapters" do
   has_pact_with "Publishing API" do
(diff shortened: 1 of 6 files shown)


test/test_helpers/pact_helper.rb
```diff
@@ -1,7 +1,15 @@
+def publishing_api_port
```
tijmenb: Would it be worth it to make this a constant so it's obvious that it's not dynamic?

tijmenb:
Nice detective work 👍

test/test_helpers/pact_helper.rb
carvil: Yup, will change it now.

carvil:
@tijmenb done, it's now a constant.

PR #576, opened 2016-09-09
Support api review spam

NeilvB:
The support API has two new endpoints. These are used to fetch lists of problem reports and to update specified reports as spam.

### DIFF
--- a/lib/gds_api/support_api.rb
+++ b/lib/gds_api/support_api.rb
@@ -47,4 +47,50 @@ class GdsApi::SupportApi < GdsApi::Base
   def feedback_export_request(id)
     get_json!("#{endpoint}/anonymous-feedback/export-requests/#{id}")
   end
+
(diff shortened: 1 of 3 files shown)


lib/gds_api/support_api.rb
```diff
@@ -47,4 +47,13 @@ def organisation(organisation_slug)
   def feedback_export_request(id)
     get_json!("#{endpoint}/anonymous-feedback/export-requests/#{id}")
   end
+
+  def problem_reports(options = {})
```
Davidslv: can you document this please?
Example: https://github.com/alphagov/gds-api-adapters/pull/581/files#diff-c279de60f2b72b505bbabe7401804a88R98

NeilvB: Sure thing @Davidslv  - please see updated comments in `lib/gds_api/support_api.rb`

PR #579, opened 2016-09-15
Support API create page improvement

thehenster:
A helper to access the new endpoint on the Support API added here: https://github.com/alphagov/support-api/pull/96

### DIFF
--- a/lib/gds_api/support_api.rb
+++ b/lib/gds_api/support_api.rb
@@ -21,6 +21,26 @@ class GdsApi::SupportApi < GdsApi::Base
     post_json!("#{endpoint}/anonymous-feedback/global-export-requests", global_export_request: request_details)
   end
 
+  # Create a Page Improvement
+  #
+  # Makes a +POST+ request to the support api to create a Page Improvement.
+  #
+  # @param params [Hash] Any attributes that relate to a Page Improvement.
+  #
+  # @return [GdsApi::Response] The wrapped http response from the api. Responds to the following:
+  #     :status       a string that is either 'success' or 'error'
+  #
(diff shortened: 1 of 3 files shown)


lib/gds_api/support_api.rb
```diff
@@ -21,6 +21,10 @@ def create_global_export_request(request_details)
     post_json!("#{endpoint}/anonymous-feedback/global-export-requests", global_export_request: request_details)
   end
 
+  def create_page_improvement(params)
```
Davidslv: Can you please document it?
Example: https://github.com/alphagov/gds-api-adapters/pull/581/files#diff-c279de60f2b72b505bbabe7401804a88R98

thehenster: Thanks. I've added some documentation of the same style.

lib/gds_api/support_api.rb
```diff
@@ -21,6 +21,26 @@ def create_global_export_request(request_details)
     post_json!("#{endpoint}/anonymous-feedback/global-export-requests", global_export_request: request_details)
   end
 
+  # Create a Page Improvement
+  #
+  # Makes a +POST+ request to the support api to create a Page Improvement.
+  #
+  # @param params [Hash] Any attributes that relate to a Page Improvement.
```
Davidslv: YARD seems to accept the type before the name of the attribute (`@param [Hash] params description` - http://www.yardoc.org/features.html)

thehenster: It looks like the type can be before or after the arguments. The docs generate correctly. As the rest of the YARD annotations in this repo have the arguments first I favour keeping it this way. Are you looking for a change here?

PR #584, opened 2016-09-26
Release version 36.1.1

NeilvB:
This release adds helpers for the support api for fetching problem reports and
for marking problem reports as spam.

### DIFF
--- a/lib/gds_api/version.rb
+++ b/lib/gds_api/version.rb
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '36.0.1'
+  VERSION = '36.1.0'
 end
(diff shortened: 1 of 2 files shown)


lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '36.0.1'
+  VERSION = '36.1.1'
```
tijmenb: This should probably be `36.1.0`

tijmenb:
Could you add something to the changelog as well?

lib/gds_api/version.rb
NeilvB: Yes, of course. Thanks

NeilvB:
Closed as I've opened a new PR with renamed branch (https://github.com/alphagov/gds-api-adapters/pull/585)

PR #586, opened 2016-09-26
Adds documentation on publishing api test helpers

Davidslv:
This commit adds documentation to the following endpoints:

- links
- expanded links

Also adds `hash_with_indifferent_access` on `expanded links` stub to
keep consistency.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -230,12 +230,93 @@ module GdsApi
         stub_request(:get, url).to_return(status: 404, body: resource_not_found(content_id, "content item").to_json, headers: {})
       end
 
+      # Stubs a request to links endpoint
+      #
+      # @param [Hash] links the structure of the links hash
+      #
+      # @example
+      #
+      #  publishing_api_has_links(
+      #    {
+      #      "content_id" => "64aadc14-9bca-40d9-abb6-4f21f9792a05",
+      #      "links" => {
+      #        "mainstream_browse_pages" => ["df2e7a3e-2078-45de-a75a-fd37d027427e"],
+      #        "parent" => ["df2e7a3e-2078-45de-a75a-fd37d027427e"],
+      #        "organisations" => ["569a9ee5-c195-4b7f-b9dc-edc17a09113f", "5c54ae52-341b-499e-a6dd-67f04633b8cf"]
+      #      },
+      #      "version" => 6
+      #    }
+      #  )
+      #
+      # Services.publishing_api.get_links("64aadc14-9bca-40d9-abb6-4f21f9792a05")
+      # => {
+      #      "content_id" => "64aadc14-9bca-40d9-abb6-4f21f9792a05",
+      #      "links" => {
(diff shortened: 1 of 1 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -230,13 +230,50 @@ def publishing_api_does_not_have_item(content_id)
...
+      #           "content_id" => "hi",
+      #           "title" => "Hi",
+      #           "base_path" => "/path/hi",
+      #           "document_type" => "guidance"
+      #         },
+      #       ]
+      #     }
+      #   )
       def publishing_api_has_expanded_links(links)
+        links = links.with_indifferent_access
```
tijmenb: We shouldn't use `with_indifferent_access` here, it's not certain this is used in a Rails app.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -230,13 +230,50 @@ def publishing_api_does_not_have_item(content_id)
         stub_request(:get, url).to_return(status: 404, body: resource_not_found(content_id, "content item").to_json, headers: {})
       end
 
+      # Stubs a request to links endpoint
+      #
+      # Requires a hash as paramenter
```
tijmenb: `@param links [Hash]` already tells that this is a hash.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -230,13 +230,50 @@ def publishing_api_does_not_have_item(content_id)
...
+      #
+      # @param links [Hash]
```
tijmenb: I think YARD likes this in the form `@param [Type] var_name And then a description`

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -230,13 +230,50 @@ def publishing_api_does_not_have_item(content_id)
...
+      #
+      # @example
+      #
+      #   publishing_api_has_links(
```
tijmenb: Perhaps use symbol-keys consistently here.

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -230,13 +230,50 @@ def publishing_api_does_not_have_item(content_id)
...
       end
 
+      # Stubs a request to expanded links endpoint
+      #
+      # Requires a hash as parameter
+      #
+      # @param links [Hash]
+      #
+      # @example
+      #   publishing_api_has_expanded_links(
```
tijmenb: The example is more useful if you provide the typical usage too:

Eg:

```
publishing_api_has_expanded_links(...foo...)
Services.publishing_api.expanded_links(content_id)
#=> ...foo...
```

lib/gds_api/test_helpers/publishing_api_v2.rb
Davidslv: Ah! but we do use it on `publishing_api_has_links` method

Davidslv:
I've updated it, thanks

PR #590, opened 2016-09-28
Update CONTRIBUTING.md

Davidslv:
We seem to be receiving some inconsistencies in some recent Pull Requests (PR).

I would like to make it clearer what we currently expect a PR to have so that everyone knows and can create a PR that respects our expectations.

This should help speed up the reviewing process and prevent a PR to be merged without the following:

- Readable code
- Documented endpoints
- Changes added to [CHANGELOG.md](https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,20 @@
+## TL; DR ##
+
+When you open a Pull Request we expect you to deliver the following:
+
+- Readable code
+- Documented endpoints
+- Changes added to [CHANGELOG.md](CHANGELOG.md)
+
 ## Git workflow ##
 
 - Pull requests must contain a succinct, clear summary of what the user need is driving this feature change.
 - Follow our [Git styleguide](https://github.com/alphagov/styleguides/blob/master/git.md)
 - Make a feature branch
 - Ensure your branch contains logical atomic commits before sending a pull request - follow our [Git styleguide](https://github.com/alphagov/styleguides/blob/master/git.md)
-- Pull requests are automatically integration tested, where applicable using [Travis CI](https://travis-ci.org/), which will report back on whether the tests still pass on your branch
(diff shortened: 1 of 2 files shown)


tijmenb:
I think we should get rid of all CONTRIBUTING.md, and link to a common one. It doesn't quite make sense to keep them up to date across all hundreds of repos.

Davidslv:
There's a risk nobody will ever read it, and some parts of if will not apply for repo X or repo Y, which will make it even more confusing to read

thehenster:
In my experience people very rarely click on CONTRIBUTING.md. How about linking to the common one from the README.md and including any amendments just above it.

eg.

```
# Contributing

* Document endpoints
* See [common GOV.UK Contributing guideliness](link)
```

MatMoore:
@thehenster's proposal makes sense to me. Putting the common guidelines in one place gets more eyes on it.

I'd also like to see examples of how to do the thing. "Readable" has a very broad meaning and can mean different things to different people, and other than the minor stylistic stuff we catch with the linter I'm not sure what kind of contribution we are trying to prevent.
 
IMO the purpose of contribution guidelines is to lower the barrier to contributing as much as possible. I've personally done the wrong thing for this repo due to ignorance and what would have been useful in a TLDR would be very straightforward guidelines like:

- Always add a changelog, here is an example you can copy and paste...
- Version changes go in a separate PR... this is what we mean by semantic versioning...
- How to generate the documentation

Davidslv:
@thehenster I like that proposal, I think it makes sense. The question I have is, where will the base one be hosted? (guidelines repo?)

@MatMoore when I wrote "readable" I had the ruby code style guide in mind, I think that's what makes sense otherwise I don't see why we would bother having one.
I also think your bullet points make sense to be added

MatMoore:
@Davidslv ah right, in that case I guess we could just link through the styleguide?

tijmenb:
Let's talk about a common `CONTRIBUTING.md` soon!

PR #592, opened 2016-09-29
Remove search-related fields from Panopticon Registerer

mobaig:
Client apps that use the Panopticon Registerer have been updated to send
their search payloads to Rummager directly for indexing. We no longer
need fields that are specifically related to Panopticon search
registration behaviour.


carvil:
👍

PR #601, opened 2016-10-04
Add helper methods to stub and assert searches

36degrees:
By not definining a default response (`to_return`) as part of the main stubbing method, this allows you to chain off `stub_any_rummager_search` to specify a suitable return value in your application’s tests.

You can use `stub_any_rummager_search_to_return_no_results` when you do not care about what Rummager will return.

Note that using stub_any_rummager_search _without_ chaining a call to `to_return` will probably result in errors as the Rummager class is expecting valid JSON to be returned.

### DIFF
--- a/lib/gds_api/test_helpers/rummager.rb
+++ b/lib/gds_api/test_helpers/rummager.rb
@@ -38,6 +38,18 @@ module GdsApi
         end
       end
 
+      def stub_any_rummager_search
+        stub_request(:get, %r{#{RUMMAGER_ENDPOINT}/search.json})
+      end
+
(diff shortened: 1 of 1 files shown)


tijmenb:
Would you mind rebasing this? This repo has govuk-lint now so should pick up some styling issues.

Out of interest, where are you using these? I'm surprised we didn't have them already.

36degrees:
Done. Using these in the service manual, to fetch a list of the most recently updated guides. I was also surprised they didn't already exist. There are some _very_ specific helpers around stubbing searches, but none of them did what I needed.

PR #602, opened 2016-10-04
Update deprecation warnings for configuration options

carvil:
This is a follow up change to commit 4ad659ff0becbc059e19bac07d9bd091dac7f788 and commit 4b1202b64d07ee619f907f7ca36bbc4e5780a7f3

We will be removing `always_raise_for_not_found` and `hash_response_for_requests` by December 2016. This change makes sure both options default to true when not configured and adds deprecation warnings to the setter methods to make sure client applications see it and are able to migrate before we remove them.

Please see commit messages for details on the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ GdsApi.configure do |config|
 end
 ```
 
+By default, `always_raise_for_not_found` will be true, so there is no need to
+configure it. By December 1st, 2016 this configuration option will no longer
+exist, defaulting to always raising exceptions for 404s and 410s.
+
 ## Logging
 
 Each HTTP request can be logged as JSON. Example:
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -32,6 +32,15 @@ GdsApi.configure do |config|
 end
 ```
 
+By default, both `always_raise_for_not_found` and `hash_response_for_requests`
```
tijmenb: The sentence about opt-in above doesn't make sense anymore. Perhaps we could rewrite this from the perspective of someone who upgrades api-adapters and sees a lot of errors in the tests?

carvil: I might even remove this message from the README and only leave the deprecation warnings. WDYT?

tijmenb:
Be sure to add something to the changelog & bump as major version.

PR #618, opened 2016-10-28
Fix publishing api expectations for schema validation

cbaines:
As a side effect of improving how the Publishing API handles govuk-content-schemas, the validation now takes place when running the pact tests. This is good, but requires changes to the expectations.


cbaines:
This is required to merge https://github.com/alphagov/publishing-api/pull/587

cbaines:
To break the deadlock with the above pull request, I'm going to merge this, even though one of the checks fails (as breaking the deadlock is what I expect to fix the test).

cbaines:
Well, that hasn't helped.

PR #622, opened 2016-12-01
Remove deprecated ostruct & 404-behaviour

tijmenb:
This gem used to return responses wrapped in OpenStructs. This often introduces issues because it's often unclear that you're talking to an object that returns `nil` for everything it doesn't know.

Using hash-accessors like `response.fetch("something")` encourages thinking about error cases and causes errors that are much easier to debug.

We've made this behaviour opt-in in https://github.com/alphagov/gds-api-adapters/pull/545 and made it default in https://github.com/alphagov/gds-api-adapters/pull/602.

Also, this gem used to return nils when the server returns 404. This makes it really easy to pass around nils that will cause errors far away from the originating code.

The option to a new always-raising behaviour was added in #544 and made default in #602.

As promised, it's December 1st and we're removing the old behaviour entirely.

cc @carvil


carvil:
Please release a new major version @tijmenb

PR #625, opened 2016-12-08
Remove deprecated v1 endpoints from PublishingApi

thomasleese:
These are being removed from the API itself so should no longer appear in here either.


tijmenb:
We should probably move the other methods in this class to `PublishingApiV2` so that Whitehall can use just one client (in a separate PR).

PR #627, opened 2016-12-08
Deprecate the bang methods

tijmenb:
We used to have bang! and non-bang versions of most methods, to indicate whether or not they would raise an error. This was inconsistently implemented, causing some confusion.

Since https://github.com/alphagov/gds-api-adapters/pull/622 this gem always raises for 404s.

This PR removes the bang-versions of the methods in content store, publishing-api and rummager.


MatMoore:
🎉

PR #629, opened 2016-12-08
Fix linting errors

tijmenb:
In https://github.com/alphagov/gds-api-adapters/pull/628 I'm attempting a large scale refactor. The linting won't pass for that PR because there are styleguide violations in all the files it's touching (we normally only run rubocop against changed files).

This fixes all styleguide violations so that that PR can be merged without noise.

### DIFF
--- a/lib/gds_api/test_helpers/business_support_helper.rb
+++ b/lib/gds_api/test_helpers/business_support_helper.rb
@@ -15,28 +15,29 @@ module GdsApi
           else
             results = @stubbed_business_supports
           end
-          {:body => plural_response_base.merge("results" => results, "total" => results.size).to_json}
+          { body: plural_response_base.merge("results" => results, "total" => results.size).to_json }
         end
-
       end
 
-      def api_has_business_support(business_support, facets={})
+      def api_has_business_support(business_support, facets = {})
         facets = sanitise_facets(facets)
+
         if business_support.is_a?(Symbol)
-          bs_with_facets = facets.merge(:title => business_support)
+          bs_with_facets = facets.merge(title: business_support)
         else
           bs_with_facets = facets.merge(business_support)
         end
+
         @stubbed_business_supports << bs_with_facets unless @stubbed_business_supports.include?(bs_with_facets)
       end
 
-      private
+    private
 
       def stubs_for_facets(facets)
         @stubbed_business_supports.select do |bs|
           facet_matches = 0
-          facets.each do |k,v|
-            if bs[k] and (v & bs[k]).any?
+          facets.each do |k, v|
+            if bs[k] && (v & bs[k]).any?
               facet_matches += 1
             end
           end
@@ -45,7 +46,7 @@ module GdsApi
       end
 
       def sanitise_facets(facets)
-        Hash[facets.map{ |k, v|
+        Hash[facets.map { |k, v|
           v = v.split(',') if v.is_a?(String)
           [k.to_sym, v]
         }]
--- a/lib/gds_api/test_helpers/common_responses.rb
+++ b/lib/gds_api/test_helpers/common_responses.rb
@@ -3,16 +3,16 @@ module GdsApi
     module CommonResponses
       def titleize_slug(slug, options = {})
         if options[:title_case]
-          slug.gsub("-", " ").gsub(/\b./) {|m| m.upcase }
+          slug.tr("-", " ").gsub(/\b./, &:upcase)
         else
-          slug.gsub("-", " ").capitalize
+          slug.tr("-", " ").capitalize
         end
       end
 
       # expects a slug like "ministry-of-funk"
       # returns an acronym like "MOF"
       def acronymize_slug(slug)
-        initials = slug.gsub(/\b\w+/) {|m| m[0] }.gsub("-", "")
+        initials = slug.gsub(/\b\w+/) { |m| m[0] }.delete("-")
         initials.upcase
       end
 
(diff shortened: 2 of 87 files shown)


jenkins.sh
```diff
@@ -47,8 +47,6 @@ for version in 2.3 2.2 2.1; do
   if [[ ${GIT_BRANCH} != "origin/master" ]]; then
     echo "Running ruby linter for $version"
     bundle exec govuk-lint-ruby \
-      --diff \
-      --cached \
```
danielroseman: Did you mean to delete these?

lib/gds_api/test_helpers/business_support_helper.rb
```diff
@@ -15,28 +15,27 @@ def setup_business_support_stubs(endpoint, path)
...
 
-      def api_has_business_support(business_support, facets={})
+      def api_has_business_support(business_support, facets = {})
         facets = sanitise_facets(facets)
-        if business_support.is_a?(Symbol)
-          bs_with_facets = facets.merge(:title => business_support)
-        else
-          bs_with_facets = facets.merge(business_support)
-        end
+        bs_with_facets = if business_support.is_a?(Symbol)
```
danielroseman: We've previously discussed disabling the Style/ConditionalAssignment cop that enforces this; it leads to very ugly results here and in a few cases below.

lib/gds_api/test_helpers/common_responses.rb
```diff
@@ -27,15 +27,13 @@ def response_base
...
-            "description" => "Tags!",
-            "total" => 100,
-            "start_index" => 1,
-            "page_size" => 100,
-            "current_page" => 1,
-            "pages" => 1,
-            "results" => []
-          }
+          "description" => "Tags!",
+"total" => 100,
```
danielroseman: indentation

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -62,19 +62,17 @@ def stub_any_email_alert_api_call
...
-        if attributes
-          matcher = ->(request) do
-            payload = JSON.parse(request.body)
-            payload.select { |k, _| attributes.key?(k) } == attributes
-          end
-        else
-          matcher = nil
-        end
+        matcher = if attributes
+                    ->(request) do
```
danielroseman: Another case where Style/ConditionalAssignment makes this unreadable.

test/list_response_test.rb
```diff
@@ -2,24 +2,22 @@
...
 
     it "should allow Enumerable access to the results array" do
       data = {
-        "results" => [
-          "foo",
-          "bar",
-          "baz",
-        ],
+        "results" => %w(
+foo
```
danielroseman: Should probably be all on one line

lib/gds_api/test_helpers/common_responses.rb
tijmenb: That's a really weird thing for rubocop to get wrong :|

lib/gds_api/test_helpers/content_api/artefact_stub.rb
```diff
@@ -38,19 +38,20 @@ def stub
...
-          end
 
-          # Ensure that all keys and values are strings 
+      private
+
+        def url_without_query
+          "#{CONTENT_API_ENDPOINT}/#{CGI.escape(slug)}.json"
+        end
+
+          # Ensure that all keys and values are strings
```
MatMoore: It might be the way github is showing the diff, but it looks like this comment is overindented now.

lib/gds_api/test_helpers/business_support_helper.rb
tijmenb: This isn't that cop actually. I get the following report:

```
lib/gds_api/test_helpers/business_support_helper.rb:25:11: C: Style/IndentationWidth: Use 2 (not -15) spaces for indentation.  (https://github.com/bbatsov/ruby-style-guide#spaces-indentation)
          facets.merge(title: business_support)
          ^^^^^^^^^^^^^^^
```

I think the `if`, `else` and `end` should be aligned. I'll check if I can make the linter pass differently.

jenkins.sh
tijmenb: Yeah, but I should've done that in a separate commit. Removed for now. Thanks!

lib/gds_api/test_helpers/business_support_helper.rb
tijmenb: Ok, I think I got it. The style used here doesn't pass the indentation rules, but `Style/ConditionalAssignment` doesn't allow us to use 

```
if bla
 foo = "a"
else
 foo = "b"
end
```

Seems like cops fighting for jurisdiction 👮. I've disabled `Style/ConditionalAssignment` and changed the examples to the "double assignment" to resolve this.

lib/gds_api/test_helpers/common_responses.rb
tijmenb: Fixed.

lib/gds_api/test_helpers/email_alert_api.rb
tijmenb: Fixed.

lib/gds_api/test_helpers/content_api/artefact_stub.rb
tijmenb: Yep, correct. Fixed!

tijmenb:
Thanks for the review @danielroseman @MatMoore - fixed the issues!

PR #631, opened 2016-12-14
Add explicit handling for Invalid URI exceptions

whoojemaflip:
Before sending a request, the ApiAdapters pass a string to the `URI` module for parsing. If this fails, we have been getting `URI::InvalidURIError` exceptions, which [we are seeing in Errbit](https://errbit.publishing.service.gov.uk/apps/53611a4b0da1151271001055/problems/584ebea66578630e38370100).  

In order to better handle errors thrown in the api-adapter, here we have created an exception class that derives from `GdsApiAdapters::BaseError` and this is thrown in the event of a malformed URL.

### DIFF
--- a/lib/gds_api/exceptions.rb
+++ b/lib/gds_api/exceptions.rb
@@ -46,6 +46,8 @@ module GdsApi
   class HTTPUnprocessableEntity < HTTPClientError
   end
 
+  class InvalidUrl < BaseError; end
+
   class NoBearerToken < BaseError; end
 
   module ExceptionHandling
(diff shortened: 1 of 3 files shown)


lib/gds_api/exceptions.rb
```diff
@@ -46,6 +46,8 @@ class HTTPConflict < HTTPClientError
   class HTTPUnprocessableEntity < HTTPClientError
   end
 
+  class InvalidUrlException < BaseError; end
```
whoojemaflip: I've got no strong feelings either way, so your vote for `InvalidUrl` is enough to swing the name-o-meter. Updated.

PR #632, opened 2016-12-14
Add rummager test helper with missing keys

rubenarakelyan:
This commit adds a rummager test helper which returns search results with missing keys to simulate unexpanded links due to unsynced data between rummager and publishing-api.

### DIFF
--- a/lib/gds_api/test_helpers/rummager.rb
+++ b/lib/gds_api/test_helpers/rummager.rb
@@ -95,6 +95,11 @@ module GdsApi
         run_example_query
       end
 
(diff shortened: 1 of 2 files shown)


tijmenb:
Perhaps the use case for this is limited to the `collections` app. Would it be easier to have the helper & fixture in that app? You can always extend the class there.

rubenarakelyan:
For now, yes this is limited to collections so I'll close this and extend the class there to handle this use case. I'll add a note there in case someone in the future needs something similar to move it here.

PR #634, opened 2016-12-15
Remove deprecated 'format' field

thomasleese:
This will shortly be removed from the publishing-api, therefore it is necessary to remove tests for it.

### DIFF
--- a/test/publishing_api_v2_test.rb
+++ b/test/publishing_api_v2_test.rb
@@ -11,7 +11,8 @@ describe GdsApi::PublishingApiV2 do
       "content_id" => content_id,
       "title" => "Instructions for crawler robots",
       "description" => "robots.txt provides rules for which parts of GOV.UK are permitted to be crawled by different bots.",
-      "format" => "special_route",
+      "schema_name" => "special_route",
+      "document_type" => "different_to_special_route",
       "public_updated_at" => "2015-07-30T13:58:11.000Z",
       "publishing_app" => "static",
       "rendering_app" => "static",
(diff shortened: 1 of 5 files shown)


test/publishing_api_v2_test.rb
```diff
@@ -260,7 +259,7 @@ def content_item_for_content_id(content_id, attrs = {})
       it "responds with 200 and the content item" do
         response = @api_client.get_content(@content_id)
         assert_equal 200, response.code
-        assert_equal @content_item["format"], response["document_type"]
+        assert_equal @content_item["schema_name"], response["document_type"]
```
tijmenb: This should probably assert that `assert_equal @content_item["document_type"], response["document_type"]`

test/publishing_api_v2_test.rb
```diff
@@ -11,7 +11,8 @@ def content_item_for_content_id(content_id, attrs = {})
       "content_id" => content_id,
       "title" => "Instructions for crawler robots",
       "description" => "robots.txt provides rules for which parts of GOV.UK are permitted to be crawled by different bots.",
-      "format" => "special_route",
+      "schema_name" => "special_route",
+      "document_type" => "special_route",
```
tijmenb: Perhaps it would be good to use a different thing for `document_type` & `schema_name` to indicate that there's a difference (you can make something up for `document_type`).

thomasleese: Thanks, I've changed that now.

PR #636, opened 2016-12-15
Bump version to 38.1.0

whoojemaflip:
There was a change in the handling of Invalid URI exceptions which, while
not part of the public API, may have been handled in downstream applications. Hence the minor-version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-# Unreleased
+# 38.1.0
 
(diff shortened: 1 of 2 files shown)


tijmenb:
The changelog needs a bump as well.

dougdroper:
just a quick note to say that even though this is reporting as green in github, if you look on ci it is red. We are currently looking into why.

PR #639, opened 2016-12-16
Remove test files from the built gem

Davidslv:
When the test/spec directory is mentioned in the gemspec it actually
builds the gem with those files.

There's also no reason to send the test files to production.

### Before

screen shot 2016-12-16 at 11 36 57

### After
screen shot 2016-12-16 at 11 40 48

### DIFF
--- a/gds-api-adapters.gemspec
+++ b/gds-api-adapters.gemspec
@@ -15,7 +15,6 @@ Gem::Specification.new do |s|
   s.description  = "A set of adapters providing easy access to the GDS GOV.UK APIs"
 
   s.files        = Dir.glob("lib/**/*") + %w(README.md Rakefile)
-  s.test_files   = Dir['test/**/*']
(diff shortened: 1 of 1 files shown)


cbaines:
@Davidslv I changed gds-api-adapters to use the new CI system on Friday, but this has only been checked on the old CI system. Could you rebase on master, as this will then include the Jenkinsfile and be tested on the new CI system?

tijmenb:
David will be back from hols next week, closing in the meantime.

Davidslv:
Opened and rebased. Hope it's all ok now 👍

cbaines:
I've just been looking at updating Whitehall's use of gds-api-adapters, and noticed that removing some of these files broke some test helpers. I've got a patch to add the fixtures back here https://github.com/alphagov/gds-api-adapters/pull/665 .

PR #641, opened 2016-12-16
Update gem author to be Government Digital Service

Davidslv:
I believe that gems under the [govuk rubygems account](https://rubygems.org/profiles/govuk) should have no individual authors but mention `Government Digital Service` as a whole.

We are a team and most of us contribute to several gems but never got to add themselves as an author. By setting this as `Government Digital Service` it's self implied that anyone that contributed is contributing for the `Government Digital Service` as a team.

My other proposal would be to follow [Slimmer](https://github.com/alphagov/slimmer/blob/master/slimmer.gemspec#L11-L12) instead


tijmenb:
The slimmer one is correct one, please use that.

PR #647, opened 2016-12-22
Bump version to 39.0.0

issyl0:
- This version will break Maslow's master, hence the major version bump.
  Maslow is the only thing that uses `need_api_has_organisations` (but
  there's a PR on Maslow to stop it using this).


issyl0:
Pact tests are failing. 😞

PR #648, opened 2016-12-22
When verifying the pact, use the local copy

cbaines:
For some reason, the pact recieved from the pact broker sometimes does
not match the file on disk, which should be published to the pact
broker.

To work around this, and make the whole situation simpler, just use the
local file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,7 +102,7 @@ node {
 
     stage("Run publishing-api pact") {
       dir("publishing-api") {
-        withEnv(["JOB_NAME=publishing-api"]) { // TODO: This environment is a hack
+        withEnv(["JOB_NAME=publishing-api", "USE_LOCAL_PACT=true"]) { // TODO: This environment is a hack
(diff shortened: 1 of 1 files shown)


cbaines:
https://github.com/alphagov/gds-api-adapters/pull/647 fails in CI, but I don't think it should. This change seems to fix the issue.

danielroseman:
This doesn't appear to be the right thing to do. We have the pact broker for a reason; if it's not providing the right version, we should fix that. We will shortly be starting a story to investigate and document exactly what the broker is doing and fix any inconsistencies.

cbaines:
> This doesn't appear to be the right thing to do. We have the pact broker for a reason; if it's not providing the right version, we should fix that. We will shortly be starting a story to investigate and document exactly what the broker is doing and fix any inconsistencies.

Ok, in the mean time, I think pushing up the same commits with a different branch name managed to work around whatever issue that was being encountered, so I'll close this for now.

PR #650, opened 2017-01-06
Make git credentials available when publishing gem

cbaines:
The publish_gem task pushes a Git tag, and I'm hoping that this will
make it work, instead of failing with an error that looks like its
missing credentials.


cbaines:
This is motivated by the failure here https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/master/18/console

PR #654, opened 2017-01-19
Fix gem publishing from CI.

danielroseman:
Jenkins is complaining "must specify $class with an implementation ofs
class", so add the missing value.

Not sure whether this is the right value, or how to test it, since this is only triggered on master, so this is a speculative fix.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -135,6 +135,7 @@ node {
           echo 'Publishing gem'
           withCredentials([
             [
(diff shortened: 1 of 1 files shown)


cbaines:
I never managed to get the Git credentials stuff to work, so I'd recommend reverting e82437325bc17fd005da92879706810e1b7f42b2 instead.

I hope that this will at least allow the Gem to be published.

PR #656, opened 2017-01-19
Add params to test helpers

binaryberry:
Allow to pass params for #publishing_api_has_item means Maslow will be able to get successive versions of a piece of content, passing in the version as a param.

This also removes the with_indifferent_access method, which is a Rails method. Although gds-api-adapters is being used in Rails apps at the moment, Rails is not listed as a dependency. This replaces the use of with_indifferent_access with deep_transform_keys to convert keys to symbols, or deep_stringify_keys to convert them to strings.

[Trello](https://trello.com/c/8YShGG7D/24-get-maslow-to-render-content-from-the-publishing-api)

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -296,7 +296,7 @@ module GdsApi
       # publishing_api_has_content allows for flexible passing in of arguments, please use instead
       def publishing_api_has_fields_for_document(document_type, items, fields)
         body = Array(items).map { |item|
-          item.with_indifferent_access.slice(*fields)
+          deep_stringify_keys(item).slice(*fields)
         }
 
         query_params = fields.map { |f|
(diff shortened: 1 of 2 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -296,7 +296,7 @@ def publishing_api_has_content(items, params = {})
       # publishing_api_has_content allows for flexible passing in of arguments, please use instead
       def publishing_api_has_fields_for_document(document_type, items, fields)
         body = Array(items).map { |item|
-          item.with_indifferent_access.slice(*fields)
+          deep_stringify_keys(item, &:to_sym).slice(*fields)
```
gpeng: Should this be `deep_transform_keys`?

binaryberry: Good spot - I think it could be either `deep_transform_keys` or `deep_stringify_keys`, as the body will be converted to json on line 308, but if I'm going to use `deep_stringify_keys` I shouldn't pass in a block. Amended in 2157c77.

PR #668, opened 2017-02-23
Add a redirects option to the Unpublish adapter in Publishing API

emmabeynon:
The Publishing API has recently added a `redirects` option to the
Unpublish endpoint.


cbaines:
I always forget, but I think its good to update the changelog when making changes, as that makes it easier when it comes around to doing the releases.

emmabeynon:
@cbaines Thanks for the reminder - now added to the Changelog!

PR #670, opened 2017-02-24
Update publishing of gem

brenetic:
This changes the Jenkinsfile to use the `publishGem` method available in the groovy scripts in Puppet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -143,7 +143,7 @@ node {
 
         stage("Publish gem") {
           echo 'Publishing gem'
-          govuk.runRakeTask("publish_gem --trace")
+          govuk.publishGem(REPOSITORY, env.BRANCH_NAME)
         }
       }
     }


sihugh:
Do we want to retain the ability to publish from a rake task? No problems about updating the `Jenkinsfile`, just wonder whether we don't want to rely on Jenkins working for this.

brenetic:
@sihugh That's a good point! I have reverted those changes

tijmenb:
I think we're deprecating the gem. @alecgibson can you elaborate?

alecgibson:
See this [Pull Request](https://github.com/alphagov/govuk-puppet/pull/5519) for more context. Basically we couldn't get the publishing gem to work with Jenkins 2.

Is there any particular reason you wouldn't rely on Jenkins working to publish a Gem? We rely on it working for all other deployment.

PR #673, opened 2017-03-09
Pass optional header params to email_alert_api.send_alert

dwhenry:
This is required when called from email_alert_service
as this has picked a message up from rabbit MQ and
needs to manually pass the govuk request id in the
header.


dwhenry:
@jennyd have fixed the rails 2.1 syntax issues

PR #678, opened 2017-03-15
Add api adapter for /v2/lookup-by-base-path

MatMoore:
Adapters for https://github.com/alphagov/publishing-api/pull/822

Trello: https://trello.com/c/MHBTiA65/544-publishing-api-support-for-lookups-of-draft-content-split-from-allow-users-to-tag-draft-content

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -53,6 +53,42 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     response.to_hash
   end
 
+  # Get the current draft and/or the current live edition matching a set of base paths.
+  # "draft" and "live" refer to the content store the content is in.
+  # The live edition contains an `unpublishing` hash if the edition is unpublished.
+  # An unpublishing hash has `type` and `redirect` keys matching the unpublish request.
+  #
+  # The request accepts both GET/POST. We use POST here to avoid long URIs.
+  # @param base_paths [Array]
+  # @return [Hash] a hash keyed by `base_path`. The value is nil if the base path is not used or belongs to an access limited draft.
+  # @example
+  #
+  #   publishing_api.lookup_by_base_paths(['/pay-vat', '/nothing'])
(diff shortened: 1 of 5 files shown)


test/test_helper.rb
```diff
@@ -14,6 +14,7 @@
 require 'mocha/mini_test'
 require 'timecop'
 require 'gds-api-adapters'
+require 'pry-byebug'
```
kevindew: is this meant to be here?

lib/gds_api/publishing_api_v2.rb
```diff
@@ -53,6 +53,42 @@ def lookup_content_ids(base_paths:)
...
+  # @param base_paths [Array]
+  # @return [Hash] a hash keyed by `base_path`. The value is nil if the base path is not used or belongs to an access limited draft.
+  # @example
+  #
+  #   publishing_api.lookup_by_base_paths(['/pay-vat', '/nothing'])
+  #   # => {"/pay-vat"=>{"live"=>{"content_id"=>"a484eaea-eeb6-48fa-92a7-b67c6cd414f6", "locale"=>"en", "document_type"=>"guide"}}, "/nothing"=>nil}
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2lookup-by-base-path
+  def lookup_by_base_paths(base_paths)
+    response = post_json("#{endpoint}/v2/lookup-by-base-path", base_paths: base_paths)
```
kevindew: any chance we could check the length of the string to determine whether it is get or post? It'd be preferable if it was mostly a get rather than post

test/test_helper.rb
MatMoore: Does it hurt?

lib/gds_api/publishing_api_v2.rb
MatMoore: I'll look into this, I don't see why not but I'd need to get the limit right.

test/test_helper.rb
kevindew: sure, just an unused dependency

lib/gds_api/publishing_api_v2.rb
kevindew: the default seems to be 8k so we could experiment with that as a benchmark

tijmenb:
We're picking up https://trello.com/c/6VPQcc7C next week, closing in the meantime.

PR #685, opened 2017-03-23
Test published special routes against the schemas

suzannehamilton:
Add govuk-content-schema-test-helpers dependency and use it to check that special route pages sent to the publishing API match the publishing schema.

It includes a commit to update the webmock dependency and fix the broken tests because I thought that was necessary to add govuk-content-schema-test-helpers. That turned out to be wrong (I misread the bundler error message), but I think it's still a useful fix.

This whole change should help us update the special-routes schemas by providing extra regression tests. I'll add this gem to the downstream schema test list so that gets run against each version of the schemas.

Trello: https://trello.com/c/u1H1rgDA/525-some-finding-pages-are-still-being-tracked-as-thing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,6 +44,11 @@ node {
       }
     }
 
+    stage("Set up content schema dependency") {
+      govuk.contentSchemaDependency('deployed-to-production')
+      govuk.setEnvar("GOVUK_CONTENT_SCHEMAS_PATH", "${pwd()}/tmp/govuk-content-schemas")
+    }
+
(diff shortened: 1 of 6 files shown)


suzannehamilton:
I've added an [extra commit](https://github.com/alphagov/gds-api-adapters/pull/685/commits/d98656012c6a274f2c13b4f727b5f471d8b2bf64) to fix the Jenkins build: the content schemas are now checked out near the start of the CI build so that they are available to this project's unit tests, not just the publishing-api Pact tests.

cc @cbaines, since you reviewed an earlier version of this PR.

PR #689, opened 2017-03-27
Release version 41.2.0

suzannehamilton:
Trello: https://trello.com/c/u1H1rgDA/525-some-finding-pages-are-still-being-tracked-as-thing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+# 41.2.0
+
+* Update webmock gem dependency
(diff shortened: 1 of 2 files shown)


suzannehamilton:
@cbaines Oh, interesting. The failed build was running at the same time a publishing-api build, and there's a line in the console output:

```
ERROR:  database "publishing_api_test" is being accessed by other users
```

I've rebuilt it and it passed, so I think the problem was that two builds were accessing the same DB simultaneously. I've also raised a separate PR (#690) which should hopefully prevent it from happening again.

PR #695, opened 2017-04-24
Make LGIL code mandatory for local links requests

sihugh:
So that we can make administering links easier, we want to be able to only
support a defined set of LGSL / LGIL combinations.  We currently have some
logic to find the "best" result if an LGIL code isn't supplied with a request
for a link, but this results in some tricky code, more links to support and
uncertainty about whether the links that we provide are as good as they might
be.

To do this we need to:

- Assign LGIL codes to all `local_transaction` formats in
Publisher (done)
- Make this change to GDS API adapters (and bump the major version)
- Bump the Gem in Frontend and make sure that everything works
- Make the LGIL code mandatory on the LLM api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Make `lgil` mandatory when requesting links from Local Links Manager
+
 # 41.5.0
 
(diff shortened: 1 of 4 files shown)


issyl0:
Making the data model simpler! 🎉

PR #697, opened 2017-04-27
Remove business support schemes

alecgibson:
We’re retiring the bespoke `business-support-api` application, because
it’s being replaced by a specialist publisher-style finder.

This application used to rely on fetching `business_support_schemes.json`
from `govuk_content_api`, but since the app is being retired we no
longer need to serve this JSON file from `govuk_content_api`, and hence
don’t need the corresponding API adapters.

### Trello

https://trello.com/c/VUl4KZzL/75-remove-business-support-schemes-json-from-content-api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Set the `update_type` to major in all of the Publishing API pact tests.
+* Remove support for `business_support_schemes`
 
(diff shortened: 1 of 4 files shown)


alecgibson:
@cbaines done

cbaines:
> @cbaines done

Great, I'm still happy for this to be merged :)

PR #698, opened 2017-04-27
Set `update_type` to major in Publishing API tests

leenagupte:
Trello card: https://trello.com/c/3S4m5wAJ

## Motivation

The publishing-api is being changed to only allow the `update_type` to be set to major on first publication. The pact tests have been updated to reflect this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Set the `update_type` to major in all of the Publishing API pact tests.
+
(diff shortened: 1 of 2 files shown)


cbaines:
I'd recommend not rolling this in with a release, as if other changes are merged in the mean time, the numbers and changelog might get a bit muddled. I'd keep the note in the changelog though.

leenagupte:
@cbaines Ok, I can split the release bump into a separate PR.

PR #699, opened 2017-04-28
Support custom schemas for Special Route Publishing

whoojemaflip:
Allows a custom schema to be passed in when publishing a special route.


cbaines:
When does a special route not match the special route schema? (I'm assuming that is why the schema name needs to be different in this use case?)

whoojemaflip:
This change is specifically to support the `homepage` https://github.com/alphagov/govuk-content-schemas/pull/610

Seems to me that the `SpecialRoutePublisher` is fundamentally more about maintaining these special pages than it is about policing the use of the `special_route` schema.

PR #700, opened 2017-04-28
Bump version to 43.0.0

whoojemaflip:


cbaines:
The removal of business support things probably makes this a major version bump? I don't particularly mind though.

whoojemaflip:
@cbaines I naively assumed that no-one would leave a major change undeployed 😢

PR #703, opened 2017-05-08
Remove support for business-support-api

rubenarakelyan:
This commit removes support for the `business-support-api`, which has now been retired.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
-* Revert changes made to `update_type` in the Publishing API pact tests in release 43.0.0.
+* Revert changes made to `update_type` in the Publishing API pact tests in release 43.0.0
```
cbaines: The diff is odd here, I'm unsure why this line is showing as deleted/added. The only difference seems to be the "." at the end.

PR #706, opened 2017-05-17
Tidy up support and support-api adapters and test helpers

h-lame:
1. Remove dead end points from support that were moved into support-api in Oct 2014
2. Rename test helpers for support-api adapter to make it clearer they work on the support-api

More details in the individual commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Drop dead endpoints from support adapter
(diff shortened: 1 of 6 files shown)


h-lame:
The renames only affect the support-api, support, and feedback tests; these can be updated when these apps are upgraded to the latest gds-api-adapters.  Whitehall uses a `stub_organisation` method in tests, but it is unrelated to the `stub_organisation` method we rename here.

PR #710, opened 2017-06-21
Add API endpoints used by info-frontend

elenatanasoiu:
For: https://trello.com/c/9JNitfXS/173-spike-info-pages

This will add an api endpoint for info-frontend to make it possible for it
to talk to the Backdrop Read API.

This will enable us to bypass metadata-api for making these api calls, and
will ultimately allow us to retire metadata-api, since it it currently only
used for these api calls.

This commit will be followed with a commit in info-frontend to use these
endpoints.

### DIFF
--- a/lib/gds_api/performance_platform/data_out.rb
+++ b/lib/gds_api/performance_platform/data_out.rb
@@ -43,6 +43,89 @@ module GdsApi
       def service_feedback(transaction_page_slug)
         get_json("#{endpoint}/data/#{transaction_page_slug}/customer-satisfaction")
       end
+
+      # Fetching statistics data from the performance platform for a given page slug
+      #
+      # Makes a +GET+ request.
+      #
+      # @param slug [String] Points to the page for which we are requesting
+      # statistics.
+      # @param is_multipart [Boolean] Flag that marks whether the slug is multipart
+      # or not:
+      #
+      # - simple: `/european-health-insurance-card`
+      # - multipart: `/european-health-insurance-card/123`
+      #
+      # # @examples
+      #
+      #  1. Without multipart filtering:
+      #
+      #  performance_platform_data_out.search_terms('/european-health-insurance-card')
+      #
+      #  2. With multipart filtering:
+      #
+      #  performance_platform_data_out.searches('/european-health-insurance-card', true)
+      #  performance_platform_data_out.page_views('/european-health-insurance-card', true)
+      #  performance_platform_data_out.problem_reports('/european-health-insurance-card', true)
+
+      def search_terms(slug)
+        options = {
+            slug: slug,
+            transaction: 'searchTerms',
+            group_by: 'searchKeyword',
+            collect: 'searchUniques:sum'
+        }
+        statistics(options)
+      end
+
+      def searches(slug, is_multipart)
+        options = {
+            slug: slug,
+            transaction: 'searchTerms',
+            group_by: 'pagePath',
+            collect: 'searchUniques:sum'
+        }
+        statistics(options, is_multipart)
+      end
+
+      def page_views(slug, is_multipart)
+        options = {
+            slug: slug,
+            transaction: 'page-statistics',
+            group_by: 'pagePath',
+            collect: 'uniquePageviews:sum'
+        }
+        statistics(options, is_multipart)
+      end
+
+      def problem_reports(slug, is_multipart)
+        options = {
+            slug: slug,
+            transaction: 'page-contacts',
+            group_by: 'pagePath',
+            collect: 'total:sum'
+        }
+        statistics(options, is_multipart)
+      end
+
+      # This can be used as a free form call to the performance platform.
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/performance_platform/data_out.rb
```diff
@@ -17,6 +17,70 @@ def stub_data_set_not_available(slug)
...
+              filter_by: "pagePath:" + slug,
+              group_by: group_by,
+              collect: collect,
+              duration: 42,
+              period: "day",
+              end_at: Date.new.to_time.getutc
+          }
+
+          data_out = GdsApi::PerformancePlatform::DataOut.new(PP_DATA_OUT_ENDPOINT)
+          query_string = data_out.send(:query_string, params)
```
h-lame: Using the implementation to get the query string version of the params feels a bit weird (on first reading I thought we were actually making the API call!).  I hoped there might be a simpler way and it turns out that [Webmock can take a `query` argument](https://github.com/bblimke/webmock#matching-query-params-using-hash) that takes a hash to avoid this.

```ruby
stub_http_request(:get, "#{PP_DATA_OUT_ENDPOINT}/data/govuk_info/page-statistics")
  .with(query: params)
  .to_return(status: 200, body: response_body.to_json)
```

lib/gds_api/performance_platform/data_out.rb
```diff
@@ -43,6 +43,119 @@ class DataOut < GdsApi::Base
...
+      # @param collect [String] The operation that should be applied
+      # on the results after they've been grouped
+      #
+      # # @example
+      #
+      #  performance_platform_data_out.search_terms('/european-health-insurance-card', 'searchKeyword', 'searchUniques:sum')
+
+      def search_terms(slug, group_by, collect)
+        params = {
+            filter_by: "pagePath:" + slug,
```
h-lame: It seems odd to me that this doesn't take the is_multipart argument to switch `filter_by` into `filter_by_prefix`.  It is [exactly what metadata API's go implementation does](https://github.com/alphagov/metadata-api/blob/87cc4db4f6801ac8936419a0a697e09e9012f405/performance_platform/statistics.go#L116-L123) though so we should leave it for now.  That said I think we should take a note to check if we should be toggling `filter_by` / `filter_by_prefix` for this API call and we can make that change later.  For now we want things to be a drop in replacement, we can improve it later.

lib/gds_api/performance_platform/data_out.rb
```diff
@@ -43,6 +43,119 @@ class DataOut < GdsApi::Base
...
+      #
+      #  performance_platform_data_out.search_terms('/european-health-insurance-card', 'pagePath', 'uniquePageviews:sum')
+
+      def page_statistics(slug, group_by, collect, is_multipart)
+        params = {
+            group_by: group_by,
+            collect: collect,
+            duration: 42,
+            period: "day",
+            end_at: Date.new.to_time.getutc
```
h-lame: This should be `Date.today` rather than `Date.new` as Date.new returns something distant like `"-4712-01-01"` (may be system-dependant).  We may have to worry about timezones, but I don't think we can rely on activesupport so we can't do `Time.zone.today` instead which is a shame.

We'll also want to add `.iso8601` to the end to turn the Time object into a good representation for sending across the wire.  `query_string` simply calls `to_s` on each param value and `to_s` on a Time instance gives us something like: `"2017-06-20 23:00:00 UTC"`, if we call `iso8601` first we get `"2017-06-20T23:00:00Z"`.  It's not clear to me if backdrop will properly parse the former and give us the correct results, but the example urls I saw in kibana were all using the latter so it's better to be safe.

This comment applies to all the `Date.new` bits.

lib/gds_api/performance_platform/data_out.rb
```diff
@@ -43,6 +43,119 @@ class DataOut < GdsApi::Base
       def service_feedback(transaction_page_slug)
         get_json("#{endpoint}/data/#{transaction_page_slug}/customer-satisfaction")
       end
+
+      # Fetches data from the performance platform for a given page slug
```
h-lame: It's great to have doc comments for these methods, but they've all got the same description, can we tweak them from `fetches data` to `fetches <whatever> data` so they're more useful? We're also missing the `@param is_multipart` for those methods that take that param.

lib/gds_api/test_helpers/performance_platform/data_out.rb
elenatanasoiu: Nice!

lib/gds_api/performance_platform/data_out.rb
h-lame: Ah, you've changed the implementation to one method since I reviewed.  Might be worth listing some examples of the transactions, group_by, and collect params to help folk understand what kinds of calls they can make.  (Maybe we can link to an API doc for backdrop read if such a thing exists?).

PR #709, opened 2017-05-22
Remove support for the Content API

alecgibson:
The [`content_api`](https://github.com/alphagov/govuk_content_api) has
been retired, and all content has been migrated to the
[`content_store`](https://github.com/alphagov/content-store).

Therefore, we no longer need to provide support for this API.

### Trello

https://trello.com/c/I4Xtuwe9/142-remove-content-api-from-gds-api-adapters

PR #712, opened 2017-06-30
Release major version 47.0.0

elenatanasoiu:
The reason for a major version bump is because we are removing support for content-api. We are also introducing 4 new endpoints that talk to backdrop read API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
-# NEXT RELEASE
+# 47.0.0
 
 * Remove support for `content-api`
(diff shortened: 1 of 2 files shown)


h-lame:
This'll have to be 47 because it includes the removal of content-api #709

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
-# NEXT RELEASE
+# 47.0.0
 
 * Remove support for `content-api`
+* Add `HTTPUnprocessableEntity` exceptions
```
h-lame: Nice - thanks for trawling through other PRs to add this!

PR #719, opened 2017-07-20
Make settting the update_type when publishing optional

thomasleese:
This parameter is being deprecated and instead we expect clients to send the update_type in the body of the content when doing a PUT request.

[Trello Card](https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk)

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -95,7 +95,7 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
   # @option options [String] locale The language, defaults to 'en' in publishing-api.
   #
   # @see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idpublish
-  def publish(content_id, update_type, options = {})
+  def publish(content_id, update_type = nil, options = {})
     params = {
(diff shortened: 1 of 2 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -95,7 +95,7 @@ def lookup_content_id(base_path:, exclude_document_types: nil, exclude_unpublish
   # @option options [String] locale The language, defaults to 'en' in publishing-api.
   #
   # @see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-v2contentcontent_idpublish
-  def publish(content_id, update_type, options = {})
+  def publish(content_id, update_type = nil, options = {})
     params = {
       update_type: update_type
     }
```
cbaines: If possible, it might be neater to make `update_type` an optional key? Could the argument just be removed, and `update_type` added to `optional_keys`?

thomasleese: I agree, that would be better, but the problem is that apps currently call this method like `publish(content_id, "major")` and that change would break them all.

PR #720, opened 2017-07-20
Release version 47.1.4

thomasleese:


cbaines:
I'd count this as a backwards compatible change, and not a bug fix, so I'd have gone with 47.2.3, assuming the intent is to use semantic versioning.

thomasleese:
Hmm, yeah, good point.

thomasleese:
Will close and re-open as the branch name is confusing.

PR #728, opened 2017-07-27
Add `get_paged_editions` method

alecgibson:
This change adds a method for paging through responses from the
`/v2/editions` endpoint.

The `editions` endpoint is paged through its returned `next` link,
which doesn’t work well with the `gds-api-adapters` pattern. This
change rolls the pagination into a method that returns a lazy
`Enumerator` that knows how to fetch the next page.

## Dependencies

- [x] https://github.com/alphagov/publishing-api/pull/982
- [x] https://github.com/alphagov/publishing-api/pull/983

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -324,6 +324,27 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     get_json(get_editions_url(params))
   end
 
+  # Returns an Enumerator of Response objects for each page of results of
+  # editions for the provided query string parameters.
+  #
+  # @param params [Hash]
+  #
+  # @return [Enumerator] an enumerator of editions responses
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2editions
+  def get_paged_editions(params = {})
+    Enumerator.new do |yielder|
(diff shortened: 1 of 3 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -324,6 +324,19 @@ def get_editions(params = {})
     get_json(get_editions_url(params))
   end
 
+  def get_paged_editions(params = {})
+    Enumerator.new do |yielder|
+      next_link = get_editions_url(params)
+      while next_link
+        yielder.yield begin
+          response = get_json(next_link)
```
thomasleese: I wonder if it's more intuitive here to iterate over the results and then yield them one by one? Or perhaps include another convenience method that does that? What do you think?

alecgibson: I'd originally thought about doing that, but it felt out-of-place in `gds-api-adapters`, which has all(?) of its methods return a `Response` object. It felt more natural to return pages of responses, rather than actual results, and then leave the clients to do things like `response.map { |response| response['results'] }.flatten` or similar.

For example, what if `publishing-api` errors? Is that really the responsibility of `gds-api-adapters`?

thomasleese: Yeah, that sounds very reasonable to me! 👍

PR #730, opened 2017-07-28
Release v47.4.0

alecgibson:


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Unreleased
+# 47.3.1
```
thomasleese: I think this should be version `47.4.0` actually since it's a new non-breaking change.

alecgibson: Yeah fine by me - was unsure if this counted as a whole new feature.

PR #752, opened 2017-09-21
Add GdsApi::AssetManager#create_whitehall_asset method

floehopper:
We want to expose the new Asset Manager API endpoint for creating Whitehall assets.

The main difference between the existing `GdsApi::AssetManager#create_asset` method and this new method is that the client must specify a `legacy_url_path` as well as a `file`. The former is used to *tell* Asset Manager the public URL path at which it should make the asset available.

This is intended to be an interim measure to allow us to migrate Whitehall assets to Asset Manager without changing their URLs. In the fullness of time the plan is for Asset Manager to determine the public URL paths for *all* assets, including Whitehall assets, and at that point Whitehall will need to take notice of the `file_url` returned in the response from the Asset Manager API.

Note that this new method should *only* be used by the Whitehall Admin app, not any of the other publishing applications.

See alphagov/asset-manager#203 for more info.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
+# 49.2.0
+
+* Add GdsApi::AssetManager#create_whitehall_asset method (#752)
+
 # 49.1.0
 
 * Remove trailing slash in call to get_link_changes
 
 # 49.0.0
 
-* Remove `GdsApi::GovUkDelivery` and helpers as `govuk-delivery` has been 
+* Remove `GdsApi::GovUkDelivery` and helpers as `govuk-delivery` has been
 retired in favour of `email-alert-api`
 * Add get_link_changes endpoint for publishing-api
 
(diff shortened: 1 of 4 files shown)


floehopper:
@chrislo: Thanks. I assume you're thinking that a deprecation warning would discourage people from using this method in other publishing apps and/or using the `legacy_url_path` parameter - is that right? I think having the word "whitehall" in the method name is sufficient to discourage people from doing this and, in any case, I don't think it's a big concern - I don't think anyone is actively making changes to the way publishing apps interact with Asset Manager. So I think it's a bit premature to add a deprecation warning and I'd rather add it when we're actually thinking about removing the method. I hope that makes sense.

chrislo:
@floehopper yes - it was the gentle nudge to avoid using it for other publishing apps. But I agree the naming covers this. I'll mark as approved!

Issue #757, opened 2017-09-27
warning: Overriding "Content-Type" header "application/json" with "multipart/form-data


chrisroos:
We see a number of these warnings when running the tests in this project (with `bundle exec rake`).

I believe it's because we're setting the `Content-Type` of _all_ requests to `application/json` and that this is being overridden when we call the `create_asset` and `create_whitehall_asset` methods.

chrisroos:
I'm going to take a look at this now.

chrisroos:
I've opened PR #758 to fix this warning.

chrisroos:
I've now merged PR #758 so I'm going to mark this as resolved.

PR #758, opened 2017-09-28
Fix Content-Type warning

chrisroos:
This change removes the following warning we were seeing when running the tests in this project, and when using the `AssetManager#create_asset`, `#create_whitehall_asset` and `#update_asset` methods in other project:

```
warning: Overriding "Content-Type" header "application/json" with "multipart/form-data; boundary=----RubyFormBoundaryX7Na6WDQqG3kLfD7" due to payload
```

Addresses issue #757.

### DIFF
--- a/lib/gds_api/json_client.rb
+++ b/lib/gds_api/json_client.rb
@@ -59,9 +59,13 @@ module GdsApi
     end
 
     def self.default_request_with_json_body_headers
-      self.default_request_headers.merge(
+      self.default_request_headers.merge(self.json_body_headers)
+    end
+
+    def self.json_body_headers
+      {
         'Content-Type' => 'application/json',
-      )
+      }
     end
 
     DEFAULT_TIMEOUT_IN_SECONDS = 4
@@ -134,6 +138,9 @@ module GdsApi
     #                  from the Net::HTTPResponse
     def do_json_request(method, url, params = nil, additional_headers = {}, &create_response)
       begin
+        if params
+          additional_headers.merge!(self.class.json_body_headers)
+        end
         response = do_request_with_cache(method, url, (params.to_json if params), additional_headers)
       rescue RestClient::Exception => e
         # Attempt to parse the body as JSON if possible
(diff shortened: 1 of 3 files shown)


lib/gds_api/json_client.rb
```diff
@@ -246,7 +246,11 @@ def do_request(method, url, params = nil, additional_headers = {})
       when :get, :delete
         default_headers = self.class.default_request_headers
       else
-        default_headers = self.class.default_request_with_json_body_headers
+        if params.respond_to?(:has_key?) && params[:multipart] == true
```
floehopper: Minor: It seems slightly odd to check for the existence of the `has_key?` method. Presumably you can't use `params.respond_to?(:[])` here because a `String` also responds to that method...? If so, I think it might make more sense to use `params.is_a?(Hash)` in this case.

lib/gds_api/json_client.rb
```diff
@@ -242,20 +239,9 @@ def do_request(method, url, params = nil, additional_headers = {})
         url: url,
       }
 
-      case method
-      when :get, :delete
```
floehopper: Minor: I found this commit a bit hard to follow. I think it might help to explain why you no longer need this condition and how come all the tests still pass.

lib/gds_api/json_client.rb
floehopper: It looks like this code is deleted in the next commit, so you can probably ignore my previous comment.

chrisroos:
Thanks for reviewing, @floehopper. I'm going to address your feedback now.

lib/gds_api/json_client.rb
chrisroos: I've updated the commit message in an attempt to better explain the change.

chrisroos:
> You might want to incorporate a version bump in this PR otherwise you'll need to open another one to get a new version of the gem released.

I'm going to do this in another branch/PR as per the instructions in https://github.com/alphagov/styleguides/blob/master/rubygems.md#versioning.

lib/gds_api/json_client.rb
floehopper: Looks great! 👍

PR #759, opened 2017-09-28
Bump version to 49.3.1

chrisroos:
I think that fixing a warning counts as a backwards-compatible bug fix so I've incremented the PATCH version as per [Semantic Versioning 2.0.0][1].

[1]: http://semver.org/


chrisroos:
Thanks, @floehopper! Merging.

PR #761, opened 2017-10-03
Bump version to 49.3.2

floehopper:
The only change is a documentation change (#760) and so it seems OK to make this a patch release.

### DIFF
--- a/lib/gds_api/version.rb
+++ b/lib/gds_api/version.rb
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '49.3.1'.freeze
+  VERSION = '49.3.2'.freeze
 end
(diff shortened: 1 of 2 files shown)


lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '49.3.1'.freeze
+  VERSION = '49.3.2'.freeze
```
cbaines: My interpretation on semver is that this is a minor change, and not a patch as it doesn't fix any bugs [1], but I'm happy either way, as the versioning on this repository isn't consistent or very important anyway.

1: `PATCH version when you make backwards-compatible bug fixes.`

floehopper: @cbaines Fair enough. I'll make it a minor version change. Thanks.

PR #765, opened 2017-10-18
Bump version to 49.6.0

chrisroos:
This is a minor version increase according to Semantic Versioning[1]
because it includes the addition of functionality in a
backwards-compatible manner.

[1]: http://semver.org/


chrisroos:
Thanks @chrislo!

PR #768, opened 2017-11-24
Add search_enum to rummager

koetsier:
Adds a method to get search results from rummager as an enumerator.
This abstracts away rummager's paging mechanism and fetches new
pages as more results are requested.

Use case: https://github.com/alphagov/content-tagger/pull/519#discussion_r152638734

### DIFF
--- a/lib/gds_api/rummager.rb
+++ b/lib/gds_api/rummager.rb
@@ -14,6 +14,30 @@ module GdsApi
       get_json(request_url, additional_headers)
     end
 
+    # Perform a search, returning the results as an enumerator.
+    #
+    # The enumerator abstracts away rummager's pagination and fetches new pages when
+    # necessary.
+    #
+    # @param args [Hash] A valid search query. See Rummager documentation for options.
+    # @param page_size [Integer] Number of results in each page.
+    #
+    # @see https://github.com/alphagov/rummager/blob/master/docs/search-api.md
+    def search_enum(args, page_size: 100, additional_headers: {})
+      Enumerator.new do |yielder|
+        (0..Float::INFINITY).step(page_size).each do |index|
+          search_params = args.merge(start: index.to_i, count: page_size)
+          results = search(search_params, additional_headers).to_h.fetch('results', [])
+          results.each do |result|
+            yielder << result
+          end
+          if results.count < page_size
+            break
+          end
+        end
+      end
+    end
+
     # Advanced search.
     #
     # @deprecated Only in use by Whitehall. Use the `#search` method.
--- a/lib/gds_api/rummager.rb
+++ b/lib/gds_api/rummager.rb
@@ -8,7 +8,7 @@ module GdsApi
     #
     # @param args [Hash] A valid search query. See Rummager documentation for options.
     #
-    # @see https://github.com/alphagov/rummager/blob/master/docs/search-api.md
+    # @see https://github.com/alphagov/rummager/blob/master/doc/search-api.md
     def search(args, additional_headers = {})
       request_url = "#{base_url}/search.json?#{Rack::Utils.build_nested_query(args)}"
       get_json(request_url, additional_headers)
@@ -22,7 +22,7 @@ module GdsApi
     # @param args [Hash] A valid search query. See Rummager documentation for options.
     # @param page_size [Integer] Number of results in each page.
     #
-    # @see https://github.com/alphagov/rummager/blob/master/docs/search-api.md
+    # @see https://github.com/alphagov/rummager/blob/master/doc/search-api.md
     def search_enum(args, page_size: 100, additional_headers: {})
       Enumerator.new do |yielder|
         (0..Float::INFINITY).step(page_size).each do |index|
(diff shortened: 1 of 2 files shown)


lib/gds_api/rummager.rb
```diff
@@ -14,6 +14,30 @@ def search(args, additional_headers = {})
...
 
+    # Perform a search, returning the results as an enumerator.
+    #
+    # The enumerator abstracts away rummager's pagination and fetches new pages when
+    # necessary.
+    #
+    # @param args [Hash] A valid search query. See Rummager documentation for options.
+    # @param page_size [Integer] Number of results in each page.
+    #
+    # @see https://github.com/alphagov/rummager/blob/master/docs/search-api.md
```
davidbasalla: this link 404's, probably needs to change '/docs' -> '/doc'

lib/gds_api/rummager.rb
```diff
@@ -14,6 +14,30 @@ def search(args, additional_headers = {})
...
+    def search_enum(args, page_size: 100, additional_headers: {})
+      Enumerator.new do |yielder|
```
davidbasalla: What's the benefit of having an `Enumerator` object here? I've not seen this used much before

test/rummager_test.rb
```diff
@@ -138,6 +138,39 @@
...
+  it "#search_enum returns two pages - last page is full" do
+    stub_request(:get, /example.com\/search/)
+      .with(query: hash_including(start: '0', count: '2'))
+      .to_return(body: { 'results' => [{ 'title' => 't1' }, { 'title' => 't2' }] }.to_json)
+
+    stub_request(:get, /example.com\/search/)
+      .with(query: hash_including(start: '2', count: '2'))
+      .to_return(body: { 'results' => [{ 'title' => 't3' }, { 'title' => 't4' }] }.to_json)
+
+    stub_request(:get, 'http://example.com/search.json')
```
davidbasalla: The other stubs are for `/example.com\/search/`. Could that be used here too? (for conistency...)

lib/gds_api/rummager.rb
```diff
@@ -14,6 +14,30 @@ def search(args, additional_headers = {})
...
+    # Perform a search, returning the results as an enumerator.
+    #
+    # The enumerator abstracts away rummager's pagination and fetches new pages when
+    # necessary.
+    #
+    # @param args [Hash] A valid search query. See Rummager documentation for options.
+    # @param page_size [Integer] Number of results in each page.
+    #
+    # @see https://github.com/alphagov/rummager/blob/master/docs/search-api.md
+    def search_enum(args, page_size: 100, additional_headers: {})
```
davidbasalla: Why does `page_size` default to `100` here?

davidbasalla:
I've left a couple of comments and questions. What's the use case for this new method? I don't know Rummager that well, so might be good to get some 👀  from @cbaines or @MatMoore on this too

lib/gds_api/rummager.rb
koetsier: Means you don't have to bother with pagination. You can use them to iterate and get results lazily, ie. `rummager.search_enum.each { |result| do_something_with(result)}` - the results will be retrieved only when needed.

lib/gds_api/rummager.rb
koetsier: Seemed like a reasonable default.

MatMoore:
Sorry I'm late to the party here but I have some thoughts:

The `TaxonomyQuery` class in the associated PR looks a lot like what I tried to do with https://github.com/alphagov/govuk_taxonomy_helpers - do you think these could be combined?

I would strongly encourage extracting code like this out of content tagger and into a shared gem so it can be reused. The taxonomy is an essential building block and it should be easy to integrate it into any publishing app, any rendering app, search, email, static, feedex, whatever. I don't think anything content tagger does with the taxonomy should be considered unique to content tagger.

I think adding `search_enum` to gds-api-adapters is useful, but using the search API in this way is quite unusual so I have concerns about how it's used. I would recommend thinking about:

- What depends on this functionality? if search goes down, does/should it break user workflows?
- In the general case, this method could make a huge number of API calls to go through all the pages - resulting in a really slow response. This would be really bad in the context of an http request where a user is waiting for it to complete. Can we make it clearer when this should and shouldn't be used?  Maybe the method name should be `search_enum!` with a bang?
- If this does cause performance problems, how do we monitor that?
- Are you confident that this enumerator gives stable results? As in, is it guaranteed that every document that could be matched in that query gets included in exactly one page?
- Are these responses cached/should they be?

In general I would be wary of treating rummager as a reliable and complete data source right now, compared to the publishing api and content store. [Draft content and document-level tagging to draft content isn't available right now](https://github.com/alphagov/govuk-rfcs/pull/86) and the data in rummager is not going to be reflective of what's in the publishing api unless it's been migrated to the [new indexing process](https://github.com/alphagov/rummager/blob/master/doc/new-indexing-process.md). 

I'm on holiday until wednesday so I can't answer any of these questions myself, but feel free to ping the search team about any of this 🙂 🏖

PR #773, opened 2017-11-27
Add GdsApi::EmailAlertApi#unsubscribe

bevanloon:
Add unsubscribe endpoint for Email Alert Api to allow Email Alert
Frontend to complete unsubscribe requests.

Relies on [PR 251](https://github.com/alphagov/email-alert-api/pull/251) in Email Alert Api being finalised and merged.

This PR currently assumes that Email Alert Api will return JSON, but if it does not, then we should update the tests here to reflect the actual returned body.

[Trello card](https://trello.com/c/Ba7F7Iny/389-add-functionality-to-email-alert-frontend-to-support-unsubscriptions)

### DIFF
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -93,6 +93,22 @@ module GdsApi
         )
       end
 
+      def email_alert_api_unsubscribes_a_subscription(uuid)
+        stub_request(:post, unsubscribe_url(uuid))
+          .with(body: "{}")
+          .to_return(status: 204)
+      end
+
+      def email_alert_api_has_no_subscription_for_uuid(uuid)
+        stub_request(:post, unsubscribe_url(uuid))
+          .with(body: "{}")
+          .to_return(status: 404)
+      end
+
+      def assert_unsubscribed(uuid)
+        assert_requested(:post, unsubscribe_url(uuid), times: 1)
+      end
+
     private
 
       def subscriber_lists_url(attributes = nil)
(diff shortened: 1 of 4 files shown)


lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -93,6 +93,19 @@ def email_alert_api_has_notification(notification)
...
+          .to_return(
+            status: 200,
+            body: { "deleted" => subscription_uuid }.to_json
+        )
+      end
+
+      def email_alert_api_does_not_unsubscribe_an_invalid_subscription(*)
+        stub_request(:post, unsubscribe_url)
+          .to_return(status: 404)
+      end
```
tuzz: Technically, this happens when a subscription can't be found for a UUID. Maybe the method name could be more like: `email_alert_api_has_no_subscription_for_uuid` or `email_alert_api_does_not_unsubscribe_a_missing_subscription` ? (I'm not very good at naming things)

bevanloon: yeah this sounds better thanks. I was struggling here as well

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -93,6 +93,19 @@ def email_alert_api_has_notification(notification)
         )
       end
 
+      def email_alert_api_unsubscribes_a_subscription(subscription_uuid)
+        stub_request(:post, unsubscribe_url)
+          .to_return(
+            status: 200,
+            body: { "deleted" => subscription_uuid }.to_json
```
tuzz: We updated this endpoint to return a `204` with no body. (sorry for the faff)

PR #777, opened 2017-11-30
add #get_paged_content_items to publishing_api_v2

koetsier:
Adds a method to get content items from the publishing api as an enumerator.
This abstracts away the  paging mechanism and fetches new pages as more 
results are requested.

Use case: https://trello.com/c/ZIOFIqoT/162-bulk-tag-content-items-that-belongs-to-a-single-taxon

needs pact in: https://github.com/alphagov/publishing-api/pull/1073

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -315,6 +315,28 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     get_json("#{endpoint}/v2/content#{query}")
   end
 
+  # Returns an Enumerator of content items for the provided
+  # query string parameters.
+  #
+  # @param params [Hash]
+  #
+  # @return [Enumerator] an enumerator of content items
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2content
+  def get_content_items_enum(params)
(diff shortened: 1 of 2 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -315,6 +315,28 @@ def get_content_items(params)
...
 
+  # Returns an Enumerator of content items for the provided
+  # query string parameters.
+  #
+  # @param params [Hash]
+  #
+  # @return [Enumerator] an enumerator of content items
+  #
+  # @see https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2content
+  def get_paged_content_items(params, per_page: 50)
```
cbaines: Given that this is a wrapper around `get_content_items` that provides an enumerator, I wonder if a name like `get_content_items_enumerator` might be clearer?

While the internals has something to do with pagination, compared to the `get_content_items` function, this is the one that involves no direct interaction with the pagination, so keeping that out of the name would be nice I think.

cbaines: Also, given that per_page is a parameter that `get_content_items` accepts, it might be more consistent to handle it through the params hash, rather than as a separate argument?

PR #778, opened 2017-12-04
Add subscribe method to email-alert-api

gpeng:
This PR adds a method to the email-alert-api client to support creating subscriptions. Also adds test helpers.

[Trello](https://trello.com/c/YtmHgU3b/408-add-page-to-email-alert-frontend-to-capture-a-users-email-address)

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -97,6 +97,17 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     post_json("#{endpoint}/unsubscribe/#{uuid}")
   end
 
+  # Subscribe
+  #
+  # @return [Hash] subscription_id
+  def subscribe(subscribable_id:, address:)
+    post_json(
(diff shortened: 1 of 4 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -97,6 +97,15 @@ def unsubscribe(uuid)
     post_json("#{endpoint}/unsubscribe/#{uuid}")
   end
 
+  # Subscribe
+  # #
+  # @param attributes [Hash] address, subscribable_id
+  #
+  # @return [Hash] subscription_id
+  def subscribe(attributes)
```
thomasleese: I'd be tempted to make this `def subscribe(subscribable_id:, address:)` so we get some documentation and we can leave it Ruby to handle the argument validation for us.

gpeng: I was making it consistent with the other methods but I agree... Fixed in 1f67417

PR #781, opened 2017-12-05
Add get_subscribable

gpeng:
This adds a method to call the `/subscribables/<reference>` endpoint on email-alert-api. The parameter has been called `reference` here although it is currently the `gov_delivery_id`. We will be renaming that so to reduce the number of changes in the clients I've used `reference`.

This endpoint is required for email-alert-frontend to enable lookup of subscribable title and id. https://github.com/alphagov/email-alert-api/pull/273 has a bit more context.

[Trello](https://trello.com/c/YtmHgU3b/408-add-page-to-email-alert-frontend-to-capture-a-users-email-address)


test/email_alert_api_test.rb
```diff
@@ -321,4 +321,30 @@
...
+        }
+      )
+      api_response = api_client.get_subscribable(reference: "test123")
+      assert_equal(200, api_response.code)
+      parsed_body = api_response.to_h
+      assert_equal(1, parsed_body["subscribable"]["id"])
+    end
+  end
+
+  describe "get_subscribable when one doesn't exsist" do
```
rubenarakelyan: Exist :)

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Unreleased
```
kevindew: Probably want to have a "##" here

gpeng: Thanks... fixed in #977b132

test/email_alert_api_test.rb
gpeng: Whoops... Fixed in 7bc127c

PR #783, opened 2017-12-07
Add document_type list and summary endpoints for support-api

elenatanasoiu:
Adds endpoints for fetching document type from support-api:

- the GET `document_type_list` endpoint of `support-api` which returns a list of document types. This will be shown in a dropdown to choose from in Feedex (`support` app) in order to filter feedback
- the GET `document_type_summary` endpoint of `support-api` which presents Feedex users with an overview of which pages have had the most feedback left recently on them (`feedback counts`)

As implemented in alphagov/support-api#121

For: https://trello.com/c/IB9P5EyE/266-introduce-new-browse-options-for-feedex

### DIFF
--- a/lib/gds_api/test_helpers/support_api.rb
+++ b/lib/gds_api/test_helpers/support_api.rb
@@ -80,13 +80,22 @@ module GdsApi
       end
 
       def stub_support_api_organisations_list(response_body = nil)
-        response_body ||= [{
-          slug: "cabinet-office",
-          web_url: "https://www.gov.uk/government/organisations/cabinet-office",
-          title: "Cabinet Office",
-          acronym: "CO",
-          govuk_status: "live"
-        }]
+        response_body ||= [
+          {
+            slug: "cabinet-office",
+            web_url: "https://www.gov.uk/government/organisations/cabinet-office",
+            title: "Cabinet Office",
+            acronym: "CO",
+            govuk_status: "live"
+          },
+          {
+            slug: "gds",
+            web_url: "https://www.gov.uk/government/organisations/gds",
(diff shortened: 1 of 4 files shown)


test/support_api_test.rb
```diff
@@ -61,12 +61,12 @@
   describe "GET /anonymous-feedback" do
     it "fetches anonymous feedback" do
       stub_get = stub_support_api_anonymous_feedback(
-        path_prefix: "/vat-rates",
+        path_prefixes: "/vat-rates",
```
h-lame: Shouldn't this be `path_prefixes: ["/vat-rates"]` ?

lib/gds_api/test_helpers/support_api.rb
```diff
@@ -80,22 +80,31 @@ def stub_support_api_anonymous_feedback_organisation_summary(slug, ordering = ni
...
         stub_http_request(:get, "#{SUPPORT_API_ENDPOINT}/organisations").
           to_return(status: 200, body: response_body.to_json)
       end
 
       def stub_support_api_organisation(slug = "cabinet-office", response_body = nil)
         response_body ||= {
-          slug: slug,
-          web_url: "https://www.gov.uk/government/organisations/#{slug}",
+          slug: "cabinet-office",
+          web_url: "https://www.gov.uk/government/organisations/cabinet-office",
```
h-lame: This doesn't seem right, I think we want to use the `slug` param if it's passed in.

elenatanasoiu: Put back slug. 10x!

PR #792, opened 2018-01-11
Add ability to disable JSON API cache from environment var

adrianclay:
As part of the publishing end to end tests caching of GET requests to the content-store is preventing updated content from being displayed on frontends.

### DIFF
--- a/lib/gds_api/json_client.rb
+++ b/lib/gds_api/json_client.rb
@@ -39,8 +39,9 @@ module GdsApi
       end
 
       @logger = options[:logger] || NullLogger.instance
+      disable_cache = options[:disable_cache] || ENV.fetch("DISABLE_JSON_API_CACHE", false)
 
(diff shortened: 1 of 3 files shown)


lib/gds_api/json_client.rb
```diff
@@ -39,8 +39,9 @@ def initialize(options = {})
       end
 
       @logger = options[:logger] || NullLogger.instance
+      disable_cache = options[:disable_cache] || ENV.fetch("DISABLE_JSON_API_CACHE", false)
 
-      if options[:disable_cache] || (options[:cache_size] == 0)
+      if disable_cache || options[:cache_size] == 0 # rubocop:disable Style/NumericPredicate
```
kevindew: D'oh bit annoying situation there. We should be ok using safe navigation operator but this Gem still supports 2.1. If I get a mo I'll switch it around to be testing against newer rubies and set minimum support of 2.3

PR #794, opened 2018-01-11
Percent encode the legacy_url_path when requesting Whitehall assets

chrisroos:
We've found a number of Whitehall assets whose filenames (and therefore
their legacy_url_path) contain non-ascii characters.

We need to URI encode the legacy_url_path to avoid `RestClient`, and
then the gds-api-adapters, from raising an InvalidUrl exception (as per
this example in Sentry[1]). Addresses issue #384.

I've used the Addressable Gem to avoid the Rubocop violation warning
that I saw when making a similar change to Asset Manager in
https://github.com/alphagov/asset-manager/pull/396.

[1]: https://sentry.io/govuk/app-whitehall/issues/429050852/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Percent encode URLs when requesting Whitehall assets from Asset Manager API
+
 # 50.9.0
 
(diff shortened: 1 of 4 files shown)


chrisroos:
Thanks @floehopper. I'll get this merged now.

PR #800, opened 2018-01-18
Add frequency param to email subscriptions

bevanloon:
Email alert api now accepts frequency (immediately, daily, weekly)
as a parameter through the `subscription/new` endpoint.
As a result, we should pass frequency through the api adapters.

[Trello](https://trello.com/c/ltl3PIhp/532-add-subscription-type-parameters-to-gds-api-adapters)

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -100,11 +100,12 @@ class GdsApi::EmailAlertApi < GdsApi::Base
   # Subscribe
   #
   # @return [Hash] subscription_id
-  def subscribe(subscribable_id:, address:)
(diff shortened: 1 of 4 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -100,11 +100,12 @@ def unsubscribe(uuid)
   # Subscribe
   #
   # @return [Hash] subscription_id
-  def subscribe(subscribable_id:, address:)
+  def subscribe(subscribable_id:, address:, frequency:)
```
thomasleese: Should we make it default to `immediately`? (Then this doesn't have to be a breaking change)

bevanloon: yep - fixed in 2f59e73

PR #805, opened 2018-01-30
Helper method to determine destination of redirect

kevindew:
This adds in the ability for the GdsApi::ContentStore class to be able to be given a content_item and return the redirect destination and status code. 

This has been added to GDS API Adapters as this would allow this method to be used across multiple apps if needed.

As this replicates some of routers logic it does feel contentious and is at risk of being inconsistent if router is ever to change it's redirect logic

## Why is this needed

The reason for this is to work around an atomicity issue we have with router and frontend apps. Which is outlined here: https://github.com/alphagov/government-frontend/issues/437 for gone responses. Unlike gone responses a redirect requires logic to determine the destination.

This issue is flagged in sentry for government-frontend a few hundred times: https://sentry.io/govuk/app-government-frontend/?query=is%3Aunresolved++redirect

This issue is more acute with [publishing-e2e-tests](https://github.com/alphagov/publishing-e2e-tests) where it can frequently cause a test fail

## What are the alternatives 

- The main alternative is to continue as we are and to return users a 500 response when this situation occurs. End-to-end tests could be modified for that. We could stop sending sentry errors (though this doesn't feel great)
- It doesn't seem practical to try introduce atomicity into router and content-store

## Related PRs for implementation

- [ ] https://github.com/alphagov/government-frontend/pull/727
- [ ] https://github.com/alphagov/govuk-content-schemas/pull/710
- [ ] https://github.com/alphagov/content-store/pull/369

### DIFF
--- a/lib/gds_api/content_store.rb
+++ b/lib/gds_api/content_store.rb
@@ -1,3 +1,5 @@
+require "plek"
+
 require_relative 'base'
 require_relative 'exceptions'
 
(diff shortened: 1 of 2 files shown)


lib/gds_api/content_store.rb
```diff
@@ -18,9 +20,85 @@ def content_item!(_)
...
   end
+
+  class RedirectResolver
+    def initialize(content_item, request_path, request_query)
+      @content_item = content_item
+      @request_path = request_path
+      @request_query = request_query
+    end
+
+    def self.call(content_item, request_path, request_query = "")
```
thomasleese: I would use `*args` here to avoid having to update this is the args to `initialize` change.

kevindew: Oh cool, does that work for all argument types? I had it as just for hashes in my head. Will give it a spin.

PR #809, opened 2018-02-26
Add support for /feedback-by-day endpoint.

peteglondon:
This is needed for content-performance-manager which
needs to get the number of feedback comments per document 
per day to store in the data warehouse.

The Trello card is [Expose Feedex numbers via Support API](https://trello.com/c/6nz6zXEE/113-5-expose-feedex-numbers-via-support-api)

There is a [related PR in support-api](https://github.com/alphagov/support-api/pull/174)

### DIFF
--- a/lib/gds_api/test_helpers/support_api.rb
+++ b/lib/gds_api/test_helpers/support_api.rb
@@ -30,6 +30,12 @@ module GdsApi
         post_stub.to_return(status: 202)
       end
 
+      def stub_support_api_feedback_by_day(date, page, per_page, response_body = { results: [] })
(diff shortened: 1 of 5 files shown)


lib/gds_api/test_helpers/support_api.rb
```diff
@@ -30,6 +30,12 @@ def stub_support_api_feedback_export_request_creation(request_details = nil)
         post_stub.to_return(status: 202)
       end
 
+      def stub_support_api_feedback_by_day(date, page, per_page)
```
pmanrubia: What about adding a way to stub the results?

PR #811, opened 2018-03-09
Remove deprecated GovDelivery methods

thomasleese:
These will be removed from email-alert-api shortly so we should start by removing them from gds-api-adapters.

[Trello Card](https://trello.com/c/iKEAnrN2/680-post-deploy-development-cleanup)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Remove deprecated `notifications` and `notification` methods from `GdsApi::EmailAlertApi`.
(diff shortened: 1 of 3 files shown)


thomasleese:
Hmm, I would say we can't be 100% sure that it won't as there may be uses we're not immediately aware of - so I would say the change itself is a potentially breaking one.

tijmenb:
Okido (wrote up https://trello.com/c/Ik8s0rvh/115-fix-gds-api-adapters instead).

PR #817, opened 2018-03-16
Add new email-alert-api endpoints

rubenarakelyan:
This commit adds three new email-alert-api endpoints to list a user’s subscriptions, change an email address and change a subscription’s frequency.

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -105,8 +105,41 @@ class GdsApi::EmailAlertApi < GdsApi::Base
   #  subscriber_count
   # }
   def get_subscribable(reference:)
-    get_json(
-      "#{endpoint}/subscribables/#{reference}"
+    get_json("#{endpoint}/subscribables/#{reference}")
+  end
+
+  # Get Subscriptions for a Subscriber
+  # #
+  # @param string Subscriber address
+  #
+  # @return [Hash] subscriber, subscriptions
+  def get_subscriptions(address:)
+    get_json("#{endpoint}/subscribers/#{address}/subscriptions")
+  end
+
+  # Patch a Subscriber
+  # #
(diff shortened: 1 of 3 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -105,8 +105,41 @@ def subscribe(subscribable_id:, address:, frequency: "immediately")
...
+    get_json("#{endpoint}/subscribers/#{address}/subscriptions")
+  end
+
+  # Patch address for a Subscriber
+  # #
+  # @param string Subscriber old_address
+  # @param string Subscriber new_address
+  #
+  # @return [Hash] subscriber
+  def change_subscriber_address(old_address:, new_address:)
```
thomasleese: I would call this `change_subscriber(address:, new_address:)` in case we add more fields to the `PATCH` endpoint in the future.

lib/gds_api/email_alert_api.rb
```diff
@@ -105,8 +105,41 @@ def subscribe(subscribable_id:, address:, frequency: "immediately")
...
+    )
+  end
+
+  # Patch frequency for a Subscription
+  # #
+  # @param string Subscription subscription_id
+  # @param string Subscription frequency
+  #
+  # @return [Hash] subscription
+  def change_subscription_frequency(subscription_id:, frequency:)
```
thomasleese: Similarly, I would call this `change_subscription(id:, frequency:)`.

lib/gds_api/email_alert_api.rb
rubenarakelyan: 👍

lib/gds_api/email_alert_api.rb
rubenarakelyan: 👍

PR #822, opened 2018-03-22
Add stub for get_links_for_content_ids

benthorner:


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -448,6 +448,35 @@ def publishing_api_has_expanded_links(links, with_drafts: true)
         stub_request(:get, url).to_return(status: 200, body: links.to_json, headers: {})
       end
 
+      # Stub a request to get links for content ids
+      #
+      # @param [Hash] links the links for each content id
+      #
+      # @example
+      #   publishing_api_get_links_for_content_ids(
```
cbaines: It says `..._get_...` here, I'm assuming that is not quite right?

PR #824, opened 2018-03-23
Change subscription management endpoints to use IDs

rubenarakelyan:
This commit changes the two endpoints created for email-alert-frontend to accept a subscriber ID rather than an email address since we’ll be using the ID as an email address can change. The endpoints are not currently in use so they can be changed.

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -127,22 +127,22 @@ class GdsApi::EmailAlertApi < GdsApi::Base
 
   # Get Subscriptions for a Subscriber
   # #
-  # @param string Subscriber address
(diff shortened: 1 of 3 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -127,22 +127,22 @@ def get_subscription(id)
 
   # Get Subscriptions for a Subscriber
   # #
-  # @param string Subscriber address
+  # @param string Subscriber id
```
kevindew: This should be updated from string (and there's another instance below)

rubenarakelyan: Ah yes, fixed

PR #826, opened 2018-03-23
Add email-alert-api unsubscribe_all endpoint

rubenarakelyan:
This commit adds the `unsubscribe_all` endpoint from email-alert-api which accepts a subscriber ID and unsubscribes them from all their subscriptions.

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -68,13 +68,22 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     get_json("#{endpoint}/topic-matches.json?#{query_string}")
   end
 
-  # Unsubscribe
+  # Unsubscribe subscriber from subscription
   # #
   # @param uuid Subscription uuid
   #
-  # @return [Hash] deleted
+  # @return null
   def unsubscribe(uuid)
     post_json("#{endpoint}/unsubscribe/#{uuid}")
+  end
+
+  # Unsubscribe subscriber from everything
+  # #
+  # @param integer Subscriber id
+  #
+  # @return null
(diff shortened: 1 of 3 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -77,6 +77,15 @@ def unsubscribe(uuid)
     post_json("#{endpoint}/unsubscribe/#{uuid}")
   end
 
+  # Unsubscribe from everything
+  # #
+  # @param integer Subscriber id
+  #
+  # @return [Hash] deleted
+  def unsubscribe_all(id)
```
thomasleese: I would call this `unsubscribe_subscriber` (and then I would call the other one `unsubscribe_subscription` but that's already in use :cry:), but I don't really mind.

lib/gds_api/email_alert_api.rb
```diff
@@ -77,6 +77,15 @@ def unsubscribe(uuid)
     post_json("#{endpoint}/unsubscribe/#{uuid}")
   end
 
+  # Unsubscribe from everything
+  # #
+  # @param integer Subscriber id
+  #
+  # @return [Hash] deleted
```
kevindew: Should this be returning a hash? Not really sure what it does return for a 204 maybe it's an empty object :-/

lib/gds_api/email_alert_api.rb
rubenarakelyan: Done :)

lib/gds_api/email_alert_api.rb
rubenarakelyan: I've marked it as `null` since that seems to be the closest thing

PR #828, opened 2018-03-27
Create auth token method for Email Alert API

kevindew:
Trello: https://trello.com/c/6YmlH3Mq/725-add-token-based-login-for-the-subscription-management-interface-to-email-alert-frontend

This adds a convenience method and a stub for the POST
/subscribers/auth-token method on Email Alert API.

This is DNM until https://github.com/alphagov/email-alert-api/pull/560 has been merged and deployed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Add `create_auth_token` to `GdsApi::EmailAlertApi`.
+
 # 52.4.0
 
 * Add `unsubscribe_subscriber` to `GdsApi::EmailAlertApi`.
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -169,6 +169,23 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     )
   end
 
+  # Create an authentication token for a subscriber
+  #
+  # @param [string]       address       Email address of subscriber to create token for
+  # @param [string]       destination   Path on GOV.UK that subscriber will be emailed
+  # @param [string, nil]  redirect      Path on GOV.UK to be encoded into the token for redirecting
+  #
+  # @return [Hash]  subscriber
+  #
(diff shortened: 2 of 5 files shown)


tijmenb:
This is probably not really the place for this discussion, but I feel it's probably okay for these methods to go into email-alert-frontend itself. Like publishing-api has a [few extra content-store API adapters methods](https://github.com/alphagov/publishing-api/blob/master/app/adapters/content_store.rb ) that must only be used inside the publishing-api. The auth token thing is probably never going to get used outside of email-alert-frontend, so there's a case for colocating them.

kevindew:
> This is probably not really the place for this discussion, but I feel it's probably okay for these methods to go into email-alert-frontend itself. Like publishing-api has a few extra content-store API adapters methods that must only be used inside the publishing-api. The auth token thing is probably never going to get used outside of email-alert-frontend, so there's a case for colocating them.

Sure, you're right that these won't likely get used outside of Email Alert Frontend as well as a variety of them. 

In my opinion it's mostly superfluous even having methods for most of these since we basically name a method based on an endpoint and have it's args as arguments you may as well just call something like `email_alert_api_adapter.post_json("/subscribers/auth-token", *args)` and skip the need for a method - although we get caught up on the [endpoint method to do so](https://github.com/alphagov/email-alert-frontend/blob/a0584370ebe37f8314edd4e194c3ced479bdaefd/app/controllers/authentication_controller.rb#L17-L22)

As for putting an adapter in specially, the Publishing API one seems to be a specific thing that wraps the GDS API adapters. Whereas to do this here we'd have to subclass the adapter class in Email Alert Frontend which feels like an introduction of indirection and inconsistency.

So in general I do agree with your concerns and would be pleased if we started to deal with API Adapters and these sort of issues. But I don't think there's a super quick route we can take here which doesn't introduce WTFs.

kevindew:
@tijmenb I've merged this in as it's quite time sensitive to get this going - but happy to chat more and change it around.

tijmenb:
Yes perfect!

PR #832, opened 2018-04-03
Enhance the special route publisher to check first

cbaines:
Previously it would always create a new draft (or update the existing
one), and then publish it.

Now it checks if the content or links have changed, and only performs
the appropriate update actions if this is the case.

This code is often invoked when deploying applications, and most of
the time it runs, it doesn't actually change the content or links in a
meaningful way. However, it does create a new needless edition, and
this does lead to unnecessary Publishing API activity (due to
dependency resolution/link expansion). This is especialy true for the
homepage.

Some of the motivation for this change can be seen here https://github.com/alphagov/govuk-app-deployment/pull/255 as that patch was a easier workaround for this issue in the case of the Frontend app. I'm hoping that if the SpecialRoutePublisher can be made smarter, then we can revert that change, and continue running it on deploy.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,6 +12,7 @@ Rake::TestTask.new("test") do |t|
   t.libs << "test"
   t.test_files = FileList["test/**/*_test.rb"]
   t.verbose = true
+  t.warning = false
 end
 task default: :test
 
--- a/lib/gds_api/publishing_api/special_route_publisher.rb
+++ b/lib/gds_api/publishing_api/special_route_publisher.rb
@@ -10,31 +10,30 @@ module GdsApi
       end
 
       def publish(options)
-        logger.info("Publishing #{options.fetch(:type)} route #{options.fetch(:base_path)}, routing to #{options.fetch(:rendering_app)}")
+        content_id = options.fetch(:content_id)
+        edition = content_hash(options)
 
-        put_content_response = publishing_api.put_content(
-          options.fetch(:content_id),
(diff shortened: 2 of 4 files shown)


cbaines:
> Rather than doing this, would it make sense to remove all of the deploy-time publishing? I think it happens very rarely that a special route page is updated or added, and it's easier to remember if we never publish on deploy at all.

I think this would be a good change. I've already removed it for frontend [1], but I was planning to put it back if this was merged.

Even if all the deploy time publishing was removed, I think this would still be useful, as many applications publish multiple special routes, so this change would avoid publishing needless editions if just one of those is changed, or if only the links are updated, ...

1: https://github.com/alphagov/govuk-app-deployment/pull/255

tijmenb:
If we remove the on-deploy thing, I wouldn't add this here. I think it might work but this seems like one of those things that will silently break and is really hard to debug.

cbaines:
> If we remove the on-deploy thing, I wouldn't add this here. I think it might work but this seems like one of those things that will silently break and is really hard to debug.

I've now put up a Pull Request to remove all the publishing from `govuk-app-deployment` [1]. I haven't looked in to the origins of this behaviour yet, so I'm a bit hesitant to remove it (hence why I initially did this).

1: https://github.com/alphagov/govuk-app-deployment/pull/256

tijmenb:
I think this is no longer necessary, since we removed the auto-publish on deploy. Feel free to reopen.

PR #833, opened 2018-04-05
Add a helper to update the links for a content_id

cbaines:


tijmenb:
I'm not entirely sure I get this. The example seems to show something that is doable with the current `patch_links` method. Where will this be used?

cbaines:
> I'm not entirely sure I get this. The example seems to show something that is doable with the current patch_links method. Where will this be used?

My hope is that this will allow simplifying cases where exactly this is done, but often without remembering to use the optimistic locking. The previous example I gave wasn't very useful, so I've updated it to a better one.

There are a couple of cases in Content Tagger where it may help simplify and improve the existing code [1][2].

1: https://github.com/alphagov/content-tagger/blob/e4423bfed8364074fe49645e0cb84137e516b6b9/app/services/tagging/untagger.rb
2: https://github.com/alphagov/content-tagger/blob/f43a884134e042afafca522c8c1631b7968ff7f2/app/services/tagging/tagger.rb

koetsier:
Getting the links with `get_links`,  updating them using `patch_links` using the versioning parameter does seem to always be used together.

If you're writing a function to combine these, maybe a retry option would be useful too ? I used this in the Tagger/Untagger code using

```
 rescue GdsApi::HTTPConflict
      retries ||= 0
      retry if (retries += 1) < max_retries
      raise
```

lib/gds_api/publishing_api_v2.rb
```diff
@@ -209,6 +209,50 @@ def get_links(content_id)
...
+  # @example
+  #
+  #   publishing_api.update_links("a-content-id") do |links|
+  #     {
+  #       'link_type' => (
+  #         links.fetch('link_type', []) | ['e10655d3-31eb-4909-84f7-5b7a2312cbfc']
+  #       )
+  #     }
+  #   end
+  def update_links(content_id, options = {})
```
tijmenb: Maybe naming this `get_and_patch_links` will make it clear that it's a combination of both?

lib/gds_api/publishing_api_v2.rb
```diff
@@ -209,6 +209,50 @@ def get_links(content_id)
...
+  #
+  # @return [GdsApi::Response] The response from the patch links
+  # request.
+  #
+  # @example
+  #
+  #   publishing_api.update_links("a-content-id") do |links|
+  #     {
+  #       'link_type' => (
+  #         links.fetch('link_type', []) | ['e10655d3-31eb-4909-84f7-5b7a2312cbfc']
```
tijmenb: I find this a little hard to follow because I always forget what `|` does. Maybe the specific use case from https://github.com/alphagov/content-tagger/blob/f43a884134e042afafca522c8c1631b7968ff7f2/app/services/tagging/tagger.rb#L4-L8 is a good example.

lib/gds_api/publishing_api_v2.rb
cbaines: Sure. I've updated the name.

lib/gds_api/publishing_api_v2.rb
cbaines: Ok, I've changed the example to use `+` and `uniq`.

tijmenb:
Closing because of inactivity, feel free to reopen!

PR #837, opened 2018-06-07
Add generate argument to `has_expanded_links` test helper

1pretz1:
https://trello.com/c/Ch9ufVJ4/139-bug-when-i-save-a-topic-selection-for-the-first-time-the-topics-dont-appear-on-the-overview-page

Now `publishing_api_has_expanded_links` can be called with `generate` as
an argument. This relates to the changes made in this pr for whitehall:
https://github.com/alphagov/whitehall/pull/4094.

`generate` defaults to `false` and in the real request would generate the
expanded links on the fly instead of getting them from stored data if set
to `true`.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -441,11 +441,16 @@ module GdsApi
       #           ]
       #         }
       #       }
-      def publishing_api_has_expanded_links(links, with_drafts: true)
+      def publishing_api_has_expanded_links(links, with_drafts: true, generate: false)
         links = deep_transform_keys(links, &:to_sym)
-        query = with_drafts ? "" : "?with_drafts=false"
-        url = PUBLISHING_API_V2_ENDPOINT + "/expanded-links/" + links[:content_id] + query
-        stub_request(:get, url).to_return(status: 200, body: links.to_json, headers: {})
+        request_params = {}
+        request_params['with_drafts'] = false if !with_drafts
+        request_params['generate'] = true if generate
+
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -441,9 +441,16 @@ def publishing_api_has_links(links)
       #           ]
       #         }
       #       }
-      def publishing_api_has_expanded_links(links, with_drafts: true)
+      def publishing_api_has_expanded_links(links, with_drafts: true, generate: false)
         links = deep_transform_keys(links, &:to_sym)
-        query = with_drafts ? "" : "?with_drafts=false"
+        if generate && !with_drafts
+          query = "?with_drafts=false&generate=true"
+        elsif generate && with_drafts || generate
```
koetsier: this is the same as `elsif generate`

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -441,9 +441,16 @@ def publishing_api_has_links(links)
       #           ]
       #         }
       #       }
-      def publishing_api_has_expanded_links(links, with_drafts: true)
+      def publishing_api_has_expanded_links(links, with_drafts: true, generate: false)
         links = deep_transform_keys(links, &:to_sym)
-        query = with_drafts ? "" : "?with_drafts=false"
+        if generate && !with_drafts
```
koetsier: Would this be easier to do as publishing_api_has_linked_items (line 528) - i.e. create a hash of parameters and then supply it to stub_request as a` with(query: hash_of_parameters)`

PR #840, opened 2018-07-23
Avoid using 'Link' header to unroll API pagination

benthorner:
The 'Link' header originally exposed by Whitehall is no longer reliable (we 
think this is due to a Fastly issue). Currently this header is used
to compact a set of paginated API responses into a single array result,
using the "next" and "previous" links contained within the header.

Pagination is an issue for the /organisations and /world-locations APIs,
both of which also have pagination fields in the body, as well as in the
special 'Link' header. Using the 'next_page_url' and 'previous_page_url'
fields to control the compaction process fixes the 'Link' header bug.

In trying to make this change, the API helpers for organisations and world
locations needed updating to include the additional fields. In order to
make this change easier, I've refactored the code by combining similar
inline conditionals into discrete blocks of code. The 'Link' header is
no longer in the stub response, to reflect that it's not reliable.

### DIFF
--- a/lib/gds_api/list_response.rb
+++ b/lib/gds_api/list_response.rb
@@ -25,30 +25,21 @@ module GdsApi
     end
 
     def has_next_page?
-      ! page_link("next").nil?
-    end
-
-    def next_page
-      # This shouldn't be a performance problem, since the cache will generally
-      # avoid us making multiple requests for the same page, but we shouldn't
-      # allow the data to change once it's already been loaded, so long as we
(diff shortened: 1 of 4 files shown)


benthorner:
@rubenarakelyan If this PR is accepted then we should remove the dodgy 'Link' header from collections.

benthorner:
The underlying issue appears to have fixed itself. Due to the lack of tests around the code, I think it's best to avoid making this change unless we have to.

PR #846, opened 2018-08-09
Add `send_unpublish_message` method to email-alert-api

1pretz1:
When called the method sends a request to the unpublishing-messages
endpoint in the email-alert-api. This endpoint starts the process of
notifying a user that their email subscription is no longer active.

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -57,6 +57,17 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     post_json("#{endpoint}/notifications", publication, headers)
   end
 
+  # Unpublishing alert
+  #
+  # @param message [Hash] content_id
+  #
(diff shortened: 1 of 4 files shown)


cbaines:
I always forget this, so don't feel bad, but could you add a note to the CHANGELOG.

1pretz1:
Sure 👍

cbaines:
> Sure +1

Perfect :star:

lib/gds_api/email_alert_api.rb
```diff
@@ -57,6 +57,13 @@ def send_alert(publication, headers = {})
     post_json("#{endpoint}/notifications", publication, headers)
   end
 
+  # Unpublishing alert
+  #
+  # @param message [Hash] base_path, content_id
```
cbaines: Maybe just say something about how it's the message the Publishing API sends about unpublished content? I'm not sure what "base_path, content_id" adds here.

PR #850, opened 2018-09-19
Add new test helpers for Asset Manager

emmabeynon:


benthorner:
This looks good. I've since added another TODO for deleting any assets, which is something supported by the test helper for the Publishing API. Could you add one for that, too?

https://github.com/alphagov/content-publisher/blob/c0de7133228830da64124c51dc060cc94b30d9c3/spec/features/delete_draft_asset_manager_down_spec.rb

emmabeynon:
@benthorner Done - let me know if you're happy with that?

PR #852, opened 2018-09-24
Create GdsApi factory methods

kevindew:
This creates helper methods for the GdsApi module for initialising
adapters with common options.

The reason for this is that we have lots of the same boilerplate code in
every GOV.UK application to achieve the basic initialisations in this
file. This feels that it's more the responsibility of the gem if every
client has the same code.

Rather than an application needing to have:

```
require 'gds_api/publishing_api_v2'

GdsApi::PublishingApiV2.new(
  Plek.new.find("publishing-api"),
  bearer_token: ENV["PUBLISHING_API_BEARER_TOKEN"] || "example"
)
```

They can instead do `GdsApi.publishing_api_v2`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Add methods to GdsApi to create instances of adapters with common options to reduce boilerplate code across apps
+
 # 53.1.0
 
 * Add Asset Manager test helpers: `asset_manager_update_asset`, `asset_manager_update_failure`, `asset_manager_delete_asset` and `asset_manager_delete_asset_failure`.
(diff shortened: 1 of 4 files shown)


cbaines:
This seems to overlap a bit with https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/helpers.rb , what do you think?

kevindew:
@cbaines oh wow, did not know that existed. Thanks for pointing that out.

I'll do a bit of digging into usage to decide whether to merge this into those helpers or to kill off those helpers for this.

tijmenb:
@kevindew I *think* I've removed most of the usage of those helpers at some point. I'd deprecate them and see what happens.

kevindew:
Sorry for the delay. Thanks @tijmenb I've had a look and there doesn't seem to be much usage like you suggested - the usages there are quite confusing from a "where does that method come from" perspective.

I've now marked GdsApi::Helpers as deprecated.

PR #864, opened 2018-11-06
Change GdsApi.organisations to use the public endpoint

cbaines:
The test helpers in the GDS API Adapters where changed in [1], but I
think it makes sense to change this as well for the same reasoning,
that depending on the public organisations API is more flexible than
Whitehall.

1: bd9d95911b5c389c691c14174a7d12d0b85fd8ee


cbaines:
This follows on from https://github.com/alphagov/gds-api-adapters/pull/857 to try and make GDS API Adapters a bit more consistent.

PR #865, opened 2018-11-13
Add page param to worldwide locations

bilbof:
The default behaviour is to return just the first page of results. This makes it possible to
get all pages via the same interface.

I didn't add a version bump because this is a backwards-compatible change. Let me know if you think changing the version would be worth doing!

```
# Before, only this works
GdsApi::Worldwide.world_locations

# After, this works
GdsApi::Worldwide.world_locations
GdsApi::Worldwide.world_locations page: 2
```

### DIFF
--- a/lib/gds_api/worldwide.rb
+++ b/lib/gds_api/worldwide.rb
@@ -1,8 +1,8 @@
 require_relative 'base'
 
 class GdsApi::Worldwide < GdsApi::Base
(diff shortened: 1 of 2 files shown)


bilbof:
For reference, this was closed because one can use `GdsApi.worldwide.world_locations.with_subsequent_pages` to paginate

PR #871, opened 2018-12-17
Change how the default logger is assigned

cbaines:
Previously, using `self.class.logger` meant that changes to
`GdsApi::Base.logger` wouldn't always apply to new clients.

I'm not to sure why, as Ruby is hard, but I am sure that this
behaviour is confusing. I also think it's unnecessary, as subclasses
can override the logger by passing it in as an option.

So, change this to use `GdsApi::Base.logger`, like how the
`default_options` works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Change how the default logger is assigned.
+
 # 55.0.0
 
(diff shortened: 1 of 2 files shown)


cbaines:
I worked around this in Smart Answers here https://github.com/alphagov/smart-answers/pull/3796 but I think changing the behaviour would be good.

PR #873, opened 2018-12-19
Extend default Content Store timeout for migration

benthorner:
https://trello.com/c/WIcZf3ZQ/393-investigate-high-smart-answers-5xx-rate

Following the migration of the Content Store to AWS, we are now seeing
increased response times from apps still hosted in Carrenza. This
increases the default time for Content Store during the migration, so
that our apps can cope with the longer response times.

### DIFF
--- a/lib/gds_api/content_store.rb
+++ b/lib/gds_api/content_store.rb
@@ -4,10 +4,17 @@ require_relative 'base'
 require_relative 'exceptions'
 
 class GdsApi::ContentStore < GdsApi::Base
+  EXTENDED_TIMEOUT_FOR_AWS = 8
+
   class ItemNotFound < GdsApi::HTTPNotFound
     def self.build_from(http_error)
       new(http_error.code, http_error.message, http_error.error_details)
     end
+  end
+
+  def initialize(endpoint_url, options = {})
+    options[:timeout] ||= EXTENDED_TIMEOUT_FOR_AWS
+    super(endpoint_url, options)
   end
 
   def content_item(base_path)


cbaines:
What effect will this have up the stack, I'm especially thinking of the NGinx timeouts?

lib/gds_api/content_store.rb
```diff
@@ -4,12 +4,19 @@
 require_relative 'exceptions'
 
 class GdsApi::ContentStore < GdsApi::Base
+  EXTENDED_TIMEOUT_FOR_AWS = 8
```
cbaines: Looking at the logs for smart-answers, these are the exceptions: `RestClient::Exceptions::OpenTimeout`. 

So we're saying it can take 4+ seconds to even establish a connection to the content store?

benthorner: Hmm...that is a bit suspicious. Sounds like it's worth some more investigation from us.

thomasleese:
@cbaines The default app timeout in Nginx is [set to 15 seconds](https://github.com/alphagov/govuk-puppet/blob/b588e4ade996e97b8975e69cb00800521fff4a48/modules/govuk/manifests/app/nginx_vhost.pp#L80) so there is still some space to play with, although we could look at increasing that. Ideally, we want to do as little as possible before the frontends move to AWS to get rid of the alerts so there is less to revert.

lib/gds_api/content_store.rb
thomasleese: I think @schmie mentioned the medium latency has gone from about 200ms to 420ms.

benthorner: I'm having trouble locating these exceptions - could you point me somewhere?

cbaines: This is the file I got them from: `/var/apps/smartanswers/log/production.api_client.log`. All the requests are logged, but the errors appear with `ERROR` in the line.

benthorner: OpenTimeout suggests we're backing up on serving requests. I notice that we've only got 2 unicorn workers on the AWS content stores, vs 8 on Carrenza. Shall we up it? Also, do you know where to get graphite metrics for AWS apps, as I don't have any actual data to back this theory up with.

benthorner:
Superseded by https://github.com/alphagov/govuk-puppet/pull/8458

PR #874, opened 2018-12-20
Catch SocketError

MuriloDalRi:
https://trello.com/c/D9f0qyLL/554-catch-socketerror-in-gds-api-adapters


kevindew:
The code looks good, It'd be good to update the commit message to explain why we're catching SocketError.

We should also update the changelog

MuriloDalRi:
I've updated the changelog and commit message @kevindew

PR #879, opened 2019-01-07
Pass bearer token to Link Checker API requests

emmabeynon:
https://trello.com/c/2IdzHdtp/26-add-authentication-to-link-checker-api

Link Checker API now authenticates POST requests so we pass the
LINK_CHECKER_API_BEARER_TOKEN to Link Checker API requests if it's present.


cbaines:
Currently, I'm trying to get some Pact changes through between gds-api-adapters and the Publishing API, which is probably causing some build failures here... sorry.

PR #882, opened 2019-01-07
Release version 57.0.0

emmabeynon:


cbaines:
We don't do versioning particularly rigorously, so it's not that important, but at least from the changelog, it doesn't look necessary to bump the major version for these changes.

PR #883, opened 2019-01-18
Add more Asset Manager test helpers

kevindew:
Trello: https://trello.com/c/5JMGsuBj/687-follow-up-work-from-versioning-review

This moves some Asset Manager test helpers that we mocked out directly in Content Publisher into GDS API Adapters. They provide ability to:

- Stub all Asset Manager calls
- Stub Asset Manager as being down
- Stub a batch of Asset Manager creates
- Stub a batch of Asset Manager updates
- Stub a batch of Asset Manager deletes

### DIFF
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -11,6 +11,16 @@ module GdsApi
         stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 503)
       end
 
+      def asset_manager_updates_any_asset(body = {})
+        stub_request(:put, %r{\A#{ASSET_MANAGER_ENDPOINT}/assets})
+          .to_return(body: body.to_json, status: 200)
+      end
+
+      def asset_manager_deletes_any_asset(body = {})
+        stub_request(:delete, %r{\A#{ASSET_MANAGER_ENDPOINT}/assets})
+          .to_return(body: body.to_json, status: 200)
+      end
+
       def asset_manager_has_an_asset(id, atts)
         response = atts.merge("_response_info" => { "status" => "ok" })
 
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -3,6 +3,14 @@ module GdsApi
     module AssetManager
       ASSET_MANAGER_ENDPOINT = Plek.current.find('asset-manager')
 
+      def stub_any_asset_manager_call
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 200)
+      end
+
+      def asset_manager_is_down
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 503)
+      end
+
       def asset_manager_has_an_asset(id, atts)
         response = atts.merge("_response_info" => { "status" => "ok" })
 
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -53,8 +53,31 @@ module GdsApi
           .to_return(body: response.to_json, status: 404)
       end
 
-      def asset_manager_receives_an_asset(response_url)
-        stub_request(:post, "#{ASSET_MANAGER_ENDPOINT}/assets").to_return(body: { file_url: response_url }.to_json, status: 200)
+      # This can take a string of an exact url or a hash of options
+      #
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -3,6 +3,32 @@ module TestHelpers
...
 
+      def stub_any_asset_manager_call
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 200)
+      end
+
+      def asset_manager_is_down
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 503)
+      end
+
+      def asset_manager_receives_assets(filename = nil)
```
benthorner: How does this one compare to `asset_manager_receives_an_asset`?

lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -3,6 +3,32 @@ module TestHelpers
...
+        stub_request(:post, "#{ASSET_MANAGER_ENDPOINT}/assets").to_return do
+          filename ||= SecureRandom.hex(8)
+          file_url = "#{ASSET_MANAGER_ENDPOINT}/media/#{SecureRandom.uuid}/#{filename}"
+          { body: { file_url: file_url }.to_json, status: 200 }
+        end
+      end
+
+      def asset_manager_updates_assets(body = {})
```
benthorner: Should this be 'any_asset'?

kevindew:
@benthorner both good points, I think I did this a bit more lift and shift than I should have - will have a quick ponder.

lib/gds_api/test_helpers/asset_manager.rb
kevindew: @benthorner So I've been in a bit of a wrestle with this. The new stub I created `asset_manager_receives_assets` is effectively the same as `asset_manager_receives_an_asset` (like you suggest)

However the challenge is in having this perform what we want it to do without breaking the existing interface.

So what Content Publisher wants is for this to return a unique URL for each asset produced without us having to tell it every single URL (as this means every time we stub it us knowing exactly which variants we'd be producing which feels like unnecessary coupling).

A weakness that is already present in this method (in my opinion) is that it expects the user to know the exact Asset Manager URL format which feels like something this stub should know.

Options I've identified for this are:

1) Modify `asset_manager_receives_an_asset` to accept a block e.g:

```
      def asset_manager_receives_an_asset(response_url = nil)
        stub_request(:post, "#{ASSET_MANAGER_ENDPOINT}/assets").to_return do
          response_url = yield if block_given?
          { body: { file_url: response_url }.to_json, status: 200 }
        end
      end
```

I wasn't a big fan of this since the interface `asset_manager_receives_an_asset(nil) { "https://asset-manager/#{SecureRandom.uuid}/100x100.jpg" }` looked a bit like it was fighting the old way of doing it.

2. Accept string or hash as the argument. This felt quite a Rails way of doing it where the method argument was type dependent which is not my favourite way of doing things but this seems to produce quite a clean interface at the expense of more complicated method

```
      def asset_manager_receives_an_asset(response_url_options = {})
        stub_request(:post, "#{ASSET_MANAGER_ENDPOINT}/assets").to_return do
          if response_url_options.is_a?(String)
            response_url = response_url_options
          else
            options = {
              id: SecureRandom.uuid,
              filename: SecureRandom.hex(8)
            }.merge(response_url_options)

            response_url = "#{ASSET_MANAGER_ENDPOINT}/media/#{options[:id]}/#{options[:filename]}"
          end
          { body: { file_url: response_url }.to_json, status: 200 }
        end
      end

```

3. Create a new method. Options for names could either reference `bulk` or `any` or maybe create unique asset. This doesn't seem ideal as it is very similar

4. Make a backwards incompatible change. I thought this would be pretty nasty but it looks like only us and specialist publisher use this stub. Though it's not immediately clear how much a different interface should change from the option suggested in 2.

What's your thoughts? I feel like I've gone down a bit of a "what's the lesser evil" path here.

benthorner: That looks good to me, except that the name `response_url` sounds more appropriate than `response_url_options` in the String case.

Another alternative might be to deprecate the original and create a new `stub_...` method, if we think this should be the common pattern.

Happy either way :-).

kevindew: @benthorner Thanks - I've updated the code.

lib/gds_api/test_helpers/asset_manager.rb
kevindew: I've made this change too.

PR #885, opened 2019-01-22
Prefer stub_ prefix for Asset Manager test helpers

benthorner:
https://trello.com/c/5JMGsuBj/581-follow-up-work-from-versioning-review

Using some of these helpers currently leads to amusing tests e.g.

   # current way
   ... some test setup ...
   asset_manager_deletes_any_asset
   ... some more test ...

   # proposed way
   ... some test setup ...
   stub_asset_manager_deletes_any_asset
   ... some more test ...

Without some kind of prefix, the behaviour of these methods is a bit
ambiguous, especially where they suggest some rather scary autonomous
action, like Asset Manager randomly deleting assets it's not keen on!

This deprecates the existing Asset Manager helpers and favours a new
stub_ prefix, similar to the WebMock and PublishingApi test helpers.

### DIFF
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -7,35 +7,35 @@ module GdsApi
         stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 200)
       end
 
-      def asset_manager_is_down
+      def stub_asset_manager_is_down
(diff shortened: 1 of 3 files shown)


benthorner:
@thomasleese true - I thought it would be presumptuous to say, but the intention is to replicate across the other test helpers if this change goes down well :-).

kevindew:
Seems good to me - be nice if they all prefixed with `stub_`

benthorner:
@thomasleese @kevindew here's the big one. So much typing...

https://github.com/alphagov/gds-api-adapters/pull/886

PR #886, opened 2019-01-22
Prefer stub_ prefix for all adapter test helpers

benthorner:
Using some of these helpers currently leads to amusing tests e.g.

   # current way
   ... some test setup ...
   email_alert_api_unsubscribes_a_subscriber
   ... some more test ...

   # proposed way
   ... some test setup ...
   stub_email_alert_api_unsubscribes_a_subscriber
   ... some more test ...

Without some kind of prefix, the behaviour of these methods is a bit
ambiguous, especially where they suggest some rather scary autonomous
action, like Email Alert API randomly deleting a subscriber by itself.

This deprecates the existing API adapter test helpers and favours a new
stub_ prefix, similar to the WebMock and PublishingApi test helpers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 57.1.0
 
+* Change all test helpers to use a stub\_ prefix (old names are now aliases)
 * Add additional Asset Manager test helper methods
 
 # 57.0.0
--- a/lib/gds_api/test_helpers/calendars.rb
+++ b/lib/gds_api/test_helpers/calendars.rb
@@ -3,17 +3,18 @@ module GdsApi
     module Calendars
       CALENDARS_ENDPOINT = Plek.current.find('calendars')
 
-      def calendars_endpoint(in_division: nil)
+
+      def stub_calendars_endpoint(in_division: nil)
         endpoint = "#{CALENDARS_ENDPOINT}/bank-holidays"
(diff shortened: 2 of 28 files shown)


benthorner:
Done - I went with 'Aliases for DEPRECATED methods', because I love shouting ;-).

benthorner:
@thomasleese I've put one in - does it make sense to you?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 57.1.0
 
+* Change all test helpers to use a stub\_ prefix (old names are now aliases)
```
thomasleese: Just a nitpick... I'm not sure we need the \ here, I would probably go for a \`stub_\`

benthorner: Erm, I think it's a reserved symbol: https://stackoverflow.com/questions/35827838/how-to-show-underscores-symbol-in-markdown

thomasleese: Interesting, have never come across that before, it looks like GitHub flavoured Markdown doesn't follow the "standard" and doesn't italicise letters within a word! Either way, I don't mind that much.

benthorner: That is interesting - I think I'll leave as it is, as it renders something awful if you have a markdown syntax highlighting 🤢 .

PR #888, opened 2019-01-23
Fix typo in test helper for Publishing API

benthorner:


lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -609,19 +609,19 @@ def stub_publishing_api_get_editions(editions, params = {})
 
       # Aliases for DEPRECATED methods
       alias_method :publishing_api_isnt_available, :stub_publishing_api_isnt_available
-      alias_method :pubishing_api_has_content, :stub_publishing_api_has_content
```
cbaines: Took me far longer than I'd like to admit to work out that this isn't how to spell "publishing"...

PR #890, opened 2019-01-24
Fix 57.2 again

kevindew:
Oh dear! I managed to make a couple of mistakes in my changes that went into https://github.com/alphagov/gds-api-adapters/pull/883

This corrects a couple of things.

- stub_asset_manager_receives_an_asset doesn't create unique filenames each call as was intended
- rename stub_asset_manager_is_down to stub_asset_manager_isnt_available for consistency

The last change is officially a breaking change, but since it's only just been introduced it seems reasonable to fix it in a bug fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Fix stub_asset_manager_receives_an_asset not returning unique filenames each call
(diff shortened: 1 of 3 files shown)


kevindew:
Thanks @cbaines

PR #891, opened 2019-01-24
Fix more stub typos for deprecated aliases

benthorner:
(Checked against all dependabot PR failures across GOV.UK.)

### DIFF
--- a/lib/gds_api/test_helpers/link_checker_api.rb
+++ b/lib/gds_api/test_helpers/link_checker_api.rb
@@ -91,12 +91,12 @@ module GdsApi
       end
 
       # Aliases for DEPRECATED methods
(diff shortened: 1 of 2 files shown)


lib/gds_api/test_helpers/link_checker_api.rb
```diff
@@ -91,12 +91,12 @@ def stub_link_checker_api_upsert_resource_monitor(app:, reference:, links:)
       end
 
       # Aliases for DEPRECATED methods
-      alias_method :link_link_checker_api_link_report_hash, :stub_link_checker_api_link_report_hash
```
kevindew: wow that'a a lot of "link" on one line 😁

PR #897, opened 2019-01-29
Add Router API stubs for getting routes

thomasleese:
This allows clients to write tests stubbing the response from Router API easily.

[Trello Card](https://trello.com/c/dskGorBB/737-rake-task-to-check-where-content-has-ended-up)

### DIFF
--- a/lib/gds_api/test_helpers/router.rb
+++ b/lib/gds_api/test_helpers/router.rb
@@ -5,6 +5,30 @@ module GdsApi
     module Router
       ROUTER_API_ENDPOINT = Plek.current.find('router-api')
 
+      def stub_router_has_route(path, route, bearer_token = ENV['ROUTER_API_BEARER_TOKEN'])
+        stub_get_route(path, bearer_token).to_return(
+          status: 200,
+          body: route.to_json,
+          headers: { "Content-Type" => "application/json" }
+        )
+      end
+
+      def stub_router_doesnt_have_route(path, bearer_token = ENV['ROUTER_API_BEARER_TOKEN'])
+        stub_get_route(path, bearer_token).to_return(status: 404)
+      end
+
+      def stub_router_has_backend_route(path, backend_id:, route_type: "exact", disabled: false)
+        stub_router_has_route(path, handler: "backend", backend_id: backend_id, disabled: disabled, route_type: route_type)
+      end
+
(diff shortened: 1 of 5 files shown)


lib/gds_api/test_helpers/router.rb
```diff
@@ -64,6 +88,11 @@ def stub_router_commit
 
     private
 
+      def stub_get_route(path)
+        stub_http_request(:get, "#{ROUTER_API_ENDPOINT}/routes")
+          .with(query: { "incoming_path" => path })
```
elliotcm: Is it worth stubbing this with the bearer token in so we know that anything using it also passes that along?

test/router_test.rb
```diff
@@ -1,9 +1,12 @@
 require 'test_helper'
 require 'gds_api/router'
+require 'gds_api/test_helpers/router'
 
 describe GdsApi::Router do
+  include GdsApi::TestHelpers::Router
+
   before do
-    @base_api_url = "http://router-api.example.com"
+    @base_api_url = Plek.current.find('router-api')
```
elliotcm: I've been on a 5 year quest to eradicate the use of `Plek.current` because it's entirely pointless.

Plek has a shortcut for `Plek.find` here:
https://github.com/alphagov/plek/blob/master/lib/plek.rb#L156

lib/gds_api/test_helpers/router.rb
thomasleese: Yep, I think that's a good idea, I've added that now along with removing `Plek.current`!

lib/gds_api/test_helpers/router.rb
```diff
@@ -5,6 +5,35 @@ module TestHelpers
     module Router
       ROUTER_API_ENDPOINT = Plek.current.find('router-api')
 
+      def stub_router_has_route(path, route, bearer_token = nil)
+        bearer_token = ENV['ROUTER_API_BEARER_TOKEN'] unless bearer_token
```
elliotcm: On the assumption that people might want to deliberately stub with a `nil` bearer token I'd suggest this:
```
def stub_router_has_route(path, route, bearer_token = ENV['ROUTER_API_BEARER_TOKEN'])
  stub_get_route # etc
```

thomasleese: Sounds reasonable, I also realised I had put that configuration in the wrong place... so that's fixed now too!

PR #902, opened 2019-02-06
Add republish method for publishing api

1pretz1:
This change adds the ability to use the republish enpointi in the
publishing api which is being introduced in the pr below. Calling this
method will send a request to pub api to republish a live document,
remove a withdrawal or send the published items data downstream.

Related to this pr: https://github.com/alphagov/publishing-api/pull/1439
Trello: https://trello.com/c/yhcKNEWr/587-allow-users-undo-withdraw

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -106,6 +106,24 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     post_json(publish_url(content_id), params)
   end
 
+  # Republish a content item
+  #
+  # The publishing-api will "republish" a live edition. This can be used to remove an unpublishing or to
+  # re-send a published edition downstream
+  #
+  # @param content_id [UUID]
+  # @param options [Hash]
+  # @option options [String] locale The language, defaults to 'en' in publishing-api.
+  #
+  # @see ...
+  def republish(content_id, options = {})
+    optional_keys = %i[locale previous_version]
+
+    params = merge_optional_keys({}, options, optional_keys)
+
+    post_json(republish_url(content_id), params)
+  end
+
   # Import content into the publishing API
   #
   # The publishing-api will delete any content which has the content
@@ -445,6 +463,11 @@ private
   def publish_url(content_id)
     validate_content_id(content_id)
     "#{endpoint}/v2/content/#{content_id}/publish"
+  end
+
+  def republish_url(content_id)
+    validate_content_id(content_id)
+    "#{endpoint}/v2/content/#{content_id}/republish"
   end
 
   def unpublish_url(content_id)
--- a/lib/gds_api/test_helpers/publishing_api_v2.rb
+++ b/lib/gds_api/test_helpers/publishing_api_v2.rb
@@ -84,6 +84,21 @@ module GdsApi
         stub_request(:post, url).with(body: body).to_return(response)
       end
 
+      # Stub a POST /v2/content/:content_id/republish request
+      #
+      # @param content_id [UUID]
+      # @param body  [String]
+      # @param response_hash [Hash]
+      def stub_publishing_api_republish(content_id, body = {}, response_hash = {})
+        url = PUBLISHING_API_V2_ENDPOINT + "/content/#{content_id}/republish"
+        response = {
+          status: 200,
+          body: '{}',
+          headers: { "Content-Type" => "application/json; charset=utf-8" }
+        }.merge(response_hash)
+        stub_request(:post, url).with(body: body).to_return(response)
+      end
+
       # Stub a POST /v2/content/:content_id/unpublish request
       #
       # @param content_id [UUID]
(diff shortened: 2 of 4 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -106,6 +106,24 @@ def publish(content_id, update_type = nil, options = {})
     post_json(publish_url(content_id), params)
   end
 
+  # Republish a content item
+  #
+  # The publishing-api will "republish" a republishable_item item, so that it will set a live edition
```
kevindew: you've got "republishable_item item" here.

I'd suggest rephrasing to: The publishing-api will "republish" a live edition. This can be used to remove an unpublishing or to re-send a published edition downstream"

lib/gds_api/publishing_api_v2.rb
```diff
@@ -106,6 +106,24 @@ def publish(content_id, update_type = nil, options = {})
...
+  # @param options [Hash]
+  # @option options [String] locale The language, defaults to 'en' in publishing-api.
+  #
+  # @see ...
+  def republish(content_id, options = {})
+    optional_keys = %i[locale previous_version]
+
+    params = merge_optional_keys({}, options, optional_keys)
+
+    post_json(republish_url(content_id), params)
```
kevindew: This looks good 👍

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -84,6 +84,21 @@ def stub_publishing_api_publish(content_id, body, response_hash = {})
...
+      # @param response_hash [Hash]
+      def stub_publishing_api_republish(content_id, body = {}, response_hash = {})
+        url = PUBLISHING_API_V2_ENDPOINT + "/content/#{content_id}/republish"
+        response = {
+          status: 200,
+          body: '{}',
+          headers: { "Content-Type" => "application/json; charset=utf-8" }
+        }.merge(response_hash)
+        stub_request(:post, url).with(body: body).to_return(response)
+      end
```
kevindew: This looks good too. I think we'd also want one of stub_any_publishing_api_republish for consistency.

There's also the `assert*` methods too but I guess we can just add them if someone wants them as they seem like something that's not too likely to be needed.

test/publishing_api_v2_test.rb
```diff
@@ -511,6 +511,152 @@ def content_item_for_content_id(content_id, attrs = {})
     end
   end
 
+  describe "#republish" do
+    describe "if the republish command succeeds" do
+      before do
+        publishing_api
+          .given("a unpublished document exists with content_id: #{@content_id}")
```
kevindew: These given's all need to correlate with set-ups in this Publishing API file: https://github.com/alphagov/publishing-api/blob/master/spec/service_consumers/pact_helper.rb

So this one here "a unpublished document exists with content_id: #{@content_id}" doesn't exist whereas we can use: "an unpublished content item exists with content_id: #{@content_id}" that references: https://github.com/alphagov/publishing-api/blob/master/spec/service_consumers/pact_helper.rb#L137

test/publishing_api_v2_test.rb
```diff
@@ -511,6 +511,152 @@ def content_item_for_content_id(content_id, attrs = {})
...
+        end
+
+        assert_equal 404, error.code
+      end
+    end
+
+    describe "if the update information contains a locale" do
+      before do
+        publishing_api
+          .given("a draft content item exists with content_id: #{@content_id} and locale: fr")
```
kevindew: This one doesn't seem right as we can't do this on a draft. If we want to do this with a pact test we'll need to add another set-up clause. Though I don't really think this adds a huge amount of value so I'd suggest we just remove this describe since setting up for it requires us to do additional work.

test/publishing_api_v2_test.rb
```diff
@@ -511,6 +511,152 @@ def content_item_for_content_id(content_id, attrs = {})
...
+        response = @api_client.republish(@content_id, locale: "fr")
+        assert_equal 200, response.code
+      end
+    end
+
+    describe "optimistic locking" do
+      describe "if the content item has not changed since it was requested" do
+        before do
+          publishing_api
+            .given("the content item #{@content_id} is at version 3")
```
kevindew: This one also references a draft but we can switch this to be https://github.com/alphagov/publishing-api/blob/master/spec/service_consumers/pact_helper.rb#L447

lib/gds_api/publishing_api_v2.rb
1pretz1: sure

lib/gds_api/test_helpers/publishing_api_v2.rb
1pretz1: Yep good point

test/publishing_api_v2_test.rb
1pretz1: Ah okay, i was confuddled by this bit and thought they were just namespaces

test/publishing_api_v2_test.rb
1pretz1: sure

test/publishing_api_v2_test.rb
1pretz1: 👍

1pretz1:
@kevindew Okay thanks for that!

PR #907, opened 2019-02-27
Rename subscribable to subscriber list

thomasleese:
At one point we were planning to rename all instances of subscriber lists with subscribables however these never happened and it's confusing to have the same thing named differently. Instead we've decided to stick with subscriber list.

This is a breaking change for users of the Email Alert API adapters, clients would be expected to update the code. However, Email Alert API itself will still accept the old style arguments.

Depends on https://github.com/alphagov/email-alert-api/pull/800.

[Trello Card](https://trello.com/c/5qd2mgF0/76-subscriberlist-and-subscribable-are-used-for-the-same-thing-in-email-alert-api)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+* Rename references of subscribable to subscriber_list in Email Alert API. Note
+  this is a breaking change for users of the Email Alert API adapters, clients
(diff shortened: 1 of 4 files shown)


cbaines:
It would be good to put some of the information in this Pull Request in to the first commit.

PR #911, opened 2019-03-15
Fix tests

benthorner:


test/test_helper.rb
```diff
@@ -1,3 +1,5 @@
+$VERBOSE = nil
```
kevindew: I think this is explicitly turned on in the rakefile https://github.com/alphagov/gds-api-adapters/blob/master/Rakefile#L14 - so we can probably remove it there rather than need to change the ruby runtime option.

benthorner: Ah, good spot, but unfortunately not. Ruby warnings aren't turned on by default, so my assumption is that one of the gems is doing it.

kevindew: Ah darn.

~~ has a play ~~

We can get the behaviour you're aiming by replacing `t.verbose = true` with `t.warning = false`

At the least we should remove the verbose off there if we do it here so we don't have one place turning verbosity up and one down.

benthorner: Yay, thank you - it's even better like that :-).

PR #912, opened 2019-03-15
Add pub api unreserve path

benthorner:
https://trello.com/c/6s9Bhzyj/632-delete-publishing-api-path-reservations-when-we-discard-drafts

This complements the new Publishing API endpoint to unreserve a path (https://github.com/alphagov/publishing-api/pull/1468).

### DIFF
--- a/lib/gds_api/json_client.rb
+++ b/lib/gds_api/json_client.rb
@@ -64,8 +64,8 @@ module GdsApi
       do_json_request(:patch, url, params, additional_headers)
     end
 
-    def delete_json(url, additional_headers = {})
-      do_json_request(:delete, url, nil, additional_headers)
+    def delete_json(url, params = {}, additional_headers = {})
+      do_json_request(:delete, url, params, additional_headers)
     end
 
     def delete_json_with_params!(url, params, additional_headers = {})
--- a/lib/gds_api/json_client.rb
+++ b/lib/gds_api/json_client.rb
@@ -68,18 +68,6 @@ module GdsApi
       do_json_request(:delete, url, params, additional_headers)
     end
 
-    def delete_json_with_params!(url, params, additional_headers = {})
-      warn <<-doc
-        DEPRECATION NOTICE: Delete requests should not include parameters.
-
-        Do not use this method as the ability to do this will be removed.
-
-        Called from: #{caller[2]}
-      doc
-
-      do_json_request(:delete, url, params, additional_headers)
-    end
-
     def post_multipart(url, params)
       r = do_raw_request(:post, url, params.merge(multipart: true))
       Response.new(r)
--- a/lib/gds_api/rummager.rb
+++ b/lib/gds_api/rummager.rb
@@ -19,6 +19,7 @@ module GdsApi
       def delete_document(type, id)
         delete_json(
           "#{documents_url}/#{id}",
+          nil,
           _type: type,
         )
       end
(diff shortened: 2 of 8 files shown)


lib/gds_api/json_client.rb
```diff
@@ -64,8 +64,8 @@ def patch_json(url, params, additional_headers = {})
       do_json_request(:patch, url, params, additional_headers)
     end
 
-    def delete_json(url, additional_headers = {})
-      do_json_request(:delete, url, nil, additional_headers)
+    def delete_json(url, params = {}, additional_headers = {})
+      do_json_request(:delete, url, params, additional_headers)
```
kevindew: Did you see the method below this one? 

I'm assuming the deprecation notice is superseded by a later RFC where it says it's ok about bodies in delete. Might need a bit of cleanup here.

test/publishing_api_test.rb
```diff
@@ -11,6 +11,91 @@
...
+          headers: {
+            "Content-Type" => "application/json; charset=utf-8"
+          },
+        )
+
+      assert_raises GdsApi::HTTPUnprocessableEntity do
+        @api_client.unreserve_path(base_path, publishing_app)
+      end
+    end
+  end
```
kevindew: These look good 👍

lib/gds_api/json_client.rb
benthorner: Ah, I feel so blind, but also naughty 😆 .

benthorner: Updated.

benthorner:
@kevindew this should be ready to review again now :-).

lib/gds_api/rummager.rb
```diff
@@ -19,6 +19,7 @@ def add_document(type, id, document)
       def delete_document(type, id)
         delete_json(
           "#{documents_url}/#{id}",
+          nil,
```
1pretz1: Why add `nil` here as an argument when this would default to an empty hash?

1pretz1: Ah okay nm, it's so`_type` is an additional_header not a param, ignore me!

lib/gds_api/json_client.rb
```diff
@@ -68,18 +68,6 @@ def delete_json(url, params = {}, additional_headers = {})
       do_json_request(:delete, url, params, additional_headers)
     end
 
-    def delete_json_with_params!(url, params, additional_headers = {})
```
1pretz1: If you're removing this method then this delegation needs to be cleaned up - https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/base.rb#L26

lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -11,15 +11,15 @@ module PublishingApi
 
       PUBLISHING_API_ENDPOINT = Plek.current.find('publishing-api')
 
-      def stub_publishing_api_unreserve_path(base_path, publishing_app)
+      def stub_publishing_api_unreserve_path(base_path, publishing_app = /.*/)
```
1pretz1: Hmm, I wonder for the amount of times we are actually going to use these stubs if it would be simpler and easier to understand to just require a `publishing_app` instead of using a regex?

benthorner: Originally I didn't default it, but then I found that it bloats the test code in Content Publisher, where most of the time we don't care about the publishing_app, and it was easier to do this than adding another 'any'-type stub.

lib/gds_api/json_client.rb
benthorner: Ah, good spot - removed too :-).

lib/gds_api/test_helpers/publishing_api.rb
1pretz1: Okay sure!

PR #913, opened 2019-03-19
Add some documentation around publishing intents

1pretz1:
This adds some brief guidance around using the intents endpoint with
links to more detailed guidance in the publishing-api.

Depends on:
https://github.com/alphagov/publishing-api/pull/1470

Trello:
https://github.com/alphagov/publishing-api/pull/1470

### DIFF
--- a/lib/gds_api/publishing_api.rb
+++ b/lib/gds_api/publishing_api.rb
@@ -2,10 +2,31 @@ require_relative 'base'
 require_relative 'exceptions'
 
 class GdsApi::PublishingApi < GdsApi::Base
+  # Create a publishing intent for a base_path.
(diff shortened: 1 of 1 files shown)


lib/gds_api/publishing_api.rb
```diff
@@ -2,10 +2,31 @@
 require_relative 'exceptions'
 
 class GdsApi::PublishingApi < GdsApi::Base
+  # Put a publishing intent for a base_path.
```
benthorner: I think 'Create' would make more sense here.

1pretz1: Sure!

PR #916, opened 2019-04-01
Expect apps to use public bank holidays API

thomasleese:
Instead of talking directly to Calendars, we should make apps talk via the public API. This is similar to https://github.com/alphagov/gds-api-adapters/pull/857.

When migrating the frontend apps to AWS, we noticed that Mainstream Publisher talks directly to Calendars rather than going through the public API which meant we had to set up a specific route for this app.

[Trello Card](https://trello.com/c/qDW6cUad/129-mainstream-publisher-talks-to-calendars)

### DIFF
--- a/lib/gds_api/test_helpers/calendars.rb
+++ b/lib/gds_api/test_helpers/calendars.rb
@@ -1,11 +1,8 @@
 module GdsApi
   module TestHelpers
     module Calendars
-      CALENDARS_ENDPOINT = Plek.current.find('calendars')
-
-
       def stub_calendars_endpoint(in_division: nil)
-        endpoint = "#{CALENDARS_ENDPOINT}/bank-holidays"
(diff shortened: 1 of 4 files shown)


lib/gds_api/test_helpers/calendars.rb
```diff
@@ -1,8 +1,7 @@
 module GdsApi
   module TestHelpers
     module Calendars
-      CALENDARS_ENDPOINT = Plek.current.find('calendars')
-
+      CALENDARS_ENDPOINT = Plek.new.website_root
```
bevanloon: Is it worth renaming this constant since it no longer represents the actual endpoint? or maybe just move `Plek.new.website_root` into `stub_calendars_endpoint` instead?

thomasleese: Sounds good, I've pushed up an amended commit to put it directly into `stub_calendars_endpoint`.

PR #918, opened 2019-04-10
[Do not merge] Add notify endpoint

steventux:
https://trello.com/c/0wiO10tC/16-email-trigger-mechanism-for-new-content-tagger-process

See https://github.com/alphagov/publishing-api/pull/1482

### DIFF
--- a/lib/gds_api/publishing_api_v2.rb
+++ b/lib/gds_api/publishing_api_v2.rb
@@ -442,6 +442,20 @@ class GdsApi::PublishingApiV2 < GdsApi::Base
     put_json(url, payload)
   end
 
+  # Generates a notification via the publishing API
+  #
+  # Returns success or failure only.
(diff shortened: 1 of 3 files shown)


lib/gds_api/publishing_api_v2.rb
```diff
@@ -442,6 +442,20 @@ def put_path(base_path, payload)
     put_json(url, payload)
   end
 
+  # Generates a notification via the publishing API
+  #
+  # Returns success or failure only.
+  #
+  # @param payload [Hash]
+  # @option payload [Hash] publishing_app The publishing application, like `content-tagger`
```
leenagupte: `publishing_app` is optional, so can it be labelled as such?

leenagupte:
We're no longer creating a new notify endpoint.

PR #929, opened 2019-05-16
Exponential backoff

bilbof:
This enables us to retry requests that return a 5XX on a GET request. It will exponentially increase the wait time between making repeated requests.

This will retry search requests a few times (with exponential backoff) if the search query responds with a 500.

The motivation for this is that we see a fair number of GdsApi::HTTPInternalServerError errors in our logs. I've seen these are happening around deploys, but they shouldn't ever happen, as they return the error to the user, telling them to repeat the request.

I'd be interested in people's thoughts on this in general and on the implementation. If you can recommend a gem we can use for this I'd be keen (from what I can see the rest client we use doesn't have a component for retries). 

Trello: https://trello.com/c/cj9a4QSa/736-retry-failed-search-queries-if-we-get-a-server-error-from-search-api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Add get_json_with_retries to json_client to permit retrying requests.
+
 # 59.2.1
 
 * Warn when `GdsApi::Rummager` is initialised and `GdsApi::TestHelpersRummager`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Retry search requests three times if we receive a server error from search-api.
 * Add get_json_with_retries to json_client to permit retrying requests.
 
 # 59.2.1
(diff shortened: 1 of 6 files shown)


bilbof:
Thanks, Simon. By deploys, I meant they occur around finder-frontend deploys rather than search-api. 

These errors have been raised a lot this week (https://sentry.io/organizations/govuk/issues/419248929/?environment=production) but they don't seem to correlate with search-api deploys (rather, finder-frontend deploys): https://grafana.blue.production.govuk.digital/dashboard/file/search-api.json?refresh=5s&orgId=1&from=now-7d&to=now.

I'll have a dig and see if I can find out what is causing them.

bilbof:
I did some digging and it looks like there isn't a strong enough argument to support retries on search requests, since we're seeing them mainly on malformed requests and timeouts. I'll close this for now.

PR #931, opened 2019-05-17
Bump to 59.3.0

ChrisBAshton:
Functionality added in https://github.com/alphagov/gds-api-adapters/pull/928.


lib/gds_api/version.rb
```diff
@@ -1,3 +1,3 @@
 module GdsApi
-  VERSION = '59.2.1'.freeze
+  VERSION = '59.2.2'.freeze
```
leenagupte: This should be 59.3.0

ChrisBAshton: 🤦‍♂

PR #932, opened 2019-05-21
Adds 'stub_email_alert_api_has_subscriptions' test method

ChrisBAshton:
Trello card: https://trello.com/c/VlnvU2Dl
Depended on by: https://github.com/alphagov/email-alert-frontend/pull/474 (please read that for more context)

## What
- Adds a `stub_email_alert_api_has_subscriptions` method, which takes an array of subscriptions to stub, so that we can test [like so](https://github.com/alphagov/email-alert-frontend/pull/474/files#diff-20c575b223c5b129c1750d0f400abda8R64).
- Ensures that `stub_email_alert_api_has_subscription` stubs the '/latest' endpoint (this is needed if we're only stubbing one subscription response).
- Refactors the shared code between these methods into a new `fill_gaps_with_default_values` method.
- Writes tests for `stub_email_alert_api_has_subscriptions`.

## Why
We added a `get_latest_matching_subscription` method in https://github.com/alphagov/gds-api-adapters/pull/928 (v59.3.0), which takes a subscription ID and returns the latest matching subscription response (e.g. a user changing frequency from Daily to Weekly, if we query '/latest' on the Daily subscription, it will return the Weekly one as that is the most recent).

Consequently, the return value of '/latest' depends upon whether or not a newer subscription has been stubbed. It was very difficult / hacky / impossible to test the '/latest' method behaviour with the `get_latest_matching_subscription` method alone, so I've added a `stub_email_alert_api_has_subscriptions` which takes an array of subscriptions and stubs them in the order they are added.

I.e. if you pass `stub_email_alert_api_has_subscriptions([subscription1, subscription2])` and then query '/latest' on `subscription1`, the response for `subscription2` should be returned.

## Note
It was quite difficult to write 'clean' code here because the `stub_email_alert_api_has_subscription` method takes three parameters (`id`, `frequency`, `'hash of options'`), which means you cannot pass `{ id: 1, frequency: 'weekly', ended: false }` etc directly but have to instead remove the options that are not explicitly allowed in that third parameter.

Consequently, I was unable to simply call `stub_email_alert_api_has_subscription` from within `stub_email_alert_api_has_subscriptions` without hacking the options parameter, and it was cleaner to simply implement the `stub_request` method in place. I also have had to return `[id, parameters]`, rather than simply `parameters` (containing `:id`) as again, this otherwise leads to a `ArgumentError: unknown keyword: id`.

In the long term, it would be great to update `stub_email_alert_api_has_subscription` to take a single hash rather than 3 params, but this will involve a change to the email-alert-frontend tests which rely upon the method. Doing this should delete several lines of code and improve code readability and consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Adds `stub_email_alert_api_has_subscriptions` test helper method.
+* Ensures that `stub_email_alert_api_has_subscription` also stubs the `/latest` endpoint.
+
 # 59.3.0
 
 * Add `get_latest_matching_subscription` method to `GdsApi::EmailAlertApi`.
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -68,6 +68,55 @@ module GdsApi
           .to_return(status: 200, body: response)
       end
 
+      # Stubs the API responses as if each subscription happened in the order they are passed.
+      # Useful if you need to query the '/latest' endpoint of a subscription.
+      # Takes an array of hashes.
+      #
+      # @example
+      #  stub_email_alert_api_has_subscriptions([
+      #    {
+      #      id: 'id-of-my-subscriber-list',
+      #      frequency: 'weekly',
+      #      ended: true,
+      #    },
+      #    {
+      #      id: 'id-of-my-subscriber-list',
+      #      frequency: 'daily',
+      #    },
+      #  ])
+      #
+      # @param subscriptions [Array]
+      def stub_email_alert_api_has_subscriptions(subscriptions)
+        subscriptions.map! { |subscription| apply_subscription_defaults(subscription) }
+        subscriptions.each do |id, params|
+          latest_id, latest_params = get_latest_matching(params, subscriptions)
+          stub_request(:get, "#{EMAIL_ALERT_API_ENDPOINT}/subscriptions/#{id}")
+            .to_return(status: 200, body: get_subscription_response(id, params).to_json)
+          stub_request(:get, "#{EMAIL_ALERT_API_ENDPOINT}/subscriptions/#{id}/latest")
+            .to_return(status: 200, body: get_subscription_response(latest_id, latest_params).to_json)
+        end
+      end
+
+      def apply_subscription_defaults(subscription)
+        parameters = {
+          title: "Some title",
+          subscriber_id: 1,
+          subscriber_list_id: 1000,
+          ended: true,
+        }.merge(subscription)
+        # Strip out ID as subsequent call to `get_subscription_response` throws `ArgumentError`
+        id = parameters.delete(:id)
+        [id, parameters]
+      end
+
+      def get_latest_matching(params, subscriptions)
+        matching = subscriptions.select do |_current_id, current_params|
+          params[:subscriber_id] == current_params[:subscriber_id] &&
+            params[:subscriber_list_id] == current_params[:subscriber_list_id]
+        end
+        matching.last
+      end
+
       def stub_email_alert_api_has_subscriber_list(attributes)
         stub_request(:get, build_subscriber_lists_url(attributes))
           .to_return(
@@ -228,6 +277,7 @@ module GdsApi
       alias_method :email_alert_api_has_subscriber_subscriptions, :stub_email_alert_api_has_subscriber_subscriptions
       alias_method :email_alert_api_does_not_have_subscriber_subscriptions, :stub_email_alert_api_does_not_have_subscriber_subscriptions
       alias_method :email_alert_api_has_subscription, :stub_email_alert_api_has_subscription
+      alias_method :email_alert_api_has_subscriptions, :stub_email_alert_api_has_subscriptions
       alias_method :email_alert_api_has_subscriber_list, :stub_email_alert_api_has_subscriber_list
       alias_method :email_alert_api_does_not_have_subscriber_list, :stub_email_alert_api_does_not_have_subscriber_list
       alias_method :email_alert_api_creates_subscriber_list, :stub_email_alert_api_creates_subscriber_list
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -53,18 +53,19 @@ module GdsApi
         subscriber_list_id: 1000,
         ended: false
       )
+        response = get_subscription_response(
+          id,
+          frequency: frequency,
+          title: title,
+          subscriber_id: subscriber_id,
+          subscriber_list_id: subscriber_list_id,
+          ended: ended,
+        ).to_json
+
         stub_request(:get, "#{EMAIL_ALERT_API_ENDPOINT}/subscriptions/#{id}")
-          .to_return(
-            status: 200,
-            body: get_subscription_response(
-              id,
-              frequency: frequency,
-              title: title,
-              subscriber_id: subscriber_id,
-              subscriber_list_id: subscriber_list_id,
-              ended: ended,
-            ).to_json,
-          )
+          .to_return(status: 200, body: response)
+        stub_request(:get, "#{EMAIL_ALERT_API_ENDPOINT}/subscriptions/#{id}/latest")
+          .to_return(status: 200, body: response)
       end
 
       def stub_email_alert_api_has_subscriber_list(attributes)
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -271,7 +271,7 @@ module GdsApi
             "id" => id,
             "frequency" => frequency,
             "source" => "user_signed_up",
-            "ended_at" => ended ? Time.now.rfc3339 : nil,
+            "ended_at" => ended ? Time.now.to_datetime.rfc3339 : nil,
             "ended_reason" => ended ? "unsubscribed" : nil,
             "subscriber" => {
               "id" => subscriber_id,
--- a/test/email_alert_api_test.rb
+++ b/test/email_alert_api_test.rb
@@ -85,7 +85,85 @@ describe GdsApi::EmailAlertApi do
           .to_hash
           .fetch("subscription")
 
-         assert_equal(1, subscription_attrs.fetch("id"))
+        assert_equal(1, subscription_attrs.fetch("id"))
+      end
+    end
+
+    describe "user unsubscribed then resubscribed at the same frequency" do
+      before do
+        stub_email_alert_api_has_subscriptions([
+          {
+            id: 1,
+            frequency: "weekly",
+            ended: true,
+          },
+          {
+            id: 2000,
+            frequency: "weekly",
+          },
+        ])
+      end
+
+      it "returns the latest subscription attributes on /latest" do
+        subscription_attrs = api_client.get_latest_matching_subscription(1)
+          .to_hash
+          .fetch("subscription")
+
+        assert_equal(2000, subscription_attrs.fetch("id"))
+      end
+    end
+
+    describe "user changed their frequency then unsubscribed altogether" do
+      before do
+        stub_email_alert_api_has_subscriptions([
+          {
+            id: 1,
+            frequency: "weekly",
+            ended: true
+          },
+          {
+            id: 2000,
+            frequency: "daily",
+            ended: true,
+          }
+        ])
+      end
+
+      it "returns the latest subscription attributes on /latest" do
+        subscription_attrs = api_client.get_latest_matching_subscription(1)
+          .to_hash
+          .fetch("subscription")
+
+        assert_equal(2000, subscription_attrs.fetch("id"))
+      end
+    end
+
+    describe "user is subscribed to multiple lists" do
+      before do
+        stub_email_alert_api_has_subscriptions([
+          {
+            id: 1,
+            frequency: "weekly",
+            subscriber_list_id: 123,
+          },
+          {
+            id: 2,
+            frequency: "weekly",
+            subscriber_list_id: 456,
+          }
+        ])
+      end
+
+      it "returns the correct subscriber list on /latest" do
+        first_subscription_attrs = api_client.get_latest_matching_subscription(1)
+          .to_hash
+          .fetch("subscription")
+        second_subscription_attrs = api_client.get_latest_matching_subscription(2)
+          .to_hash
+          .fetch("subscription")
+
+        assert_equal(1, first_subscription_attrs.fetch("id"))
+        assert_equal(2, second_subscription_attrs.fetch("id"))
       end
     end
   end
--- a/test/email_alert_api_test.rb
+++ b/test/email_alert_api_test.rb
@@ -79,6 +79,14 @@ describe GdsApi::EmailAlertApi do
 
         assert_equal("weekly", subscription_attrs.fetch("frequency"))
       end
+
+      it "returns the subscription attributes on /latest" do
+        subscription_attrs = api_client.get_latest_matching_subscription(1)
+          .to_hash
+          .fetch("subscription")
+
+         assert_equal(1, subscription_attrs.fetch("id"))
+      end
     end
   end
 


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 59.3.1
```
leenagupte: Updates to the version number should happen in a separate PR. For now, the changes should be listed as `# Unreleased`

ChrisBAshton: Oops - good to know! Will do that.

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -45,26 +45,25 @@ def stub_email_alert_api_does_not_have_subscriber_subscriptions(id)
...
 
-      def stub_email_alert_api_has_subscription(
-        id,
-        frequency,
-        title: "Some title",
-        subscriber_id: 1,
-        subscriber_list_id: 1000,
-        ended: false
-      )
+      def stub_email_alert_api_has_subscription(id, frequency, optional_overrides = {})
```
leenagupte: @ChrisBAshton The method already allows you to override the values for title, subscriber_id, subscriber_list_id, and ended. Leaving them as named parameters makes it easier to find out what these values are.
It also means that you don't need a separate method to add these values.

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -227,6 +226,7 @@ def stub_email_alert_api_does_not_have_subscriber_list_by_slug(slug:)
       alias_method :email_alert_api_has_subscriber_subscriptions, :stub_email_alert_api_has_subscriber_subscriptions
       alias_method :email_alert_api_does_not_have_subscriber_subscriptions, :stub_email_alert_api_does_not_have_subscriber_subscriptions
       alias_method :email_alert_api_has_subscription, :stub_email_alert_api_has_subscription
+      alias_method :email_alert_api_has_subscriptions, :stub_email_alert_api_has_subscriptions
```
leenagupte: Was this missed in the previous PR?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -270,7 +270,7 @@ def get_subscription_response(
             "id" => id,
             "frequency" => frequency,
             "source" => "user_signed_up",
-            "ended_at" => ended ? Time.now.rfc3339 : nil,
+            "ended_at" => ended ? Time.now.to_datetime.rfc3339 : nil,
```
leenagupte: Why has this changed? Can you move it to a separate commit and add an explanation?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -45,26 +45,25 @@ def stub_email_alert_api_does_not_have_subscriber_subscriptions(id)
...
-              id,
-              frequency: frequency,
-              title: title,
-              subscriber_id: subscriber_id,
-              subscriber_list_id: subscriber_list_id,
-              ended: ended,
-            ).to_json,
-          )
+          .to_return(status: 200, body: response)
+        stub_request(:get, "#{EMAIL_ALERT_API_ENDPOINT}/subscriptions/#{id}/latest")
```
leenagupte: Should this be in a separate method? i.e. `stub_email_alert_api_has_latest_subscription`?

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: No, the previous PR added a new API method that can be used by email-alert-frontend.

This PR adds a new test helper, which needs exposing here.

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: The list of keys was in two places (the `stub_email_alert_api_has_subscription` method parameters, and the `get_subscription_response` call below it), and in creating my `stub_email_alert_api_has_subscriptions` method I risked duplicating it a third time and the default values a second time (in my `get_subscription_response` call).

Felt like a good idea to extract that out so that we only have one place to update if we want to change a default value or add a new override. If we'd prefer the duplication I can do that.

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: That crossed my mind, but _every_ subscription should have a valid response if you call `/latest`. So it would mean otherwise writing this in our tests:

```ruby
stub_email_alert_api_has_subscription(1, 'weekly', :ended)
stub_email_alert_api_has_latest_subscription(1, 'weekly', :ended)
stub_email_alert_api_has_subscription(2, 'weekly')
stub_email_alert_api_has_latest_subscription(2, 'weekly')
```

I think this makes things less readable - it can be misinterpreted as "this subscription should be the latest one", which is not what we want. What we want, if anything, is "this subscription should be queryable against the '/latest' endpoint". But if you've stubbed the email API for a subscription, that should just be done for you IMO.

What do you think?

leenagupte: Can you add this explanation to the commit note?

lib/gds_api/test_helpers/email_alert_api.rb
leenagupte: At this point (the first commit) you only have one method (`stub_email_alert_api_has_subscription`) so extracting out the default values it's necessary.

ChrisBAshton: Ah, agreed. Should I split it out so that it's repeated initially, then refactored into its own method in a subsequent commit?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -57,6 +57,21 @@ def stub_email_alert_api_has_subscription(id, frequency, optional_overrides = {}
           .to_return(status: 200, body: response)
       end
 
+      # Stubs the API responses as if each subscription happened in the order they are passed
+      def stub_email_alert_api_has_subscriptions(*subscriptions)
+        subscriptions.map! do |id, frequency, optional_overrides|
```
leenagupte: Rather that specifying the parameters in the block definition, you can just use 
`subscriptions.map! do |subscription|` and use `subscription[:id]`  to populate the id.

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -57,6 +57,21 @@ def stub_email_alert_api_has_subscription(id, frequency, optional_overrides = {}
           .to_return(status: 200, body: response)
       end
 
+      # Stubs the API responses as if each subscription happened in the order they are passed
+      def stub_email_alert_api_has_subscriptions(*subscriptions)
```
leenagupte: What is the benefit of using the a splat here?

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: I don't think you can? `subscription` is a tuple of two strings and a hash:

```ruby
[
  'abc123',
  'daily',
  {
    ended: false,
    subscription_reason: 'etc',
  }
]
```

leenagupte: How would you get the `id` for `subscriptions[0]`? That's what you'd use in the loop.

lib/gds_api/test_helpers/email_alert_api.rb
leenagupte: I think this is in the wrong commit then, as `stub_email_alert_api_has_subscriptions` doesn't yet.

test/email_alert_api_test.rb
```diff
@@ -79,6 +79,48 @@
...
+    describe "user unsubscribed then resubscribed at the same frequency" do
+      before do
+        stub_email_alert_api_has_subscriptions(
+          [1, "weekly", ended: true],
+          [2, "weekly"]
+        )
+      end
+
+      it "returns the latest subscription attributes on /latest" do
+        subscription_attrs = api_client.get_latest_matching_subscription(1) # notice the 1
```
leenagupte: Rather use a comment, could the ids be changed to something more obvious? i.e. a larger number?

test/email_alert_api_test.rb
```diff
@@ -79,6 +79,48 @@
...
+          .fetch("subscription")
+
+        assert_equal(1, subscription_attrs.fetch("id"))
+      end
+    end
+
+    describe "user unsubscribed then resubscribed at the same frequency" do
+      before do
+        stub_email_alert_api_has_subscriptions(
+          [1, "weekly", ended: true],
```
leenagupte: What happens if the user has more than one active subscription?

test/email_alert_api_test.rb
```diff
@@ -79,6 +79,48 @@
 
         assert_equal("weekly", subscription_attrs.fetch("frequency"))
       end
+
+      it "returns the subscription attributes on /latest" do
```
leenagupte: This test looks like a duplicate of line 100 `it "returns the latest subscription attributes on /latest"` are they both needed?

test/email_alert_api_test.rb
leenagupte: Also what happens if a user's subscriptions look like: 

```
[1, "weekly"],
 [2, "daily", ended: true]
```

lib/gds_api/test_helpers/email_alert_api.rb
leenagupte: I think you can justify expanding the comment on the method to clarify what a `subscriptions` object looks like, and how the method should be called. 
There are some examples in the publishing-api test helpers: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb

lib/gds_api/test_helpers/email_alert_api.rb
leenagupte: I think it's ok to keep the duplication. It's likely that the two test helpers will be used independently of each other, so I think it'll be more helpful to the person using them to be able to clearly see the defaults.

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: If I do `subscriptions.map! do |subscription|` then inside the loop it will either be a):

```ruby
id, frequency, optional_overrides = subscription
```

or b):

```ruby
id = subscription[0]
frequency = subscription[1]
optional_overrides = subscription[2]
```

Which would be best?

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Good point! None :)

ChrisBAshton: Because we pass subscriptions as a 0..* list, not an array

```ruby
stub_email_alert_api_has_subscriptions(
  [1, 'weekly', :ended]
  [2, 'daily']
)
```

Happy to change the API to take an array instead, i.e.

```ruby
stub_email_alert_api_has_subscriptions([
  [1, 'weekly', :ended]
  [2, 'daily']
])
```

lib/gds_api/test_helpers/email_alert_api.rb
leenagupte: Could `subscriptions` be changed to be an array of hashes?
e.g.
```
[
  {
    id: 'abc123', 
    frequency: 'daily',
    ended: false,
    subscription_reason: 'etc',
  }
]
```
It might make it easier to call from a test.

test/email_alert_api_test.rb
leenagupte: That's not true. At the moment you can create a daily and weekly etc, subscription to the same list.

ChrisBAshton: I thought one automatically unsubscribed the other?

leenagupte: I thought @brucebolt  created two yesterday. It's worth checking, and adding a test if required and a commit note if not.

ChrisBAshton: Bruce & I just checked this; I signed up for Daily and then Immediately on integration (both from the `/email/subscriptions/new?topic_id=find-eu-exit-guidance-for-your-business-appear-in-find-eu-exit-guidance-business-finder` endpoint).

In the database, the Daily one was `ended` with reason `frequency_changed`, and the Immediately one has a 'source' of `frequency_changed`, so it doesn't look like it's possible to have multiple active subscriptions.

But we did look for a test and couldn't find one that captures this, so I'll add one in a PR 👍

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Done.

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Whoops!

test/email_alert_api_test.rb
ChrisBAshton: Oh actually it is being tested here: https://github.com/alphagov/email-alert-api/blob/8c68924f632760c22154757e54b903841fb7a86d/spec/integration/create_subscription_spec.rb#L35-L58

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -68,6 +68,54 @@ def stub_email_alert_api_has_subscription(
...
+      #      ended: true,
+      #    },
+      #    {
+      #      id: 'id-of-my-subscriber-list',
+      #      frequency: 'daily',
+      #    },
+      #  ])
+      #
+      # @param subscriptions [Array]
+      def stub_email_alert_api_has_subscriptions(subscriptions)
```
leenagupte: As the last subscription in the list should be the "latest" I wonder whether there's any point in passing `ended` into the hash, and instead automatically set it to true on all but the last subscription in the list?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -68,6 +68,54 @@ def stub_email_alert_api_has_subscription(
...
+        subscriptions.map! { |subscription| apply_subscription_defaults(subscription) }
+        subscriptions.each do |id, params|
```
leenagupte: Are two loops needed here?

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Still worth passing it to make the tests clear, I think - but can definitely override all but the last subscription to be `ended: true`. 👍

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Yes, as we pass the modified `subscriptions` as a parameter on the next line. We've got a nice API for calling the method, and we need to transform the data into something we can pass around.

Having two loops as we do here is a complexity of `2n`; the alternative is a loop within a loop (`n²`).

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: I now set `ended: true` by default in `apply_subscription_defaults`.

lib/gds_api/test_helpers/email_alert_api.rb
brucebolt: Would it be possible to have something like this to avoid two loops?

```
subscriptions.each do |subscription|
  subscription = apply_subscription_defaults(subscription)
  ...
```

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -68,6 +68,54 @@ def stub_email_alert_api_has_subscription(
...
+      end
+
+      def apply_subscription_defaults(subscription)
+        parameters = {
+          title: "Some title",
+          subscriber_id: 1,
+          subscriber_list_id: 1000,
+          ended: true,
+        }.merge(subscription)
+        id = parameters.delete(:id) # Otherwise `ArgumentError: unknown keyword: id`
```
brucebolt: I'm not clear on why the `id` value has to be deleted from `parameters`.  Where does the `ArgumentError` occur if this isn't deleted?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -68,6 +68,54 @@ def stub_email_alert_api_has_subscription(
...
+        [id, parameters]
+      end
+
+      def get_latest_matching(params, subscriptions)
+        matching = subscriptions.select do |_current_id, current_params|
+          params[:subscriber_id] == current_params[:subscriber_id] &&
+            params[:subscriber_list_id] == current_params[:subscriber_list_id]
+        end
+        matching.last
```
brucebolt: Do you need to specify the ordering of results for the `select` query so `.last` is always the final result?  I can't see a `default_scope` for the Subscription model, so the return of results may not always be in a consistent order.

lib/gds_api/test_helpers/email_alert_api.rb
leenagupte: Is the description being merged in from the subscription? Is it better to be explicit about what can be added to the parameters hash? i.e. only adding `frequency` and `ended` if `ended` exists?

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Not reeeeeally, because we'd be using `subscriptions` before all of the subscriptions have been transformed, like:

```ruby
subscriptions.each do |subscription|
  subscription = apply_subscription_defaults(subscription)
  foo(subscriptions) # oops - we now need to apply defaults to all of these somehow
```

Which is why I've done it outside the loop:

```ruby
subscriptions = transform_the_subs(subscriptions)
subscriptions.each do |subscription|
  subscription = apply_subscription_defaults(subscription)
  foo(subscriptions) # now we can call this safely :)
```

If it's code _efficiency_ we're worried about, I think it's better as-is, otherwise within `foo(subscriptions)` we have to call `apply_subscription_defaults` on each subscription (a loop within a loop).

If it's just readability we're worried about, I can look at calling `apply_subscription_defaults` in the `get_latest_matching` method...?

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: The id hack is in order to support the nice API you suggested here: https://github.com/alphagov/gds-api-adapters/pull/932#discussion_r286040228

Unfortunately I can't pass the parameters 'blindly' to `get_subscription_response` as it complains that it doesn't recognise `:id`.

The allowed parameters are in the defaults (`title`, `subscriber_id`, `subscriber_list_id`, `ended`), apart from `id` (which will be stripped out) and `frequency`. If `frequency` is omitted it will error, and anything extra that's added accidentally will cause the `ArgumentError`. Could have had a default for 'frequency' but think it's better for it to be explicitly provided every time. Hope that makes sense!

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: > The return of results may not always be in time order.

What do you mean by this? Does Ruby not guarantee the order of arrays being traversed? I know in JavaScript you can't rely on the order of JSON properties being traversed, but don't expect that to be a problem here...?

The `stub_email_alert_api_has_subscriptions` works on the basis that the subscriptions are applied in the order they are passed, i.e. you should not be putting the most recent subscription halfway through the arguments.

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Fair point - I'll make the comment more descriptive 👍

lib/gds_api/test_helpers/email_alert_api.rb
brucebolt: I see, so the input to `stub_email_alert_api_has_subscriptions` will always be in the correct order, then `.last` is fine here.

lib/gds_api/test_helpers/email_alert_api.rb
ChrisBAshton: Have tweaked the comment to hopefully make this clearer.

lib/gds_api/test_helpers/email_alert_api.rb
brucebolt: Ah, yes, I see now.  I cannot see any way the two loops could be condensed into one without any further code complexity elsewhere.

PR #933, opened 2019-05-23
Bump to 59.3.1

ChrisBAshton:
To release the work done in https://github.com/alphagov/gds-api-adapters/pull/932

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 59.3.1
 
 * Adds `stub_email_alert_api_has_subscriptions` test helper method.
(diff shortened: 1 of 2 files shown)


ChrisBAshton:
Fair point. I hadn't thought of it as new functionality as it is a new test stub method rather than a new API method but on second thoughts you're right!

brucebolt:
The history suggests both patch and minor releases have been used for new stubs, so there's no consistent history.  I'd be happier if it were a minor release.

ChrisBAshton:
Have raised https://github.com/alphagov/gds-api-adapters/pull/934.

PR #937, opened 2019-06-10
Allow stubbing of email-alert-api requests that use combine_mode

brucebolt:
Add `combine_mode` parameter to `build_subscriber_lists_url` to allow us to stub requests to email-alert-api where a value for `combine_mode` has been specified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 59.5.1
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Unreleased
+# 59.5.1
```
leenagupte: You should really do this in a separate PR.

PR #943, opened 2019-07-16
Add access to whitehall endpoint to test - wip

1pretz1:
https://trello.com/c/gtyLTQ8P/985-work-out-how-to-get-data-out-of-whitehall


cbaines:
If the intent is to just use this in Content Publisher, it might be less costly to just put this functionality in Content Publisher.

PR #944, opened 2019-07-17
Use debug level for logger when forwarding headers

theseanything:
The changes the existing logging at info level to debug level to prevent noisy logging. Also fixed header casing.

### DIFF
--- a/lib/gds_api/railtie.rb
+++ b/lib/gds_api/railtie.rb
@@ -3,27 +3,27 @@ require_relative 'middleware/govuk_header_sniffer'
 module GdsApi
   class Railtie < Rails::Railtie
     initializer "gds_api.initialize_govuk_request_id_sniffer" do |app|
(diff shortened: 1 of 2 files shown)


lib/gds_api/railtie.rb
```diff
@@ -3,27 +3,27 @@
 module GdsApi
   class Railtie < Rails::Railtie
     initializer "gds_api.initialize_govuk_request_id_sniffer" do |app|
-      Rails.logger.info "Using middleware GdsApi::GovukHeaderSniffer to sniff for GOVUK-Request-Id header"
+      Rails.logger.debug "Using middleware GdsApi::GovukHeaderSniffer to sniff for GOVUK-Request-Id header"
```
kevindew: I think this and the one below should be Govuk rather than caps.

theseanything:
@kevindew fixed the casing!

PR #947, opened 2019-07-24
Stop publishing released version pact

kevindew:
There was an intention that we would have pact versions stored that
reflected each version of the gem that was published. So we have a pact
for version 59.5.1 of GDS API Adapters published [1] and that correlates
to version 59.5.1 of the gem.

This however is incorrect, the pact is published any time master builds
which is often numerous times before a release. This means that the
pacts are overridden with changes made after their release.

To resolve this we could set up some sort of checks to see if the pact
is already published or not but this feels unnecessary since it doesn't
appear that Publishing API uses any of these.

[1]: https://pact-broker.cloudapps.digital/hal-browser/browser.html#/pacticipants/GDS%20API%20Adapters/versions/59.5.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Removes publishing pacts for released versions as they incorrectly referenced
+  unreleased changes
+
 # 60.1.0
 
 * Update get_subscriptions endpoint for email-alert-api to accept order param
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Released version is no published as a pact as it was inaccurate
```
theseanything:
```suggestion
* Removes publishing pacts for released versions as they incorrectly referenced unreleased commits.
```

Also don't know if this should be in the change log, as it's not a user facing change?

kevindew: Argh my English 🙈 thanks, I've updated.

It's somewhat moot since no-one uses them but I figured since it was something public that is consumed by consumers then it counts as a user land change

PR #948, opened 2019-08-22
Rename the Email Alert API send_alert method to create_content_change

thomasleese:
It also renames the associated test helper method. This reflects a change in the underling endpoint from `/notifications` to `/content-changes`. https://github.com/alphagov/email-alert-api/pull/931

This is a breaking change for users of the Email Alert API adapters. I've decided to make it a breaking change rather than deprecrate the old method because there aren't too many apps which talk to email-alert-api so it should be pretty quick to fix them one by one.

[Trello Card](https://trello.com/c/5rtAmAjc/54-finish-the-messages-endpoint-work-in-email-alert-api)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
   related test helper methods) to reflect a change in the underling endpoint.
   **Note:** this is a breaking change for users of the Email Alert API
   adapters.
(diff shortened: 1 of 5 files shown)


thomasleese:
The tests are failing but they seem unrelated to this change since master is also failing for me with the same error. Just trying to figure out why...

PR #961, opened 2019-11-26
Add and extend stubs for email-alert-api

benthorner:
https://trello.com/c/dIAlk2dp/278-double-opt-in-create-a-subscription-before-the-success-page

This extends one of the existing stubs to make it more flexible,
as well as adding a few more for edge case scenarios.

I was tempted to just have a generic 'stub_request' one-liner for
the edge case stubs, but on reflection think it's better if the
stubs fully communicate the realistic API behaviour.

### DIFF
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -37,11 +37,13 @@ module GdsApi
           .to_return(status: 404)
       end
 
-      def stub_email_alert_api_has_subscriber_subscriptions(id, address, order)
-        stub_request(:get, "#{EMAIL_ALERT_API_ENDPOINT}/subscribers/#{id}/subscriptions?order=#{order}")
+      def stub_email_alert_api_has_subscriber_subscriptions(id, address, order = nil, subscriptions: nil)
+        params = order ? "?order=#{order}" : ""
+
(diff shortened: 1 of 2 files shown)


lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -240,6 +245,16 @@ def stub_email_alert_api_creates_an_auth_token(subscriber_id, address)
           )
       end
 
+      def stub_email_alert_api_invalid_auth_token
+        stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscribers/auth-token")
+          .to_return(stauts: 422)
```
emmabeynon: Looks like there's a typo here: "stauts" => "status"

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -240,6 +245,16 @@ def stub_email_alert_api_creates_an_auth_token(subscriber_id, address)
...
+      end
+
+      def stub_email_alert_api_auth_token_no_subscriber
+        stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscribers/auth-token")
+          .to_return(stauts: 404)
```
emmabeynon: Looks like there's a typo here: "stauts" => "status"

benthorner:
Thanks for spotting - fixed 👍

PR #963, opened 2019-11-27
Creates subscriber and subscription verification adapters

huwd:
Relates to: https://trello.com/c/9FmE3rPn/281-double-opt-in-create-api-adapter-to-send-double-opt-in-email

---

Hold until https://trello.com/c/Dte8k1Aa/280-double-opt-in-create-an-internal-api-to-send-double-opt-in-emails is merged or API is confirmed

---

Deprecates the existing create_auth_token and replaces it with send_subscriber_verification_email.
Also adds a new adapter for sending an email to verify a subscription.

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -222,7 +222,7 @@ class GdsApi::EmailAlertApi < GdsApi::Base
 
   # Verify a subscriber has control of a provided email
   #
-  # @param [string]       address       Email address of subscriber to create token for
+  # @param [string]       address       Address to send verification email to
   # @param [string]       destination   Path on GOV.UK that subscriber will be emailed
   # @param [string, nil]  redirect      Path on GOV.UK to be encoded into the token for redirecting
   #
@@ -234,6 +234,23 @@ class GdsApi::EmailAlertApi < GdsApi::Base
       address: address,
       destination: destination,
       redirect: redirect,
+    )
+  end
+
+  # Verify a subscriber intends to be added to a subscription
+  #
+  # @param [string]       address       Address to send verification email to
+  # @param [string]       frequency     How often the subscriber wishes to be notified of new items
+  # @param [string]       topic_id      The slugs/ID for the topic being subscribed to
+  #
+  # return [Hash]  subscription
+  #
+  def send_subscription_verification_email(address:, frequency:, topic_id:)
+    post_json(
+      "#{endpoint}/subscriptions/auth-token",
+      address: address,
+      frequency: frequency,
+      topic_id: topic_id,
     )
   end
 
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -220,7 +220,7 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     )
   end
 
-  # Create an authentication token for a subscriber
+  # Verify a subscriber has control of a provided email
   #
   # @param [string]       address       Email address of subscriber to create token for
   # @param [string]       destination   Path on GOV.UK that subscriber will be emailed
@@ -228,7 +228,7 @@ class GdsApi::EmailAlertApi < GdsApi::Base
   #
   # @return [Hash]  subscriber
   #
-  def create_auth_token(address:, destination:, redirect: nil)
+  def send_subscriber_verification_email(address:, destination:, redirect: nil)
     post_json(
       "#{endpoint}/subscribers/auth-token",
       address: address,
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -239,7 +239,8 @@ module GdsApi
         ).to_return(status: 422)
       end
 
-      def stub_email_alert_api_creates_an_auth_token(subscriber_id, address)
+
+      def stub_email_alert_api_sends_subscriber_verification_email(subscriber_id, address)
         stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscribers/auth-token")
           .to_return(
             status: 201,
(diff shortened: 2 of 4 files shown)


lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -307,12 +315,23 @@ def stub_email_alert_api_does_not_have_subscriber_list_by_slug(slug:)
       alias_method :email_alert_api_creates_a_subscription, :stub_email_alert_api_creates_a_subscription
       alias_method :email_alert_api_creates_an_existing_subscription, :stub_email_alert_api_creates_an_existing_subscription
       alias_method :email_alert_api_refuses_to_create_subscription, :stub_email_alert_api_refuses_to_create_subscription
-      alias_method :email_alert_api_creates_an_auth_token, :stub_email_alert_api_creates_an_auth_token
       alias_method :email_alert_api_has_subscriber_list_by_slug, :stub_email_alert_api_has_subscriber_list_by_slug
       alias_method :email_alert_api_does_not_have_subscriber_list_by_slug, :stub_email_alert_api_does_not_have_subscriber_list_by_slug
+      alias_method :stub_email_alert_api_creates_an_auth_token, :stub_email_alert_api_sends_subscriber_verification_email
```
huwd: Might want to check I've got these right, my aims were to preserve the older methods and route them to the newer ones.

lib/gds_api/email_alert_api.rb
```diff
@@ -220,15 +220,21 @@ def change_subscription(id:, frequency:)
     )
   end
 
-  # Create an authentication token for a subscriber
+  # Deprecated method, issues warning and runs alias send_subscriber_verification_email
+  def create_auth_token(address:, destination:, redirect: nil)
```
huwd: Could lose this, wasn't sure if giving a warning was better than just aliasing create_auth_token to send_subscriber_verification_email

koetsier:
Might want to split up the commits to have one do the renaming and one adding the new api endpoint

lib/gds_api/email_alert_api.rb
```diff
@@ -237,6 +243,23 @@ def create_auth_token(address:, destination:, redirect: nil)
     )
   end
 
+  # Verify a subscriber intends to be added to a subscription
+  #
+  # @param [string]       address       Email address of subscriber to create token for
+  # @param [string]       frequency     How often the subscriber wishes to be notified of new items
+  # @param [string]       topic_id      The slugs/IDs for topics to be subscribed to
```
benthorner: This is actually a singular item.

lib/gds_api/email_alert_api.rb
```diff
@@ -237,6 +243,23 @@ def create_auth_token(address:, destination:, redirect: nil)
     )
   end
 
+  # Verify a subscriber intends to be added to a subscription
+  #
+  # @param [string]       address       Email address of subscriber to create token for
```
benthorner: How about: `Address to send the verification email to`?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -307,12 +315,23 @@ def stub_email_alert_api_does_not_have_subscriber_list_by_slug(slug:)
...
+      alias_method :email_alert_api_sends_subscriber_verification_email, :stub_email_alert_api_sends_subscriber_verification_email
 
     private
 
+      def get_subscription_verification_response(id, address, topic_id)
```
benthorner: I don't think this is right. Comparing [with the controller action](https://github.com/alphagov/email-alert-api/blob/master/app/controllers/subscriptions_auth_token_controller.rb#L13), I think we just return an empty hash.

benthorner: Also, the status code is 200/OK.

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -307,12 +315,23 @@ def stub_email_alert_api_does_not_have_subscriber_list_by_slug(slug:)
       alias_method :email_alert_api_creates_a_subscription, :stub_email_alert_api_creates_a_subscription
       alias_method :email_alert_api_creates_an_existing_subscription, :stub_email_alert_api_creates_an_existing_subscription
       alias_method :email_alert_api_refuses_to_create_subscription, :stub_email_alert_api_refuses_to_create_subscription
-      alias_method :email_alert_api_creates_an_auth_token, :stub_email_alert_api_creates_an_auth_token
       alias_method :email_alert_api_has_subscriber_list_by_slug, :stub_email_alert_api_has_subscriber_list_by_slug
       alias_method :email_alert_api_does_not_have_subscriber_list_by_slug, :stub_email_alert_api_does_not_have_subscriber_list_by_slug
+      alias_method :stub_email_alert_api_creates_an_auth_token, :stub_email_alert_api_sends_subscriber_verification_email
+      alias_method :email_alert_api_sends_subscriber_verification_email, :stub_email_alert_api_sends_subscriber_verification_email
```
benthorner: Since this method is brand new, we don't need aliases for it, since there are no old usages of it.

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -239,20 +239,30 @@ def stub_email_alert_api_refuses_to_create_subscription(subscriber_list_id, addr
         ).to_return(status: 422)
       end
 
-      def stub_email_alert_api_creates_an_auth_token(subscriber_id, address)
+      def stub_email_alert_api_sends_subscription_verification_email(subscriber_id, address, topic_id)
```
benthorner: Comparing with the adapter, I think we need to replace `subscriber_id` -> `frequency`.

benthorner: I think I introduced this one 😬

PR #964, opened 2019-12-04
Suppress and check for deprecated warning

benthorner:
https://trello.com/c/9FmE3rPn/281-double-opt-in-create-api-adapter-to-send-double-opt-in-email

This stops the many 'deprecated' warnings for Rummager cluttering the
test output when running 'rake'.

### DIFF
--- a/test/rummager_test.rb
+++ b/test/rummager_test.rb
@@ -7,6 +7,10 @@ describe GdsApi::Rummager do
     stub_request(:get, /example.com\/batch_search/).to_return(body: "[]")
   end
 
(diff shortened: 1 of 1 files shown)


issyl0:
Can we instead stop using the deprecated functionality?

benthorner:
@issyl0 indeed, although I'm not sure where that might be. @barrucadu @sihugh do you know if we're still using these adapters anywhere?

benthorner:
Superseded by https://github.com/alphagov/gds-api-adapters/pull/966

PR #968, opened 2019-12-11
Add `stub_publishing_api_path_reservation` test helper method

leenagupte:
Trello: https://trello.com/c/NC10wudi/1250-add-a-test-helper-to-stub-the-putpath-method-in-gds-api-adapters

Follows on from: https://github.com/alphagov/content-publisher/commit/5e58e4cf242d0549546e675f3143b69c1b472880

## What's changed?

Allow the path reservation for a specific base path to be stubbed.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api.rb
+++ b/lib/gds_api/test_helpers/publishing_api.rb
@@ -689,6 +689,28 @@ module GdsApi
         end
       end
 
+      # Stub a PUT /v2/paths/:base_path request with the given content id and request body.
+      #
+      # @example
+      #   stub_publishing_api_path_reservation(
+      #     my_content_id,
+      #     publishing_app: "content-publisher",
+      #     override_existing: true,
+      #   )
+      #
+      # @param base_path [String]
+      # @param params [Hash]
+      def stub_publishing_api_path_reservation(base_path, params = {})
+        url = PUBLISHING_API_ENDPOINT + "/paths#{base_path}"
+        response = {
+          status: 200,
+          headers: { content_type: "application/json" },
+          body: params.merge(base_path: base_path).to_json,
+        }
+
+        stub_request(:put, url).with(body: params).to_return(response)
+      end
+
       def stub_default_publishing_api_path_reservation
         stub_request(:put, %r[\A#{PUBLISHING_API_ENDPOINT}/paths/]).to_return { |request|
           base_path = request.uri.path.sub(%r{\A/paths/}, "")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Add `stub_publishing_api_path_reservation` test helper method.
 * Combine `GdsApi::PublishingAPI` and `GdsApi::PublishingApiV2`
 
 # 63.0.0


lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -75,6 +75,29 @@ def stub_publishing_api_isnt_available
...
+      #
+      # @param content_id [UUID]
+      # @param body  [String]
+      # @param response_hash [Hash]
+      def stub_publishing_api_path_reservation(base_path, body, response_hash = {})
+        url = PUBLISHING_API_ENDPOINT + "/paths#{base_path}"
+        response = {
+          status: 200,
+          body: "{}",
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
```
kevindew: Any reason we've done this differently to https://github.com/alphagov/gds-api-adapters/blob/ecec6d7cda2d4d4fdce8c35e1e901029f33de329/lib/gds_api/test_helpers/publishing_api.rb#L104 ?

lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -75,6 +75,29 @@ def stub_publishing_api_isnt_available
...
+      #   )
+      #
+      # @param content_id [UUID]
+      # @param body  [String]
+      # @param response_hash [Hash]
+      def stub_publishing_api_path_reservation(base_path, body, response_hash = {})
+        url = PUBLISHING_API_ENDPOINT + "/paths#{base_path}"
+        response = {
+          status: 200,
+          body: "{}",
```
kevindew: We probably should use the method to set the response: https://github.com/alphagov/gds-api-adapters/blob/ecec6d7cda2d4d4fdce8c35e1e901029f33de329/lib/gds_api/test_helpers/publishing_api.rb#L105

kevindew: We also probably want to change the response_hash argument to be response_body and pass that into https://github.com/alphagov/gds-api-adapters/blob/ecec6d7cda2d4d4fdce8c35e1e901029f33de329/lib/gds_api/test_helpers/publishing_api.rb#L174 as the override attributes

lib/gds_api/test_helpers/publishing_api.rb
leenagupte: Nope. I just copied the response from the publishing-api V2 helpers.

lib/gds_api/test_helpers/publishing_api.rb
leenagupte: Oh. `put_path` only returns `success` or `failure` so I can probably combine `body` and `response_hash`.

lib/gds_api.rb
```diff
@@ -138,17 +137,14 @@ def self.publishing_api(options = {})
...
   # variable is set
   #
-  # @return [GdsApi::PublishingApiV2]
+  # @return [GdsApi::PublishingApi]
   def self.publishing_api_v2(options = {})
-    GdsApi::PublishingApiV2.new(
-      Plek.find("publishing-api"),
-      { bearer_token: ENV["PUBLISHING_API_BEARER_TOKEN"] }.merge(options),
-    )
+    self.publishing_api(options)
```
kevindew: I think this should create an instance of `GdsApi::PublishingApiV2` still so that it continues to return the same class.

lib/gds_api/publishing_api_v2.rb
```diff
@@ -1,497 +1,9 @@
...
-    optional_keys.each_with_object(params) do |optional_key, hash|
-      hash.merge!(optional_key => options[optional_key]) if options[optional_key]
-    end
-  end
-
-  def validate_content_id(content_id)
-    raise ArgumentError, "content_id cannot be nil" unless content_id
-  end
-end
+class GdsApi::PublishingApiV2 < GdsApi::PublishingApi; end
```
kevindew: There's probably a deprecation warning opportunity in the initialize method for this class: https://github.com/alphagov/gds-api-adapters/blob/5d732cc35ffc7441da6c7cfb7ac3fdb5d56464dc/lib/gds_api/rummager.rb

lib/gds_api/test_helpers/publishing_api_v2.rb
```diff
@@ -5,698 +5,8 @@
...
-      def version_conflict(expected_version, actual_version = expected_version + 1)
-        {
-          error: {
-            code: 409,
-            message: "A lock-version conflict occurred. The `previous_version` you've sent (#{expected_version}) is not the same as the current lock version of the edition (#{actual_version}).",
-            fields: { previous_version: ["does not match"] },
-          },
-        }
-      end
+      include PublishingApi
```
kevindew: We've got a deprecation warning opportunity here: https://github.com/alphagov/gds-api-adapters/blob/5d732cc35ffc7441da6c7cfb7ac3fdb5d56464dc/lib/gds_api/test_helpers/rummager.rb#L10-L12

lib/gds_api.rb
leenagupte: I'll remove this commit and fix up.

leenagupte:
@kevindew Thanks for your comments. I've addressed the ones regarding combining the publishing-api helpers here: https://github.com/alphagov/gds-api-adapters/pull/969

PR #969, opened 2019-12-17
Combine publishing api helpers

leenagupte:
Trello: https://trello.com/c/5BYnJ2A9

## What's changed and why?

Publishing API has two endpoints, the original and a "V2" one. There has been some effort to reflect is in gds-api-adapters, but this isn't always consistent. The path reservation helper was moved to the "V2" class, but the endpoint it uses is for the original publishing-api
endpoint.

Combining the classes means that the caller doesn't need to worry about calling the correct version of publishing-api, though the "V2" classes haven't been removed to make it backwards compatible.


cbaines:
I'm guessing the intent in having a `V2` API was that this would eventually replace the V1 API.

Obviously that hasn't happened at any point in the last few years, but this change would make uses of the V1 API harder to find, as you couldn't just search for uses of the `GdsApi::PublishingApi` class

kevindew:
> I'm guessing the intent in having a `V2` API was that this would eventually replace the V1 API.
> 
> Obviously that hasn't happened at any point in the last few years, but this change would make uses of the V1 API harder to find, as you couldn't just search for uses of the `GdsApi::PublishingApi` class

There are now no longer any v1 alternative methods in GdsApi::PublishingApi since it just has put_intent, destroy_intent, and unreserve_path left. So it's not so much that there are v1 methods still in operation just a confusing muddle in the Publishing API about a /v2/ route prefix and an odd balance where we have a those without that prefix sometimes in `GdsApi::PublishingApi` and other times in `GdsApi::PublishingApiV2`

PR #970, opened 2019-12-18
Fix various put path related issues for Publishing API helpers

kevindew:
In https://github.com/alphagov/gds-api-adapters/pull/968 we discovered that the stub methods for the Publishing API's PUT /path method were returning strange custom errors and responses that weren't representative of what the Publishing API actually did.

I've sorted those out and fixed various other issues in the immediate vicinity.

### DIFF
--- a/lib/gds_api/test_helpers/publishing_api.rb
+++ b/lib/gds_api/test_helpers/publishing_api.rb
@@ -652,7 +652,7 @@ module GdsApi
         stub_request(:delete, url).to_return(status: 200, body: "{}", headers: { "Content-Type" => "application/json; charset=utf-8" })
       end
 
-      def stub_default_publishing_api_put_intent
+      def stub_any_publishing_api_put_intent
         stub_request(:put, %r{\A#{PUBLISHING_API_ENDPOINT}/publish-intent})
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -754,12 +754,27 @@ def stub_publishing_api_has_path_reservation_for(path, publishing_app)
...
 
-      def stub_publishing_api_returns_path_reservation_validation_error_for(path, error_details = nil)
-        error_details ||= { "base" => ["computer says no"] }
-        error_data = publishing_api_path_data_for(path).merge("errors" => error_details)
+      # Stub a PUT /paths/:base_path request for a particular publishing
+      # application. Calling for a different publishing application will return
+      # a 422 response.
+      #
+      # @example
+      #   stub_publishing_api_has_path_reservation_for("/foo", "content-publisher")
```
leenagupte: Copy and paste error: The example doesn't match the method definition.

kevindew: Good catch - sorted.

PR #975, opened 2020-01-03
Add new method to get live content

leenagupte:
Trello: https://trello.com/c/JHMoTkfA
Depends on: https://github.com/alphagov/publishing-api/pull/1679

## What's changed and why?
By default, `get_content` gets the latest version of the content item regardless of the publication state. Sometimes, it is useful to be able to get the "live" version of the content.

This code has already been implemented in [collections-publisher](https://github.com/alphagov/collections-publisher/blob/f04fdc32409165f238b32245d8189faa5ac216a3/app/helpers/publishing_api_helper.rb) to discard drafts, but now something similar needs to be added to content-publisher to allow content items migrated from Whitehall to be compared with the information publishing-api already has about them, so it seemed like a good time to add this functionality somewhere common.

### DIFF
--- a/lib/gds_api/publishing_api.rb
+++ b/lib/gds_api/publishing_api.rb
@@ -8,6 +8,8 @@ require_relative "exceptions"
 # @see https://github.com/alphagov/publishing-api/blob/master/doc/model.md
 # @api documented
 class GdsApi::PublishingApi < GdsApi::Base
+  class NoLiveVersion < GdsApi::BaseError; end
+
   # Put a content item
   #
   # @param content_id [UUID]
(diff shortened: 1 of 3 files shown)


lib/gds_api/publishing_api.rb
```diff
@@ -34,6 +36,31 @@ def get_content(content_id, params = {})
     get_json(content_url(content_id, params))
   end
 
+  # Return a live content item
```
kevindew: Probably worth expanding this comment to state that live means published or unpublished.

lib/gds_api/publishing_api.rb
```diff
@@ -34,6 +36,31 @@ def get_content(content_id, params = {})
...
+  #
+  # Raises exception if the item doesn't exist.
+  # Raises an exception if the version number is passed in the params
+  #
+  # @param content_id [UUID]
+  # @param locale [String]
+  # @option locale [String] locale The language, defaults to 'en' in publishing-api.
```
kevindew: How come we have a @param and @option?

lib/gds_api/publishing_api.rb
```diff
@@ -34,6 +36,31 @@ def get_content(content_id, params = {})
     get_json(content_url(content_id, params))
   end
 
+  # Return a live content item
+  #
+  # Raises exception if the item doesn't exist.
+  # Raises an exception if the version number is passed in the params
```
kevindew: Do we need to state these if we have a raise part of this docblock?

test/publishing_api_test.rb
```diff
@@ -514,6 +514,195 @@ def content_item_for_content_id(content_id, attrs = {})
...
+      end
+
+      it "responds with NoLiveVersion" do
+        assert_raises(GdsApi::PublishingApi::NoLiveVersion) do
+          @api_client.get_live_content(@content_id)
+        end
+      end
+    end
+
+    describe "when an older version of the content item is live" do
```
kevindew: I think this describe isn't as clear as it could be. Might be better to phrase it as "when there is a draft version of live content" ?

test/publishing_api_test.rb
```diff
@@ -514,6 +514,195 @@ def content_item_for_content_id(content_id, attrs = {})
...
+      end
+
+      it "returns the live content item" do
+        response = @api_client.get_live_content(@content_id)
+        assert_equal 200, response.code
+        assert_equal response["publication_state"], "published"
+      end
+    end
+
+    describe "when the content item has been unpublished" do
```
kevindew: This assertion feels a bit out of place since we don't have one about when content is published, just one about live. 

It feels like we'd either want to test both published and unpublished or we just want to test live as one assertion.

lib/gds_api/publishing_api.rb
leenagupte: Oops! I'll remove `@param` as locale is optional.

lib/gds_api/publishing_api.rb
leenagupte: I'll remove the version number statement as that's no longer true.

test/publishing_api_test.rb
leenagupte: I'll update the test description to match.

PR #976, opened 2020-01-03
Some small Publishing API fixes

kevindew:
This fixes a few small issues I found with Publishing API when I was writing the code https://github.com/alphagov/gds-api-adapters/pull/974. More details in the commits.


test/test_helpers/publishing_api_test.rb
```diff
@@ -328,19 +328,15 @@
...
+        publishing_api.lookup_content_id(base_path: "")
       end
-    end
-
-    it "returns a 503 for V1 requests" do
-      stub_publishing_api_isnt_available
 
       assert_raises GdsApi::BaseError do
-        publishing_api.lookup_content_id(base_path: "")
+        publishing_api.get_content_items({})
```
leenagupte: It's odd that these lines have switched order, but ok.

lib/gds_api/test_helpers/publishing_api.rb
```diff
@@ -643,10 +641,33 @@ def stub_publishing_api_unreserve_path_invalid(base_path, publishing_app = /.*/)
...
+      #   )
+      #
+      # @param base_path [String]
+      # @param params [Hash]
+      def stub_publishing_api_put_intent(base_path, params = {})
+        url = PUBLISHING_API_ENDPOINT + "/publish-intent#{base_path}"
+        body = if params.is_a?(String)
+                 params
+               else
+                 { base_path: base_path }.merge(params).to_json
```
leenagupte: Why does `base_path` need to be merged here?

test/test_helpers/publishing_api_test.rb
kevindew: Sure, unintentional. I've swapped them back now.

lib/gds_api/test_helpers/publishing_api.rb
kevindew: Ah good catch - I was under the impression that Publishing API did return the payload with a base_path (as per [path_reservation](https://github.com/alphagov/gds-api-adapters/blob/6193489874ce1305ec093f1a0ff14712e440940f/lib/gds_api/test_helpers/publishing_api.rb#L710)) 

However because of the rather strange conditional in https://github.com/alphagov/publishing-api/blob/master/app/commands/put_publish_intent.rb#L6-L9 it won't actually include a base_path.

kevindew:
@leenagupte I've amended

Issue #979, opened 2020-01-16
Apps will be accessing worldwide api via whitehall-frontend after migration to AWS


whi-tw:
The following PRs:
* https://github.com/alphagov/smart-answers/pull/4282
* https://github.com/alphagov/contacts-admin/pull/696
* https://github.com/alphagov/contacts-admin/pull/696

Change the respective apps to access the worldwide API via `whitehall-frontend` not `whitehall-admin`.

I believe [this line](https://github.com/alphagov/gds-api-adapters/blob/6f044f68b3b7dc3bda4acbc2a52e5738838264e4/lib/gds_api/test_helpers/worldwide.rb#L11) will have to be modified to reflect this in the test helpers

whi-tw:
resolved in #980

PR #980, opened 2020-01-16
Switch the Worldwide API endpoint

cbaines:
The whitehall-frontend and whitehall-admin hosts should provide
identical responses, as they're the same code. The whitehall-frontend
hosts are setup for handling frontend requests, and have a read-only
database connection. The whitehall-admin hosts are setup for handling
the admin interface.

As the Worldwide API doesn't do any writing, it's suitable to be
handled by the whitehall-frontend hosts. This change is motivated by
the migration to AWS, as it'll be easier to keep whitehall-frontend up
during the migration (compared to whitehall-admin).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+* Change Worldwide API requests to be routed to whitehall-frontend by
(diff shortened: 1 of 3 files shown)


whi-tw:
resolves #979

PR #991, opened 2020-04-16
Disable buggy gemspec Ruby version cop

benthorner:
Please see the commits for more details.


benthorner:
@laurentqro I think you had a better idea with increasing the minimum version, as disabling the target version cop could mean Rubocop suggests things that aren't compatible with the minimum version. I've updated to reflect this - could you take another look?

laurentqro:
@benthorner Agreed! Looks good to me 👌

PR #992, opened 2020-04-17
Remove params from error reporting

laurentqro:
The request body can contain personally identifiable information, like email addresses, so I suggest we remove the params from the reporting.

This will probably cause pain for certain debugging scenarios, but trying to filter out all PII is a very hard challenge and will never be perfect for all future scenarios.

We would probably do well to use Sentry's `context` object in each application, where we can configure which fields should be redacted for that application: https://docs.sentry.io/clients/ruby/context/#additional-context.

---

https://trello.com/c/bfmbw28t/89-remove-pii-data-on-email-subscriptions-from-sentry

### DIFF
--- a/lib/gds_api/exceptions.rb
+++ b/lib/gds_api/exceptions.rb
@@ -22,11 +22,10 @@ module GdsApi
   class HTTPErrorResponse < BaseError
     attr_accessor :code, :error_details
 
-    def initialize(code, message = nil, error_details = nil, request_body = nil)
+    def initialize(code, message = nil, error_details = nil)
       super(message)
       @code = code
       @error_details = error_details
-      @request_body = request_body
     end
   end
 
@@ -53,8 +52,8 @@ module GdsApi
   class HTTPGatewayTimeout < HTTPIntermittentServerError; end
 
   module ExceptionHandling
-    def build_specific_http_error(error, url, details = nil, request_body = nil)
-      message = "URL: #{url}\nResponse body:\n#{error.http_body}\n\nRequest body:\n#{request_body}"
+    def build_specific_http_error(error, url, details = nil)
+      message = "URL: #{url}\nResponse body:\n#{error.http_body}"
       code = error.http_code
       error_class_for_code(code).new(code, message, details)
     end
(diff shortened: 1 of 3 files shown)


kevindew:
This is looking good to me 👍 and seems reasonable. I imagine we get most value from response rather than params (you'd like to think a developer could work out the params).

I see this is a draft and failing build so didn't want to give this an approval just yet but shout if this is all done.

lib/gds_api/exceptions.rb
```diff
@@ -53,8 +53,8 @@ class HTTPUnavailable < HTTPIntermittentServerError; end
   class HTTPGatewayTimeout < HTTPIntermittentServerError; end
 
   module ExceptionHandling
-    def build_specific_http_error(error, url, details = nil, request_body = nil)
-      message = "URL: #{url}\nResponse body:\n#{error.http_body}\n\nRequest body:\n#{request_body}"
+    def build_specific_http_error(error, url, details = nil)
```
vanitabarrett, line 56: Does this change mean we can remove `request_body` from `initialise()` too? 
https://github.com/alphagov/gds-api-adapters/blob/119ccfcb4659ee7afda1664923f8caca1d219700/lib/gds_api/exceptions.rb#L25

benthorner:
@laurentqro I put this one up last week, which should help with the build: https://github.com/alphagov/gds-api-adapters/pull/991

lib/gds_api/exceptions.rb
laurentqro: Yeah it looks like it, good spot. Actually, it looks like we're never passing that argument from anywhere in the codebase, and we never noticed, because the argument value defaults to `nil`.

laurentqro: Just noticed that if we do ever want request body info in Sentry, we could pass it to this `raven_context` method in `BaseError` https://github.com/alphagov/gds-api-adapters/blob/119ccfcb4659ee7afda1664923f8caca1d219700/lib/gds_api/exceptions.rb#L6, and make full use of Sentry's key filtering.

PR #994, opened 2020-05-17
Run RuboCop as part of `bundle exec rake` defaults

issyl0:
- This is a prerequisite for being able to move more of our apps to
  GitHub Actions. We want one single step bundle exec rake that runs
  everything. This ensures parity between local dev and what CI runs
  (something that we don't always have now).

https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,3 +4,7 @@ inherit_gem:
 
 AllCops:
   TargetRubyVersion: 2.4
+
+Naming/FileName:
+  Exclude:
+    - lib/gds-api-adapters.rb
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,4 +7,7 @@ AllCops:
 
 Naming/FileName:
   Exclude:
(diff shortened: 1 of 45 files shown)


.rubocop.yml
```diff
@@ -4,3 +4,7 @@ inherit_gem:
 
 AllCops:
   TargetRubyVersion: 2.4
+
+Naming/FileName:
+  Exclude:
+    - lib/gds-api-adapters.rb
```
benthorner, line 10: Could we add a comment to indicate why this is necessary?

issyl0:
```suggestion
    # RuboCop wants this to be renamed to `lib/gds_api_adapters.rb`. But I don't know what that affect that'll have on the rest of this gem and don't particularly have the brain to find out. So I'm ignoring it for now. -- Issy, 2020-05.
    - lib/gds-api-adapters.rb
```

PR #997, opened 2020-05-27
Release 67.0.0

kevindew:
See: https://github.com/alphagov/gds-api-adapters/pull/995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 67.0.0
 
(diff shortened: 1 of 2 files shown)


theseanything:
Just curious - how come we went from v63 to v67 instead of v64?

kevindew:
Arrghh! 🤦 

I incremented the minor version (6) to end up with 7. Hmmm how best to resolve this.

leenagupte:
It turns out that I can't count either 🤦

PR #999, opened 2020-07-05
Remove dependency on rack-cache gem

pixeltrix:
This gem is using in a large number of apps and it's always a good idea to reduce dependencies where possible since each gem's lib directory is added to the load path and will be scanned when using `require`.

Using rack-cache to parse the Cache-Control header made sense when the gem was included as a dependency of the actionpack gem but since the release of Rails 4.0 it's no longer included by default. This means that the rack-cache dependency is forced on all apps that use this gem, regardless of whether they use it elsewhere (in fact some apps explicitly set `config.rack_cache = false`). By reimplementing this internally we can remove the dependency for all of those apps that don't need it.

Since the `cache_control` method on the response object is public this commit reimplements the exact same functionality even though the response object only ever uses the `max_age` method.

**NOTE: The implementation of the `CacheControl` class inevitably mirrors closely the implementation in the rack-cache gem (however the method of parsing the header is different). It also borrows the structure of the tests to ensure compatibility but they have been rewritten to follow the assertion style of the tests in this repo and not the original expectation style in the rack-cache repo. Due to this I'm unsure whether we need to include an attribution to comply with the MIT license and if we do where it should go.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Remove dependency on rack-cache gem
+
 # 67.0.0
 
 * Note: this release was misnumbered and should have been version 64 - there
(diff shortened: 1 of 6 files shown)


pixeltrix:
Let me know if that's not enough (most other entries appeared to be single lines so didn't want put a wall of text in there).

PR #1000, opened 2020-07-20
URI encode strings passed into Email Alert API URLs

kevindew:
Trello: https://trello.com/c/koo6gJov/355-increase-visibility-of-sentry-alerts-for-email-apps-and-services

We see errors in Sentry [1] where clients are requesting information from
Email Alert Frontend with dubious strings that are then passed onto
email Alert API raw. To resolve this I've applied the `uri_encode`
function to all of the parameters that are used in Email Alert API URIs,
this will filter out any characters that should not be in a URL
including hacking attempts. The alternative to this would be to filter
characters in Email Alert Frontend before sending the data to Email
Alert API, this was rejected as it meant duplicating Email Alert API
logic in Email Alert Frontend.

In applying this I decided that this should be done on all URIs for
Email Alert API and not just particular exposed ones, this is because
this repo should not have any knowledge over how these endpoints are
used. Thus, if we consider one exposed to user input we should consider
them all the same.

Ideally this uri encoding would be done implicitly rather than needing
explicit definitions for each uri. It seemed a rather bold thing to
define at this point in the project but something like
`put_json("unsubscribe/:id", params: { id: "test" })` would be preferable.

A pain point revealed by the explicit need was deciding what level of
testing to do. Testing every method felt a bit over-the-top, but that's
the approach I ended up taking since there I felt that it should be
tested and no method should seem special.

[1]: https://sentry.io/organizations/govuk/issues/468258104/?project=202221&query=&statsPeriod=14d


lib/gds_api/email_alert_api.rb
```diff
@@ -101,7 +101,7 @@ def topic_matches(attributes)
   #
   # @return [nil]
   def unsubscribe(uuid)
-    post_json("#{endpoint}/unsubscribe/#{uuid}")
+    post_json("#{endpoint}/unsubscribe/#{uri_encode(uuid)}")
```
cbaines, line 104: I'm not sure any characters in UUIDs need encoding?

lib/gds_api/email_alert_api.rb
```diff
@@ -110,7 +110,7 @@ def unsubscribe(uuid)
   #
   # @return [nil]
   def unsubscribe_subscriber(id)
-    delete_json("#{endpoint}/subscribers/#{id}")
+    delete_json("#{endpoint}/subscribers/#{uri_encode(id)}")
```
cbaines, line 113: From the database, it looks like this will always be a number, which won't need encoding?

kevindew:
Thanks for the review @cbaines. You're right that both uuid and integer ids shouldn't need uri-encoding. The same situation also applies to slugs where, if they're formatted as per email-alert-api rules they also shouldn't need uri-encoding. 

The problem lies in that none of these apply any pre-condition that the type is a particular format, nor is there convention for applications to apply pre-conditions to input before injecting it into gds-api-adapters. So bad input can equal invalid uris. Doing the uri encoding means that even bad input can form a request and the service can reject with a 404 or 400 if necessary.

It doesn't feel the worlds nicest practice but it is with precedence. Whereas setting up a pre-condition assertion before making a request would be a brand new approach to either GDS API or a client application.

cbaines:
> Thanks for the review @cbaines. You're right that both uuid and integer ids shouldn't need uri-encoding. The same situation also applies to slugs where, if they're formatted as per email-alert-api rules they also shouldn't need uri-encoding.
> 
> The problem lies in that none of these apply any pre-condition that the type is a particular format, nor is there convention for applications to apply pre-conditions to input before injecting it into gds-api-adapters. So bad input can equal invalid uris. Doing the uri encoding means that even bad input can form a request and the service can reject with a 404 or 400 if necessary.

Right, I should have probably put more effort in to reading your commit message.

Adding route constraints in to apps or additional form validation is potentially useful as well, but I understand the motivation for some "defence in depth".

> It doesn't feel the worlds nicest practice but it is with precedence. Whereas setting up a pre-condition assertion before making a request would be a brand new approach to either GDS API or a client application.

Yeah, and renaming GDS API Adapters variables `uuid` -> `hopefully_a_uuid` probably isn't helpful either...

kevindew:
Thanks @cbaines 👍

PR #1003, opened 2020-09-17
Upgrade to Ruby 2.7.1

cdccollins:
These are the changes necessary for Ruby 2.7.1

I've left the Ruby version as 2.6.6 since there is a script that updates all the apps' Ruby versions at once.


Rhian:
This change looks good, but we should probably also update the Changelog.md as well.

Rhian (approved):
Changes look good to me.

PR #1008, opened 2020-11-25
Add helper for creating lists in Email Alert API

benthorner:
https://trello.com/c/mk5M2aKD/619-general-design-and-stylistic-updates-to-signup-workflow

Currently we stub the adapter in Email Alert Frontend [1], so we're
vulnerable to it changing in future. Note that we can't reuse the
"stub_email_alert_api_creates_subscriber_list" test helper for this
(and assert the requet was made) since the attributes passed to the
helper can be specific to the response (although many will also be
in the request), and we've lost our chance to have a more flexible
"(attributes, returned_attributes)" interface for the helper.

[1]: https://github.com/alphagov/email-alert-frontend/blob/d00c3784fe5b1f7f937b593b8a84082849f879d8/spec/models/content_item_subscriber_list_spec.rb#L17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-# Unreleased
+# 67.1.0
 
+* Add assert_email_alert_api_creates_subscriber_list test helper
 * Add double splat operator to last arguments that are keyword parameters
 * Accept params in `stub_publishing_api_does_not_have_item`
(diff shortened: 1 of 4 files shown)


kevindew (approved):

benthorner:
Closing as the new helper wasn't required in practice.

PR #1011, opened 2020-12-09
Support param to skip subscribe confirmation email

benthorner:
https://trello.com/c/gbYVX8lK/690-send-users-a-link-to-their-results-as-part-of-subscribing-to-the-checker

Relates to: https://github.com/alphagov/email-alert-api/pull/1503

This is required to prevent sending two emails with conflicting content /
links when a user subscribes via a GOV.UK Account, which sends its own
confirmation email. In order to keep the test stubs simple, I've switched
to using "hash_including" for stubs where the parameter is (for now) not
relevant.

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -116,12 +116,13 @@ class GdsApi::EmailAlertApi < GdsApi::Base
   # Subscribe
   #
   # @return [Hash] subscription_id
-  def subscribe(subscriber_list_id:, address:, frequency: "immediately")
+  def subscribe(subscriber_list_id:, address:, frequency: "immediately", skip_confirmation_email: false)
     post_json(
       "#{endpoint}/subscriptions",
       subscriber_list_id: subscriber_list_id,
       address: address,
       frequency: frequency,
+      skip_confirmation_email: skip_confirmation_email,
     )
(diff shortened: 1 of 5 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -116,12 +116,13 @@ def unsubscribe_subscriber(id)
   # Subscribe
   #
   # @return [Hash] subscription_id
-  def subscribe(subscriber_list_id:, address:, frequency: "immediately")
+  def subscribe(subscriber_list_id:, address:, frequency: "immediately", skip_confirmation_email: false)
```
sihugh, line 119: Just to be clear, this will send the confirmation email by default? (Presumably so the accounts stuff keeps on working)

sihugh (approved):
All sounds sensible

lib/gds_api/email_alert_api.rb
benthorner: The email system should send an email by default, but this allows accounts to turn it off because they send their own confirmation email as part of their setup/subscribe process. Does that make sense?

sihugh: Ah, right - I had it backwards. Thanks for the clarification 👍

PR #1012, opened 2020-12-09
Allow locales for special routes

sihugh:
Locale seems to be the only field that we can't override (apart from details).

We need to publish a series of related routes that link to translations of each other. Locales are the way to do this, so let's allow that please.


hannako (approved):
👍

PR #1015, opened 2020-12-11
Update subscribe test helper to match new API

benthorner:
https://trello.com/c/hdOrNhuC/669-sign-in-after-subscribing

This makes a couple of breaking changes:

- Change the stub response for the test helper
- Make the interface to the test helper more flexible

Although these are breaking changes, they're only to the
test helper, so doesn't seem worthy of a major version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unversioned
+
+* BREAKING: update email test helper to match new API behaviour
+
 # 67.2.1
 
 * Fix bug in special route publishing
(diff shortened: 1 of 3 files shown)


sihugh (changes requested):
We should really be [versioning in a separate PR](https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html#versioning) according to the docs. We could then package up both changes in one release and save dependabot a bit of work?
Also, I know it's only a test helper, but it looks a bit weird to say we're following SemVer and then have `BREAKING` in a minor update? It's only a number, and ruby gems doesn't care.

sihugh (approved):
Sorry about the impending merge conflict

PR #1018, opened 2020-12-21
Add country_name to Mapit API response and update Local Links Manager test helper stubs

gclssvglx:
## What

The `country_name` - which is available from [Mapit](https://github.com/alphagov/mapit) - is not present in the current Mapit API response.

## Why

Add `country_name` to the Mapit API response so that we can provide a better user experience for the Test and Trace Payment service in situations where the service is unavailable from the user's given postcode.

[Trello](https://trello.com/c/QSY1CoAB/643-test-and-trace-nation-validation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 68.2.0
+
+* Update the local links manager adapter stubs to include country_name - defaults to `England`
+* Add country_name to the Mapit `location_for_postcode` API response
+
 # 68.1.0
 
 * Add a `GdsApi::HTTPBadRequest` exception
--- a/lib/gds_api/mapit.rb
+++ b/lib/gds_api/mapit.rb
@@ -37,6 +37,10 @@ class GdsApi::Mapit < GdsApi::Base
     def postcode
       @response["postcode"]
     end
+
+    def country_name
+      @response["country_name"]
+    end
   end
 
 private
--- a/lib/gds_api/test_helpers/local_links_manager.rb
+++ b/lib/gds_api/test_helpers/local_links_manager.rb
@@ -5,13 +5,14 @@ module GdsApi
     module LocalLinksManager
       LOCAL_LINKS_MANAGER_ENDPOINT = Plek.current.find("local-links-manager")
 
-      def stub_local_links_manager_has_a_link(authority_slug:, lgsl:, lgil:, url:)
+      def stub_local_links_manager_has_a_link(authority_slug:, lgsl:, lgil:, url:, country_name: "England")
         response = {
           "local_authority" => {
             "name" => authority_slug.capitalize,
             "snac" => "00AG",
             "tier" => "unitary",
             "homepage_url" => "http://#{authority_slug}.example.com",
+            "country_name" => country_name,
           },
           "local_interaction" => {
             "lgsl_code" => lgsl,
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-# Unreleased
+# 68.0.1
```
theseanything, line 1: This should probably be a minor bump as this add functionality, rather than just a bug fix.
```suggestion
# 68.1.0
```

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
...
 
+* Add country_name to the local links manager adatper
```
theseanything, line 3: spelling:
```suggestion
* Add country_name to the local links manager adapter
```

lib/gds_api/mapit.rb
```diff
@@ -37,6 +37,10 @@ def areas
     def postcode
       @response["postcode"]
     end
+
+    def country_name
+      @response["country_name"]
+    end
```
theseanything, line 43: Is this meant to be in local_links_manager.rb not mapit.rb?

lib/gds_api/test_helpers/local_links_manager.rb
```diff
@@ -5,13 +5,14 @@ module TestHelpers
     module LocalLinksManager
       LOCAL_LINKS_MANAGER_ENDPOINT = Plek.current.find("local-links-manager")
 
-      def stub_local_links_manager_has_a_link(authority_slug:, lgsl:, lgil:, url:)
+      def stub_local_links_manager_has_a_link(authority_slug:, lgsl:, lgil:, url:, country_name:)
```
theseanything, line 8: As these are helpers and used in other codebases I'd add a default value for country name. Otherwise it'll be a breaking change.
```suggestion
      def stub_local_links_manager_has_a_link(authority_slug:, lgsl:, lgil:, url:, country_name: "Wales")
```

theseanything (changes requested):

CHANGELOG.md
gclssvglx: Sorry, fixed

CHANGELOG.md
gclssvglx: Corrected

lib/gds_api/mapit.rb
gclssvglx: Thanks Sean - interesting question.  No, I don't believe so.  An instance of `Location` is created with the response from the `/postcode/` API call in `location_for_postcode` within this file.  That is [called in frontend](https://github.com/alphagov/frontend/blob/7572a3ac88d7f86969bb9ef48ed0d53e0af2e148/app/controllers/local_transaction_controller.rb#L62) in the `LocalTransactionController` which is the controller we need to update in order to surface the 'unavailable in country' error.

lib/gds_api/test_helpers/local_links_manager.rb
gclssvglx: Good spot - thanks Sean - added `England` as the default.

lib/gds_api/mapit.rb
theseanything: From chatting in person - we've recommended creating two separate commits, one for the test stubs for local links manager and the another for the Mapit response. Remove reference about adding country name to local links manager adapter - as these endpoint return the whole JSON blob anyway.

theseanything (approved):

PR #1020, opened 2021-01-04
Remove redundant adapter to send ad-hoc email

benthorner:
This is unused [1].

[1]: https://github.com/alphagov/email-alert-api/issues/1531


1pretz1 (approved):
Conflict concerning the CHANGELOG but other than that, looks good! 👍

PR #1023, opened 2021-01-13
[DO NOT MERGE] Switch email list adapter to just use "create"

benthorner:
https://trello.com/c/dgdLhy2H/777-update-transition-to-brexit-in-email-notifications-for-transition-taxon

Depends on: https://github.com/alphagov/email-alert-api/pull/1569

Previously calling "find_or_create_subscriber_list" would result in
up to 2 API calls. Since we've changed the "create" endpoint to also
return an existing list [1], we can simplify the adapter to just use
that, with the added benefit of keeping titles/URLs in-sync.

[1]: https://github.com/alphagov/email-alert-api/pull/1569

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* BREAKING: remove redundant test helpers for finding subscriber lists
 * BREAKING: Switch "find_or_create_subscriber_list" to only call "create" endpoint
 
 # 68.2.0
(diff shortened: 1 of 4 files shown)


sihugh (approved):
Looks good to me, when the time is right :-)

benthorner:
@sihugh I've added another commit to this, to remove a couple of test helpers. This more clearly makes this a breaking change, so I've specified as much in the CHANGELOG.

Would you like to take another look?

PR #1024, opened 2021-01-13
Contract tests with Collections's Organisations API

hannako:
## What
Collections is the home of the [Organisations api](https://github.com/alphagov/collections/blob/master/docs/api.md#collections-api). For collections to [meet the standards](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-128-continuous-deployment.md#check-api-contract-tests-pass-ie-adapters-work-with-their-apis) required for continuously deployment, we need contract tests between collections and its api consumer, gds-api-adaptors.
This PR is the first step, allowing the contract tests to be run locally against https://github.com/alphagov/collections/pull/2230.

## Steps to run tests
1. From this branch, `bundle exec rake TEST=test/organisations_api_test.rb` creates a pactfile which defines the pacts with collections. 
2. cd collections and checkout https://github.com/alphagov/collections/pull/2230
3. `bundle exec rake pact:verify`

See https://github.com/alphagov/collections/pull/2230 for next steps. 

-- 
[Trello](https://trello.com/c/Qpd7jgOq/628-add-contract-testing-to-the-collections-app)

### DIFF
--- a/test/organisations_api_test.rb
+++ b/test/organisations_api_test.rb
@@ -108,4 +108,31 @@ describe GdsApi::Organisations do
       assert_equal 40, response.with_subsequent_pages.count
     end
   end
+
+  describe "fetching an organisation by slug" do
+    let(:hmrc) { "hm-revenue-customs" }
+
+    before do
+      organisation_api
+        .given("the organisation hmrc exists")
+        .upon_receiving("a request for hm-revenue-customs")
+        .with(
+          method: :get,
+          path: "/api/organisations/#{hmrc}",
+          headers: GdsApi::JsonClient.default_request_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          body: organisation(slug: hmrc),
+        )
+    end
+
+    it "responds with 200 and the organisation" do
+      response = api_client.organisation(hmrc)
+
+      id = "www.gov.uk/api/organisations/#{hmrc}"
+      assert_equal 200, response.code
+      assert_equal id, response["id"]
+    end
+  end
 end
--- a/test/organisations_api_test.rb
+++ b/test/organisations_api_test.rb
@@ -135,4 +135,28 @@ describe GdsApi::Organisations do
       assert_equal id, response["id"]
     end
   end
+
+  describe "an organisation doesn't exist for a given slug" do
+    before do
+      organisation_api
+        .given("no organisation exists")
+        .upon_receiving("a request for a non-existant organisation")
+        .with(
+          method: :get,
+          path: "/api/organisations/department-for-making-life-better",
+          headers: GdsApi::JsonClient.default_request_headers,
+        )
+        .will_respond_with(
+          status: 404,
+          body: "404 error",
+        )
+    end
+
+    it "returns a 404 error code" do
+      error = assert_raises(GdsApi::HTTPNotFound) do
+        api_client.organisation("department-for-making-life-better")
+      end
+      assert_equal 404, error.code
+    end
+  end
 end
--- a/test/organisations_api_test.rb
+++ b/test/organisations_api_test.rb
@@ -5,5 +5,59 @@ require "gds_api/test_helpers/organisations"
 describe GdsApi::Organisations do
   include GdsApi::TestHelpers::Organisations
   include PactTest
-  
+
+  def base_api_url
+    Plek.new.website_root
+  end
+
+  def api_client
+    @api_client ||= GdsApi::Organisations.new(organisation_api_host)
+  end
+
+  def organisation(slug: "test-department")
+    {
+      "id" => Pact.like("www.gov.uk/api/organisations/#{slug}"),
+      "title" => Pact.like("Test Department"),
+      "updated_at" => Pact.like("2019-05-15T12:12:17.000+01:00"),
+      "web_url" => Pact.like("www.gov.uk/government/organisations/#{slug}"),
+      "details" => {
+        "slug" => Pact.like(slug),
+        "content_id" => Pact.like("b854f170-53c8-4098-bf77-e8ef42f93107"),
+      },
+      "analytics_identifier" => Pact.like("OT1276"),
+      "child_organisations" => [],
+      "superseded_organisations" => [],
+    }
+  end
+
+  describe "fetching list of organisations" do
+    before do
+      organisation_api
+        .given("there is a list of organisations")
+        .upon_receiving("a request for the organisation list")
+        .with(
+          method: :get,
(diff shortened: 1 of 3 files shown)


test/organisations_api_test.rb
```diff
@@ -4,54 +4,216 @@
...
-    it "should raise error if endpoint 404s" do
-      stub_request(:get, "#{@base_api_url}/api/organisations").to_return(status: 404)
-      assert_raises GdsApi::HTTPNotFound do
-        @api.organisations
+    it "should handle pagination" do
+      def json(param = nil)
+        GdsApi::JsonClient.new.get_raw("#{organisation_api_host}/api/organisations#{param}")
       end
+
+      # Stub the response to get_list called by GdsApi::Organisations.organisations, because the Pact tests
```
benthorner, line 120: I had a look at this, and unfortunately couldn't get to the bottom of it. For some reason in Collections `request.host` is `example.org`. It doesn't matter what the host is, and if we could force it to be `localhost:4000`, then we could remove these stubs. But I couldn't figure out where it's coming from.

(Would be great if you're able to figure it out, as I expect it'll cause pain in future tests.) I think it would be helpful to give a bit more info in this comment, as it's quite a subtle problem. I think the main points are:

- Pagination relies on "next" links in the response headers / body.
- The link includes the host, which we could set to the api host here.
- But on the provider side, it gets overridden to `example.org`.
- We therefore have to cope with `example.org` here so it works for both 😠.

benthorner (commented):
Our next example of Pact on GOV.UK is on the way 🏅. It's pretty much there on this side, with just a few more major comments to look at first.

test/organisations_api_test.rb
```diff
@@ -4,55 +4,6 @@
...
-
-    it "should raise for a non-existent organisation" do
-      stub_organisations_api_does_not_have_organisation("non-existent")
-
-      assert_raises(GdsApi::HTTPNotFound) do
-        @api.organisation("non-existent")
-      end
-    end
-  end
+  include PactTest
```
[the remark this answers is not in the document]
hannako: Ah I did that on purpose to make the intent of moving the organisations_api_test to a new folder obvious. But if it's confusing I will amend.

test/organisations_api_test.rb
```diff
@@ -71,4 +71,68 @@ def organisation_matchers(slug)
...
+      {
+        method: :get,
+        path: "/api/organisations",
+        headers: GdsApi::JsonClient.default_request_headers,
+      }
+    end
+    let(:body) do
+      {
+        results: Pact.each_like({}, min: 20),
+        page_size: 20,
```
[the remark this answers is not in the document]
hannako: The page size is [hardcoded](https://github.com/alphagov/collections/blob/91d3b6478af69145e6e729fdabd35654ee1436a4/app/controllers/organisations_api_controller.rb#L31) to 20, so I don't think there's much value in doing that.

test/organisations_api_test.rb
```diff
@@ -5,5 +5,70 @@
 describe GdsApi::Organisations do
   include GdsApi::TestHelpers::Organisations
   include PactTest
-  
+
+  def base_api_url
+    Plek.new.website_root
+  end
+
+  def mocked_api
```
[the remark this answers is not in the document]
hannako: Sure 👍

test/organisations_api_test.rb
```diff
@@ -5,5 +5,70 @@
...
+    GdsApi::Organisations.new(organisation_api_host)
+  end
+
+  def organisation_matchers(slug)
+    title = slug.split("-").map(&:capitalize).join(" ")
```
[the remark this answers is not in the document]
hannako: Thanks - I did this to make the pact file less abstract. Ie use a Pact.like so the contract tests are not brittle, but generate a real value so the pactfile is a bit nicer to read and debug. But happy to remove if you think it's overkill.

test/organisations_api_test.rb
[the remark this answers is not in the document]
benthorner: Oh yes, I see now 👍.

test/organisations_api_test.rb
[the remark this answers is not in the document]
benthorner: I agree there's value in using `Pact.like` with some realistic data. But I do think it's overkill to start making it dynamic, when it doesn't actually make a difference - that could itself be a source of confusion. Hopefully a compromise is just to have things like `Pact.like("Any old title")` and `Pact.like("any-old-slug")`, etc.

test/organisations_api_test.rb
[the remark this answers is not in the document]
benthorner: Oh, that's a shame 😢. Thanks for digging into it.

test/organisations_api_test.rb
hannako: Ah ha! Figured it out. See [here](https://github.com/rails/rails/blob/291a3d2ef29a3842d1156ada7526f4ee60dd2b59/actionpack/lib/action_controller/renderer.rb#L39) and [here](https://stackoverflow.com/questions/598933/how-do-i-change-the-default-www-example-com-domain-for-testing-in-rails).

Will remove the stubs and amend in collections eg
```
RSpec.configure do |config|
  config.before(:each, type: :request) do
    host! "localhost:4000"
  end
end
```

hannako: I've created a proxy so we can overwrite the default request.host in collections 

see: https://github.com/alphagov/collections/pull/2230/commits/946e2d3592ab2da17011064ee21bf14f7edc1db4

test/test_helpers/pact_helper.rb
```diff
@@ -1,13 +1,26 @@
 PUBLISHING_API_PORT = 3001
+ORGANISATION_API_PORT = 4000
```
sihugh, line 2: Is this just arbitrary?

hannako: I couldn't find any documentation on it, so yes I just picked 4000.

sihugh (commented):
This is great - just reckon we should pick a tidy number :-)

test/test_helpers/pact_helper.rb
hannako: Yep will do.

hannako: https://github.com/alphagov/collections/pull/2278

hannako: :please: :thankyou:

sihugh (approved):

PR #1032, opened 2021-02-12
Configure Jenkins to run pact tests with Collections

hannako:
## What

Following pact tests defined in #1024, run the pact tests as part of the deployment pipeline.

This PR is blocked by https://github.com/alphagov/collections/pull/2281 which defines the pact:verify:branch rake task that [this PR configures Jenkins to run](https://github.com/alphagov/gds-api-adapters/blob/5cd13fa8d0d170e83bdd8b1ddd35a6760a7e8003/Jenkinsfile#L63).

Related PR's in collections:
- https://github.com/alphagov/collections/pull/2230
- https://github.com/alphagov/collections/pull/2281

[Trello](https://trello.com/c/Qpd7jgOq/628-add-contract-testing-to-the-collections-app)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,11 @@ node("postgresql-9.6") {
         name: 'PUBLISHING_API_BRANCH',
         defaultValue: 'master',
         description: 'Branch of publishing-api to run pacts against'
+      ),
+      stringParam(
+        name: 'COLLECTIONS_BRANCH',
+        defaultValue: 'master',
+        description: 'Branch of collections to run pacts against'
       )
     ],
     afterTest: {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -49,3 +55,13 @@ def runPublishingApiPactTests(govuk) {
     }
   }
 }
+def runCollectionsOrganisationApiPactTests(govuk){
```
thomasleese, line 58:
```suggestion
def runCollectionsPactTests(govuk) {
```

I realise this is true at the moment, but I guess it helps to keep it generic for if we were to add pact tests for other parts of Collections.

thomasleese, line 58: I think this is for a later PR, but this and `runPublishingApiPactTests` are very similar so I wonder if there's some refactoring we could do here. I suspect we'll learn this as we add pact tests to more of our apps though, most likely we'll end up putting something in [govuk-jenkinslib](https://github.com/alphagov/govuk-jenkinslib).

hannako:
Thanks @thomasleese I made those small amends, and will make a cards on our board for the refactors you've suggested on this PR, and the equivalent in collections.

thomasleese (approved):

PR #1034, opened 2021-02-24
Update rubocop-govuk to 4.0.0.pre.1 

kevindew:
Trello: https://trello.com/c/pqXbKZiC/215-test-rubocop-govuk-400pre1-prior-to-400-release

This bumps to the 4.0.0 pre release of rubocop-govuk which is part of [trialling the new rubocop-govuk release](https://github.com/alphagov/rubocop-govuk/issues/129). The gem pin to 4.0.0.pre.1 is intended to be removed once the final release of rubocop-govuk 4.0.0 is out.

Applying this revealed we were supporting a rather old Ruby version (2.4) so I've bumped our support to 2.6 and marked this as a breaking change.


lib/gds_api/test_helpers/content_store.rb
```diff
@@ -7,7 +7,7 @@ module TestHelpers
     module ContentStore
       include ContentItemHelpers
 
-      def content_store_endpoint(draft = false)
+      def content_store_endpoint(draft = false) # rubocop:disable Style/OptionalBooleanParameter
```
leenagupte: So the linter wants this changed to `def content_store_endpoint(draft: false)`?

test/cache_control_test.rb
```diff
@@ -2,44 +2,42 @@
 require "gds_api/response"
 
 describe GdsApi::Response::CacheControl do
-  CacheControl = GdsApi::Response::CacheControl
-
   it "takes no args and initializes with an empty set of values" do
-    cache_control = CacheControl.new
+    cache_control = GdsApi::Response::CacheControl.new
```
leenagupte, line 6: Does `described_class` only work with RSpec?

test/json_client_test.rb
```diff
@@ -395,7 +395,7 @@ def test_client_can_set_custom_headers_on_gets
     )
 
     assert_requested(:get, %r{/some.json}) do |request|
-      headers_with_uppercase_names = Hash[request.headers.collect { |key, value| [key.upcase, value] }]
+      headers_with_uppercase_names = request.headers.transform_keys(&:upcase)
```
leenagupte, line 398: Well this is much nicer!

leenagupte (commented):
This is looking good 🎉. I have a couple of inline questions.

lib/gds_api/test_helpers/content_store.rb
kevindew: Yup that's correct. On the basis that method calls like `content_store_endpoint(true)` don't really mean anything but `content_store_endpoint(draft: true)` does. 

I've fixed this as a breaking change in https://github.com/alphagov/gds-api-adapters/pull/1034/commits/4a7954c4ddcf92c481a44ab5c9f472dd55b26848 (though I don't think it's a breaking change that'll affect any apps as I can't find any apps - outside of this one - that use)

test/cache_control_test.rb
kevindew: Yes, sadly so: https://github.com/seattlerb/minitest/issues/526

leenagupte (approved):
Thanks for answering my questions. It all looks good to me! 👍 💯

kevindew:
Thanks @leenagupte 👍

PR #1037, opened 2021-03-03
Allow developer to specify PACT_URI for Frontend Pact tests

ChrisBAshton:
This is to simplify the Frontend codebase. See
https://github.com/alphagov/frontend/pull/2647

I've made this configurable field called `FRONTEND_PACT_URI`,
rather than `PACT_URI`, so that we can provide different `PACT_URI`
values for each provider (otherwise if they all implement the
`PACT_URI` approach, all but one would fail as a result of
pointing to the wrong pactfile).


ChrisBAshton:
Superseded by https://github.com/alphagov/frontend/pull/2648.

PR #1040, opened 2021-03-12
Add account-api attribute endpoints

barrucadu:
See https://github.com/alphagov/account-api/pull/8

---

[Trello card](https://trello.com/c/uRMyiEHZ/654-move-oauth-attribute-logic-from-the-transition-checker-to-the-new-app)

### DIFF
--- a/lib/gds_api/account_api.rb
+++ b/lib/gds_api/account_api.rb
@@ -57,6 +57,27 @@ class GdsApi::AccountApi < GdsApi::Base
     post_json("#{endpoint}/api/transition-checker-email-subscription", { slug: slug }, auth_headers(govuk_account_session))
   end
 
+  # Look up the values of a user's attributes
+  #
+  # @param [String] attributes Names of the attributes to check
+  # @param [String] govuk_account_session Value of the session header
+  #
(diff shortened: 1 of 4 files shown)


lib/gds_api/account_api.rb
```diff
@@ -57,6 +57,27 @@ def set_email_subscription(govuk_account_session:, slug:)
...
+  end
+
+  # Create or update attributes for a user
+  #
+  # @param [String] attributes Hash of new attribute values
+  # @param [String] govuk_account_session Value of the session header
+  #
+  # @return [Hash] A new session header
+  def set_attributes(attributes:, govuk_account_session:)
+    patch_json("#{endpoint}/api/attributes", { attributes: attributes.transform_values(&:to_json) }, auth_headers(govuk_account_session))
```
huwd, line 78: This is the `.to_json` mentioned here right?
https://github.com/alphagov/account-api/pull/8

huwd (approved):

lib/gds_api/account_api.rb
barrucadu: It is, yes

PR #1043, opened 2021-03-17
Write Pact consumer tests for account-api 

barrucadu:
This is step 1 of the [Writing Pact tests](https://docs.publishing.service.gov.uk/manual/pact-broker.html#writing-pact-tests) docs, see also https://github.com/alphagov/account-api/pull/13.

---

Pact consumer tests have a few different parts:

1. They define request/response templates (much like webmock, but with
a different DSL); this is the `before` blocks

2. They run the tests and complain if:
   - any unexpected requests are made
   - any defined requests aren't made

The templates are saved to a JSON file, which is used by the provider
tests (in this case, the account-api tests) to validate that the
provider returns matching outputs for the same inputs.

The "given" part of the set-up corresponds to a "provider state",
which is a named `before` block, run as part of the provider tests.

---

[Trello card](https://trello.com/c/OcRffYwv/674-set-up-pact-tests-for-our-account-api-app)

### DIFF
--- a/test/account_api/account_api_test.rb
+++ b/test/account_api/account_api_test.rb
@@ -5,8 +5,7 @@ require "gds_api/test_helpers/account_api"
 describe GdsApi::AccountApi do
   include GdsApi::TestHelpers::AccountApi
 
-  let(:base_url)      { Plek.find("account-api") }
-  let(:api_client)    { GdsApi::AccountApi.new(base_url) }
+  let(:api_client) { GdsApi::AccountApi.new(Plek.find("account-api")) }
 
   it "gets a sign in URL" do
     stub_account_api_get_sign_in_url(auth_uri: "https://www.example.com")
--- a/test/account_api_test.rb
+++ b/test/account_api_test.rb
@@ -1,0 +1,278 @@
+require "test_helper"
+require "gds_api/account_api"
+
+describe GdsApi::AccountApi do
+  include PactTest
+
+  let(:api_client) { GdsApi::AccountApi.new(account_api_host) }
+
+  let(:authenticated_headers) { { "GOVUK-Account-Session" => govuk_account_session } }
+  let(:govuk_account_session) { "logged-in-user-session" }
+
+  describe "getting a sign-in URL" do
+    let(:redirect_path) { nil }
+
+    before do
+      account_api
+        .upon_receiving("a sign-in request")
+        .with(
+          method: :get,
+          path: "/api/oauth2/sign-in",
+          headers: GdsApi::JsonClient.default_request_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
+          body: {
+            auth_uri: Pact.like("http://authentication-provider/some/oauth/url"),
+            state: Pact.like("value-to-use-for-csrf-prevention"),
+          },
+        )
+    end
+
+    it "responds with 200 OK, an authentication URI, and a state for CSRF protection" do
+      response = api_client.get_sign_in_url
+      assert response["auth_uri"].present?
+      assert response["state"].present?
+      assert_equal 200, response.code
+    end
+  end
+
+  describe "validating an OAuth response" do
+    before do
+      account_api
+        .given("there is a valid OAuth response")
+        .upon_receiving("a validation request")
+        .with(
+          method: :post,
+          path: "/api/oauth2/callback",
+          body: {
+            code: "code",
+            state: "state",
+          },
+          headers: GdsApi::JsonClient.default_request_with_json_body_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
+          body: {
+            govuk_account_session: Pact.like("user-session-id"),
+          },
+        )
+    end
+
+    it "responds with 200 OK and a govuk_account_session" do
+      response = api_client.validate_auth_response(code: "code", state: "state")
+      assert response["govuk_account_session"].present?
+      assert response["redirect_path"].nil?
+      assert_equal 200, response.code
+    end
+  end
+
+  describe "validating an OAuth response with a redirect path" do
+    let(:redirect_path) { "/some-arbitrary-path" }
+
+    before do
+      account_api
+        .given("there is a valid OAuth response, with the redirect path '/some-arbitrary-path'")
+        .upon_receiving("a validation request")
+        .with(
+          method: :post,
+          path: "/api/oauth2/callback",
+          body: {
+            code: "code",
+            state: "state",
+          },
+          headers: GdsApi::JsonClient.default_request_with_json_body_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
+          body: {
+            govuk_account_session: Pact.like("user-session-id"),
+            redirect_path: Pact.like(redirect_path),
+          },
+        )
+    end
+
+    it "responds with a redirect_path" do
+      response = api_client.validate_auth_response(code: "code", state: "state")
+      assert_equal redirect_path, response["redirect_path"]
+    end
+  end
+
+  describe "creating a registration state" do
+    let(:attributes) { { foo: "bar" } }
+
+    before do
+      account_api
+        .upon_receiving("a create-state request")
+        .with(
+          method: :post,
+          path: "/api/oauth2/state",
+          body: { attributes: attributes },
+          headers: GdsApi::JsonClient.default_request_with_json_body_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
+          body: {
+            state_id: Pact.like("reference-to-pass-to-get_sign_in_url"),
+          },
+        )
+    end
+
+    it "responds with 200 OK and a state_id" do
+      response = api_client.create_registration_state(attributes: attributes)
+      assert response["state_id"].present?
+      assert_equal 200, response.code
+    end
+  end
+
+  describe "checking for a transition checker email subscription" do
+    describe "the user is logged in" do
+      let(:given) { "there is a valid user session" }
+      let(:has_subscription) { false }
+
+      before do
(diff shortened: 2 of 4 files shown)


test/account_api_test.rb
```diff
@@ -0,0 +1,407 @@
...
+          method: :get,
+          path: "/api/oauth2/sign-in",
+          headers: GdsApi::JsonClient.default_request_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
+          body: {
+            auth_uri: Pact.like("string"),
+            state: Pact.like("string"),
```
ChrisBAshton, line 29: I think [Pact matching convention](https://docs.pact.io/getting_started/matching/) is to put an example value here, rather than describing the type. Pact's own matching will infer that the example value is a string.

test/account_api_test.rb
```diff
@@ -0,0 +1,407 @@
...
+            status: 200,
+            headers: { "Content-Type" => "application/json; charset=utf-8" },
+            body: {
+              govuk_account_session: Pact.like("string"),
+              has_subscription: has_subscription,
+            },
+          )
+      end
+
+      it "responds with 200 OK, a new govuk_account_session, and says that the subscription does not exists" do
```
ChrisBAshton, line 190:
```suggestion
      it "responds with 200 OK, a new govuk_account_session, and says that the subscription does not exist" do
```

test/account_api_test.rb
```diff
@@ -0,0 +1,407 @@
...
+        assert response["govuk_account_session"].present?
+        assert_equal false, response["has_subscription"]
+        assert_equal 200, response.code
+      end
+
+      describe "a subscription does not exist" do
+        let(:given) { "there is a valid user session, with a transition checker email subscription" }
+        let(:has_subscription) { true }
+
+        it "says that the subscription does not exist" do
```
ChrisBAshton, line 201: Is that what's being asserted below? 🤔

ChrisBAshton (commented):
Approach LGTM 👍 Just a couple of minor comments below.

test/account_api_test.rb
```diff
@@ -1,81 +1,407 @@
...
+        .will_respond_with(
+          status: 422,
+          headers: { "Content-Type" => "application/json; charset=UTF-8" },
+        )
+    end
+
+    it "responds with 422 Unprocessable Entity" do
+      assert_raises GdsApi::HTTPUnprocessableEntity do
+        api_client.get_sign_in_url(redirect_path: redirect_path)
+      end
```
kevindew, line 64: A quick observation of mine is that we probably don't need pact tests for negative scenarios when this app doesn't have any code to handle those scenarios. It doesn't seem that app needs a contract that Account API will act this way in a particular scenario because it doesn't have a view on it.

(I imagine this is a carryover from Publishing API ones that may have a bunch of scenarios tested that may not actually add value).

The positive ones I've glanced at look good though. I've not got time for a full review but awesome to see you're working on these and seeking feedback 👍

test/account_api_test.rb
barrucadu: Whoops, I flipped the logic of this one and didn't update the name

test/account_api_test.rb
```diff
@@ -1,81 +1,407 @@
...
-      api_client.validate_auth_response(code: "foo", state: "bar")
+    before do
+      account_api
+        .given(given)
+        .upon_receiving("a validation request")
+        .with(
+          method: :post,
+          path: "/api/oauth2/callback",
+          body: {
+            code: Pact.like("code"),
```
hannako, line 80: You could use `Pact.like(code: "code", state: "state")` as you have on line 121. Similar comment for line 153

barrucadu: I think I'll actually just change those to a hash.  I've just been reading https://docs.pact.io/consumer/, which says

> Typically, exact matching is most appropriate for Pact tests on the consumer side that are running at the unit test level. The same person is responsible for both the expectation and the actual request, so making sure that they match should be straightforward.

test/account_api_test.rb
barrucadu: Sure, I see where you're coming from. And reading [the pact docs](https://docs.pact.io/consumer/), it says you shouldn't use pact tests to ensure functional correctness of the provider.  So I'll remove the error case tests.

test/account_api_test.rb
```diff
@@ -1,81 +1,407 @@
...
+            state: Pact.like("state"),
+          },
+          headers: GdsApi::JsonClient.default_request_with_json_body_headers,
+        )
+        .will_respond_with(
+          status: 200,
+          headers: { "Content-Type" => "application/json; charset=utf-8" },
+          body: {
+            govuk_account_session: Pact.like("user-session-id"),
+            redirect_path: Pact.like(redirect_path),
```
hannako, line 64: I found it a bit confusing that `redirect_path` is nil. What does Pact.like(nil) allow to pass?

barrucadu: Ah, this can go away now.  It was useful when I had the test for the error case, so I didn't have to write two almost-identical before blocks, but now that's gone.

test/account_api/account_api_test.rb
```diff
@@ -5,8 +5,7 @@
 describe GdsApi::AccountApi do
   include GdsApi::TestHelpers::AccountApi
 
-  let(:base_url)      { Plek.find("account-api") }
-  let(:api_client)    { GdsApi::AccountApi.new(base_url) }
+  let(:api_client) { GdsApi::AccountApi.new(Plek.find("account-api")) }
```
huwd, line 8: Do we need the Plek line here?
I've been l[ooking through the GDS API](https://github.com/alphagov/gds-api-adapters/blob/603fb33f3894559029728eb9143d1644a7cd858e/lib/gds_api.rb#L32-L35)

Looked a bit like we could call `GdsApi.account_api` which also fetches stuff out of plek?

barrucadu: Yeah, I did this to be consistent with the other tests

test/account_api_test.rb
```diff
@@ -0,0 +1,278 @@
+require "test_helper"
+require "gds_api/account_api"
+
+describe GdsApi::AccountApi do
+  include PactTest
+
+  let(:api_client) { GdsApi::AccountApi.new(account_api_host) }
+
+  let(:authenticated_headers) { { "GOVUK-Account-Session" => govuk_account_session } }
```
huwd, line 9: Given the travails in testing these over in frontend, might this be better as a standard capitalised + HTTP_ prefixed name?

huwd:
```suggestion
  let(:authenticated_headers) { { "HTTP_GOVUK_ACCOUNT_SESSION" => govuk_account_session } }
```

test/account_api_test.rb
```diff
@@ -0,0 +1,278 @@
...
+  let(:govuk_account_session) { "logged-in-user-session" }
```
huwd, line 10: huh, I didn't know we could use let in minitest. I'd assumed that was an Rspec thing!
cool.

test/account_api_test.rb
```diff
@@ -0,0 +1,278 @@
...
+
+  describe "getting a sign-in URL" do
+    let(:redirect_path) { nil }
+
+    before do
+      account_api
+        .upon_receiving("a sign-in request")
```
huwd, line 17: These lines seem to just be human readable strings that don't repeat or match anywhere.
So what are they used for? Are these just Pact versions of `describe` blocks for test stack traces?

test/account_api_test.rb
```diff
@@ -0,0 +1,278 @@
...
+      let(:given) { "there is a valid user session" }
+      let(:has_subscription) { false }
+
+      before do
+        account_api
+          .given(given)
+          .upon_receiving("a has-subscription request")
+          .with(
+            method: :get,
+            path: "/api/transition-checker-email-subscription",
```
huwd, line 143: Now i see it alongside the others, this path in particular stands out like a sore thumb doesn't it.
So specific where the others are generic and flexible 🤔.

Makes me wonder about if we'll get a chance to tackle subscriptions integrations again and de-snowflake this.

test/test_helpers/pact_helper.rb
```diff
@@ -1,6 +1,7 @@
 PUBLISHING_API_PORT = 3001
 ORGANISATION_API_PORT = 3002
 BANK_HOLIDAYS_API_PORT = 3003
+ACCOUNT_API_PORT = 3004
```
huwd, line 4: Interesting that they all need to be on different ports 🤔 Why is that?
Is this as a precaution in case we ever run more than two of these in the same test block in future?

Or does it complain about a port clash even if we're not pact testing ACCOUNT_API and BANK_HOLIDAYS_API in the same request?

huwd (changes requested):

test/account_api_test.rb
barrucadu: I think so, yes.  The `given`s correspond to set-up blocks in account-api, but these `upon_receiving`s are descriptive.

test/account_api_test.rb
barrucadu: No, the request header is `GOVUK-Account-Session` - the header-name-mangling is done by Rails: see https://github.com/alphagov/frontend/pull/2668#issuecomment-802843544

test/account_api_test.rb
barrucadu: Either we'll decide we don't need this feature once the transition checker goes away, and so we can drop it; or we'll decide we want it for more things, and so we'll have to rework it.  But I don't think there's much point in reworking it to be more generic until we need it.

test/test_helpers/pact_helper.rb
barrucadu: Pact throws an error if multiple services use the same port.  We might be able to fix that by moving the `Pact.service_consumer`s into the test files, but I think that's out of scope of the PR (and may be more complicated if we have to explicitly unregister the `service_consumer`s too).

test/account_api_test.rb
huwd: ah ha, great - cheers for the link!

huwd (approved):

PR #1045, opened 2021-03-19
Remove explicit JSON encoding of account-api attribute values

barrucadu:
account-api previously required attribute values to be given as
strings of JSON, rather than just as part of a regular JSON PATCH
request.  So in practice rather than sending data like

    {
      "attributes": {
        "foo": {
          "bar": ["baz", "bat", "qux"]
        }
      }
    }

We would instead send

    {
      "attributes": {
        "foo": "{\"bar\": [\"baz\", \"bat\", \"qux\"]}"
      }
    }

The need for this has been removed:
https://github.com/alphagov/account-api/pull/17

So now gds-api-adapters can remove the extra encoding step.  When this
change has gone out, the compatibility code which makes both encodings
work can be removed from account-api, switching us fully over to the
new encoding.

---

[Trello card](https://trello.com/c/E5J5kzNP/679-remove-the-redundant-json-encoding-of-attribute-values-in-account-api)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * BREAKING: Removed support for the Performance Platform.
 * Add Pact tests for account-api
+* Remove explicit JSON encoding of attribute values now that account-api parses non-string values correctly
 
 # 70.0.0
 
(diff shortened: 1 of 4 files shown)


huwd (approved):
Nice, looks much cleaner

PR #1049, opened 2021-04-12
Set `PACT_CONSUMER_VERSION` ENV before running Pact tests

ChrisBAshton:
This PR is necessary to remove the unneeded `pact:verify:branch` Rake tasks from the Pact provider apps, as per below:

- https://github.com/alphagov/frontend/pull/2712
- https://github.com/alphagov/account-api/pull/40
- https://github.com/alphagov/collections/pull/2355
- https://github.com/alphagov/publishing-api/pull/1932

Here is a passing build that references all of the branches above: https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/simplify-pact-verify/8/console. Therefore we can be reasonably confident everything works nicely 🎉  

Once all of these are merged, the end result should be the same: Pact tests that are run on every PR for both provider/Consumer. This is just a simplified way of achieving that.

We're doing the same thing in the Publishing API / Content Store Pact tests: https://github.com/alphagov/publishing-api/pull/1934

Trello: https://trello.com/c/rnfUFP5o/2453-experiment-with-common-configuration-for-pact-tests-timebox-2-days

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,7 +41,7 @@ node("postgresql-9.6") {
         ]
       ]) {
         publishPacts(govuk, env.BRANCH_NAME == 'master')
-        govuk.setEnvar("GDS_API_ADAPTERS_PACT_VERSION", "branch-${env.BRANCH_NAME}")
+        govuk.setEnvar("PACT_CONSUMER_VERSION", "branch-${env.BRANCH_NAME}")
         runPublishingApiPactTests(govuk)
         runCollectionsPactTests(govuk)
         runFrontendPactTests(govuk)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,6 +7,7 @@ node("postgresql-9.6") {
   def pact_branch = (env.BRANCH_NAME == 'master' ? 'master' : "branch-${env.BRANCH_NAME}")
   govuk.setEnvar("PACT_TARGET_BRANCH", pact_branch)
   govuk.setEnvar("PACT_BROKER_BASE_URL", "https://pact-broker.cloudapps.digital")
+  govuk.setEnvar("PACT_CONSUMER_VERSION", "branch-${env.BRANCH_NAME}")
 
   govuk.buildProject(
     extraParameters: [
@@ -41,7 +42,6 @@ node("postgresql-9.6") {
         ]
       ]) {
         publishPacts(govuk, env.BRANCH_NAME == 'master')
-        govuk.setEnvar("PACT_CONSUMER_VERSION", "branch-${env.BRANCH_NAME}")
         runPublishingApiPactTests(govuk)
         runCollectionsPactTests(govuk)
         runFrontendPactTests(govuk)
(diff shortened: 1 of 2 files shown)


thomasleese (approved):

test/organisations_api_test.rb
```diff
@@ -63,8 +63,18 @@ def organisation(slug: "test-department")
...
+        /^<https?:\/\/(?:[^\/]+)\/api\/organisations\?page=2>; rel="next", <https?:\/\/(?:[^\/]+)\/api\/organisations\?page=1>; rel="self"$/,
+        %(<#{api_client_endpoint}?page=2>; rel="next", <#{api_client_endpoint}?page=1>; rel="self"),
+      )
+    end
+    let(:page_two_links) do
+      Pact.term(
+        /^<https?:\/\/(?:[^\/]+)\/api\/organisations\?page=1>; rel="previous", <https?:\/\/(?:[^\/]+)\/api\/organisations\?page=2>; rel="self"$/,
+        %(<#{api_client_endpoint}?page=1>; rel="previous", <#{api_client_endpoint}?page=2>; rel="self"),
+      )
+    end
```
kevindew, line 77: These look great, thanks for your digging into these. 

I have a minor syntax suggestion to make them a bit easier to read:

```
let(:host_agnostic_endpoint_regex) { %r{https?://(?:[^/]+)/api/organisations} }
let(:api_client_endpoint) { "#{organisation_api_host}/api/organisations" }

let(:page_one_links) do
  Pact.term(
    /\A<#{host_agnostic_endpoint_regex}\?page=2>; rel="next", <#{host_agnostic_endpoint_regex}\?page=1>; rel="self"\z/,
    %(<#{api_client_endpoint}?page=2>; rel="next", <#{api_client_endpoint}?page=1>; rel="self"),
  )
end
```

Do you think it'll be also easier to read if you we use the keyword arguments of generate and matcher as per the examples https://docs.pact.io/getting_started/matching/  ? - that might make it clearer what the two arguments mean (though generate is a rather opaque keyword).

ChrisBAshton: Good idea - I've amended as suggested, and used the keywords, which do I think improve readability slightly.

Jenkinsfile
```diff
@@ -41,6 +41,7 @@ node("postgresql-9.6") {
         ]
       ]) {
         publishPacts(govuk, env.BRANCH_NAME == 'master')
+        govuk.setEnvar("PACT_CONSUMER_VERSION", "branch-${env.BRANCH_NAME}")
```
kevindew, line 44: This might sit a little more coherently with the other pact env vars at the top of the file: https://github.com/alphagov/gds-api-adapters/blob/e54f8aa465e1c31e51f073ad719930aca4bf5080/Jenkinsfile#L8-L9

kevindew (approved):
Looks good, I've put a minor comment in and don't think you'll need a follow-up review if you action it (not particularly concerned if you don't action it either)

PR #1050, opened 2021-04-26
Combine all Pact service consumers

thomasleese:
We don't need them to be in separate `service_consumer` blocks as they're all part of the same "GDS API Adapters" service consumer.


ChrisBAshton (approved):
Nice 👍

PR #1053, opened 2021-04-29
Remove references to resource monitors

thomasleese:
These were added to Link Checker API but we didn't end up using them, and they were eventually removed in https://github.com/alphagov/link-checker-api/pull/140 so we don't need API adapters for them anymore.


kevindew (approved):
Woop - nice clean up 💯

PR #1055, opened 2021-04-29
Remove special case for master branch

thomasleese:
This ensures that the `master` branch is published to the Pact broker as `branch-master` as with all the other branches for consistency. This also avoids us needing to have this sort of logic in our other apps.

I believe this will also fix an issue where the `master` branch of GDS API Adapters has been failing since #1049 but rather than re-introduce the logic, I'd prefer `master` to not be treated special. ([Example failing build](https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/master/407/console))

I notice this was originally introduced in d12aeb5 but I don't fully understand the reasoning.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,7 +41,7 @@ node("postgresql-9.6") {
           passwordVariable: 'PACT_BROKER_PASSWORD'
         ]
       ]) {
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👏

PR #1059, opened 2021-05-04
DRY up the Pact tests

ChrisBAshton:
Throughout our Pact tests, we've duplicated our intentions, both
in the `will_respond_with` declaration AND in manual assertions
afterwards (e.g. `assert_equal(reponse_hash, content_id)`).

We should let the Pact gem do the heavy lifting here, and declare
our expected behaviour in just one place. It is up to Pact to fail
the build if the response is not as we expected.

The result should be easier-to-read tests, as we've removed some
nesting in cases where we had `before` blocks which served just
one `it` block. We made a decision in Content Publisher to move
setup code as close as possible to the assertions, which this
commit is enabling: https://docs.publishing.service.gov.uk/apps/content-publisher/testing-strategy.html

The build still fails if we introduce an incorrect expectation, e.g. https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/dry-pact-tests-test/1/console (from commit https://github.com/alphagov/gds-api-adapters/commit/4c1da4011b10a011d644c7594c80bb0b9b6e7b0a)

### DIFF
--- a/test/organisations_api_test.rb
+++ b/test/organisations_api_test.rb
@@ -26,35 +26,29 @@ describe GdsApi::Organisations do
     }
   end
 
-  describe "fetching list of organisations" do
-    before do
-      organisation_api
-        .given("there is a list of organisations")
-        .upon_receiving("a request for the organisation list")
-        .with(
-          method: :get,
-          path: "/api/organisations",
-          headers: GdsApi::JsonClient.default_request_headers,
-        )
-        .will_respond_with(
-          status: 200,
-          body: {
-            results: [
-              organisation,
-              organisation,
-            ],
-          },
-          headers: {
-            "Content-Type" => "application/json; charset=utf-8",
-          },
-        )
-    end
+  it "fetches a list of organisations" do
+    organisation_api
+      .given("there is a list of organisations")
+      .upon_receiving("a request for the organisation list")
+      .with(
+        method: :get,
+        path: "/api/organisations",
+        headers: GdsApi::JsonClient.default_request_headers,
+      )
+      .will_respond_with(
+        status: 200,
+        body: {
+          results: [
+            organisation,
+            organisation,
+          ],
+        },
+        headers: {
+          "Content-Type" => "application/json; charset=utf-8",
+        },
+      )
 
-    it "responds with 200 OK and a list of organisations" do
-      response = api_client.organisations
-      assert_equal 2, response["results"].count
-      assert_equal 200, response.code
-    end
+    api_client.organisations
   end
 
   describe "fetching a paginated list of organisations" do
@@ -94,7 +88,7 @@ describe GdsApi::Organisations do
       }
     end
 
-    before do
+    it "handles pagination" do
       organisation_api
         .given("the organisation list is paginated, beginning at page 1")
         .upon_receiving("a request without a query param")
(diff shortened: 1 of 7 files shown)


test/organisations_api_test.rb
```diff
@@ -106,64 +100,48 @@ def organisation(slug: "test-department")
...
         .with(request.merge(query: "page=2"))
         .will_respond_with(response.merge(headers: { "link" => page_two_links }))
-    end
 
-    it "should handle pagination" do
-      response = api_client.organisations
-
-      assert_equal 20, response["results"].count
-      assert_equal 40, response.with_subsequent_pages.count
+      api_client.organisations.with_subsequent_pages.count
```
hannako, line 104: No need to call `.count` here anymore

test/organisations_api_test.rb
```diff
@@ -108,7 +108,10 @@ def organisation(slug: "test-department")
   it "fetches an organisation by slug" do
     hmrc = "hm-revenue-customs"
     api_response = organisation(slug: hmrc)
-    api_response["id"] = "www.gov.uk/api/organisations/#{hmrc}"
+    api_response["id"] = Pact.term(
+      generate: "www.gov.uk/api/organisations/#{hmrc}",
+      matcher: /gov\.uk\/api\/organisations\/#{hmrc}$/,
+    )
```
hannako, line 114: Minor comment: I find it a bit strange that we are using two different ways of matching the host in the same file. Maybe we should use the regex here as well for consistency?

hannako (approved):
Looks like a good plan to me 👍 

The diff for the publishing api test is a bit of a monster, but I can't spot anything contentious!

test/organisations_api_test.rb
ChrisBAshton: Thanks - I'll look at removing this now. I seem to remember the second HTTP request didn't happen unless I appended both `.with_subsequent_pages` and `.count`, but I'll take another look to confirm.

ChrisBAshton: Yes - the `.count` is needed. Removing it causes this failure:

```
  1) Error:
GdsApi::Organisations::fetching a paginated list of organisations#test_0001_handles pagination:
Pact::MockService::VerificationFailedError: Actual interactions do not match expected interactions for mock Collections Organisation API.

Missing requests:
	GET /api/organisations?page=2

See log/collections_organisation_api_mock_service.log for details.
```

test/organisations_api_test.rb
ChrisBAshton: Good idea. I did consider it, but dismissed it because the previous test value was `"www.gov.uk/api/organisations/#{hmrc}"`, i.e. missing the `http(s)://` prefix. But looking at it again, I don't think it matters too much - it's quite possible the test value was arbitrarily chosen and that the presence of a HTTP prefix doesn't matter.

I've rewritten the commit, moving the `host_agnostic_endpoint_regex` out of the single test it is in so that it can be reused here.

thomasleese (approved):
My only concern was if we lost the tests of the API client itself, but that seems to be covered here still: https://github.com/alphagov/gds-api-adapters/blob/master/test/test_helpers/publishing_api_test.rb

PR #1060, opened 2021-05-05
Test api/attributes/names endpoint in account-api

AlanGabbianelli:
This endpoint allows querying if the user has a value stored for an
attribute, without returning the value.

It is very similar to /api/attributes, with the difference that it only
returns the name of the attribute instead of the name and the value.

See where this endpoint was introduced: https://github.com/alphagov/account-api/pull/58

Trello card: https://trello.com/c/e7JQlXx9/710-add-an-endpoint-to-account-api-to-check-if-an-attribute-is-defined

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Add optional `level_of_authentication` parameter to Account API `get_sign_in_url`.
 * Store response body in new `http_body` field of HTTP exceptions
 * Added Pact tests for the `LinkCheckerApi` adapters.
+* Add get_attributes_names method for [account-api's /api/attributes/names endpoint](https://github.com/alphagov/account-api/pull/58) + Pact tests.
 
 # 71.0.0
 
--- a/lib/gds_api/account_api.rb
+++ b/lib/gds_api/account_api.rb
@@ -85,6 +85,17 @@ class GdsApi::AccountApi < GdsApi::Base
     patch_json("#{endpoint}/api/attributes", { attributes: attributes }, auth_headers(govuk_account_session))
   end
 
+  # Look up the names of a user's attributes
+  #
+  # @param [String] attributes Names of the attributes to check
+  # @param [String] govuk_account_session Value of the session header
+  #
+  # @return [Hash] The attribute names (if present), and a new session header
+  def get_attributes_names(attributes:, govuk_account_session:)
+    querystring = nested_query_string({ attributes: attributes }.compact)
+    get_json("#{endpoint}/api/attributes/names?#{querystring}", auth_headers(govuk_account_session))
+  end
+
 private
 
   def nested_query_string(params)
--- a/lib/gds_api/test_helpers/account_api.rb
+++ b/lib/gds_api/test_helpers/account_api.rb
@@ -95,6 +95,18 @@ module GdsApi
             .to_return(status: 200, body: { govuk_account_session: new_govuk_account_session }.compact.to_json)
         end
       end
+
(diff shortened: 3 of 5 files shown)


lib/gds_api/account_api.rb
```diff
@@ -78,6 +78,17 @@ def set_attributes(attributes:, govuk_account_session:)
     patch_json("#{endpoint}/api/attributes", { attributes: attributes }, auth_headers(govuk_account_session))
   end
 
+  # Look up the names of a user's attributes
+  #
+  # @param [String] attributes Names of the attributes to check
+  # @param [String] govuk_account_session Value of the session header
+  #
+  # @return [Hash] The attribute names (if present), and a new session header
+  def get_attributes_names(attributes_names:, govuk_account_session:)
```
barrucadu, line 87: I think call the parameter `attributes`, as the two plurals in `attributes_names` feels weird to me

AlanGabbianelli: What about `attribute_names`?

barrucadu: `get_attributes` uses `attributes` for its parameter, which is a list of names; so for consistency I think this should too.

test/account_api_test.rb
```diff
@@ -275,4 +275,62 @@
...
+
+      describe "attributes with no value exist" do
+        let(:given) { "there is a valid user session, with an attribute called 'foo' that has no value" }
+        let(:returned_attribute_names) { [] }
+
+        it "responds with the attribute values" do
+          response = api_client.get_attributes_names(govuk_account_session: govuk_account_session, attributes: queried_attribute_names)
+          assert_equal returned_attribute_names, response["values"]
+        end
+      end
```
barrucadu, line 333: Is this testing anything more than the `responds with 200 OK, a new govuk_account_session, and no attributes` test?

barrucadu: That test is also checking that a 404ing attribute is not included in the response

AlanGabbianelli: I just closed the [related PR](https://github.com/alphagov/account-api/pull/62) that introduced the related `provider_state`. If an attribute isn't present or it has a `nil` value, the response would be the same, and we don't want attributes with `nil` values anyway (see [this comment](https://github.com/alphagov/account-api/pull/62#issuecomment-833427248)).

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Add optional `level_of_authentication` parameter to Account API `get_sign_in_url`.
 * Store response body in new `http_body` field of HTTP exceptions
 * Added Pact tests for the `LinkCheckerApi` adapters.
+* Added Pact tests for the `AccountApi` adapters.
```
barrucadu, line 7: The interesting change here is adding the `get_attributes_names` method, not that we also added a test for it

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -95,6 +95,18 @@ def stub_account_api_set_attributes(govuk_account_session: nil, attributes: nil,
             .to_return(status: 200, body: { govuk_account_session: new_govuk_account_session }.compact.to_json)
         end
       end
+
+      def stub_account_api_get_attributes_names(govuk_account_session: nil, attributes: [], new_govuk_account_session: nil)
+        querystring = Rack::Utils.build_nested_query({ attributes: attributes }.compact)
+        if govuk_account_session
```
huwd, line 101: How doe sthe the else block work here?
It seems to return the same thing without headers.

Wouldn't this be an authenticated endpoint that returns 401 if there's not session?

barrucadu (approved):

PR #1063, opened 2021-05-06
Store response body in new `http_body` field of HTTP exceptions

barrucadu:
We don't currently make the response body easily available - it can
only be extracted from the `message` field, by splitting it at the
"Response body:\n" bit, which is a bit unpleasant.  Better for it to
be directly available without manipulating strings.

This is useful for the accounts work, as the account-api is going to
start returning 403 errors when a user doesn't have a sufficiently
high level-of-authentication to access an attribute, and it'll return
the required level-of-authentication in the response body.

An alternative implementation would be to return a 200 response with a
body like:

    { "error": "level of authentication too low", ... }

But this is (1) error prone (you need to remember to check for an
"error" field), and (2) confusing (an error response should have an
error code).  Returning a 403 response, which throws an exception from
gds-api-adapters and so is *guaranteed* to interrupt the normal flow
of control, is much better.

This is a non-breaking change, as the new parameter is the last of
`HTTPErrorResponse` and defaults to `nil`, so anyone constructing
these exceptions in their own code won't have to change things.

---

[Trello card](https://trello.com/c/G2uYUDuz/716-check-the-level-of-authentication-when-using-attributes-in-the-account-api)

### DIFF
--- a/test/json_client_test.rb
+++ b/test/json_client_test.rb
@@ -85,6 +85,13 @@ class JsonClientTest < MiniTest::Spec
     assert_equal GdsApi::Response, @client.get_json(url).class
   end
 
+  def test_should_include_body_for_http_error
+    url = "http://some.endpoint/some.json"
+    stub_request(:get, url).to_return(body: "response body goes here", status: 404)
+    error = assert_raises(GdsApi::HTTPErrorResponse) { @client.get_json(url) }
+    assert_equal "response body goes here", error.http_body
+  end
+
   def test_get_bang_should_raise_http_not_found_if_404_returned_from_endpoint
     url = "http://some.endpoint/some.json"
     stub_request(:get, url).to_return(body: "{}", status: 404)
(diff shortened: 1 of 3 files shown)


test/json_client_test.rb
```diff
@@ -85,6 +85,17 @@ def test_should_fetch_and_parse_json_into_response
...
+    url = "http://some.endpoint/some.json"
+    stub_request(:get, url).to_return(body: "response body goes here", status: 404)
+    begin
+      @client.get_json(url)
+      assert_equal false, true
+    rescue GdsApi::HTTPErrorResponse => e
+      assert_equal "response body goes here", e.http_body
+    end
+  end
+
```
AlanGabbianelli, line 98: This is a bit confusing. Can't we stub it to raise an error instead of relying on a failing assertion and a rescue block?

barrucadu: It is raising an error, the `assert_equal true, false` bit is so that the test fails if the exception *isn't* raised

barrucadu: Aha, `assert_raises` returns the error, I can simplify this: https://www.rubydoc.info/gems/minitest/Minitest%2FAssertions:assert_raises

AlanGabbianelli: Much clearer, thank you

AlanGabbianelli (approved):

PR #1064, opened 2021-05-06
Add 401 and 403 test helpers for Account API methods

barrucadu:
We don't currently expose an easy way to create error cases in tests, but those will be useful as we roll out levels of authentication and want to test that apps do the right thing in the failure cases.

### DIFF
--- a/lib/gds_api/test_helpers/account_api.rb
+++ b/lib/gds_api/test_helpers/account_api.rb
@@ -56,11 +56,51 @@ module GdsApi
         )
       end
 
-      def stub_account_api_set_email_subscription(slug: "slug", **options)
+      def stub_account_api_unauthorized_get_email_subscription(**options)
+        stub_account_api_request(
+          :get,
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/account_api.rb
```diff
@@ -38,73 +38,69 @@ def stub_account_api_create_registration_state(attributes: nil, state_id: "state
...
-            .to_return(status: 200, body: { govuk_account_session: new_govuk_account_session, has_subscription: true }.compact.to_json)
-        end
+      def stub_account_api_has_email_subscription(**options)
+        stub_account_api_request(
+          :get,
+          "/api/transition-checker-email-subscription",
+          response_body: { has_subscription: true },
+          **options,
+        )
       end
```
AlanGabbianelli, line 48: Nice refactor, these are much easier to read now ⭐

AlanGabbianelli (approved):

PR #1066, opened 2021-05-11
Add Pact tests for Imminence API

ChrisBAshton:
Trello: https://trello.com/c/clyNd0c5/2521-3-finish-adding-contract-tests-for-imminence

Sister PR: https://github.com/alphagov/imminence/pull/644


test/imminence/imminence_api_pact_test.rb
```diff
@@ -11,45 +11,47 @@
 
     it "responds with all responses for the given dataset" do
       imminence_api
+        .given("a service exists called number-plate-supplier with places")
         .upon_receiving("the request to retrieve all places for the current dataset")
         .with(
           method: :get,
           path: "/places/number-plate-supplier.json",
           headers: GdsApi::JsonClient.default_request_headers,
+          query: { lat: "1.11", lng: "2.22", limit: "5" },
```
ChrisBAshton: Could fee44565e66f83c7854a254108d25265ade70f5e be rebased into cdc3135a75555a948c918115f26f90a1c99fafaa?

test/imminence/imminence_api_pact_test.rb
```diff
@@ -17,7 +17,7 @@
           method: :get,
           path: "/places/number-plate-supplier.json",
           headers: GdsApi::JsonClient.default_request_headers,
-          query: { lat: "1.11", lng: "2.22", limit: "5" },
+          query: { lat: "-2.01", lng: "53.1", limit: "5" },
```
ChrisBAshton, line 20: Could this be rebased into that commit too?
And then in your commit message - or in an inline comment in the code here - explain why these coordinates are used. If I remember correctly, they're somewhat arbitrary but should be similar to those added in https://github.com/alphagov/imminence/pull/644, but not exactly the same since we want to test the proximity logic works properly.

test/imminence/imminence_api_pact_test.rb
```diff
@@ -26,9 +26,9 @@
               _id: { "$oid": "60867a0ee90e0703aed18e46" },
               access_notes: nil,
               address1: "Yarrow Road Tower Park",
-              address2: "Poole",
+              address2: nil,
```
ChrisBAshton, line 29: These changes can be rebased too.

Jenkinsfile
```diff
@@ -56,7 +56,7 @@ node("postgresql-9.6") {
         runPactTests(govuk, "frontend", FRONTEND_BRANCH)
         runPactTests(govuk, "account-api", ACCOUNT_API_BRANCH, [ resetDatabase: true ])
         runPactTests(govuk, "link-checker-api", LINK_CHECKER_API_BRANCH, [ resetDatabase: true ])
-        runPactTests(govuk, "imminence", IMMINENCE_BRANCH, [ resetDatabase: true ])
+        runPactTests(govuk, "imminence", IMMINENCE_BRANCH, [ resetDatabase: true, createIndexes: true ])
```
ChrisBAshton, line 59: These changes can be rebased into 2c47625da56bd00f4a64453db732f93fd3b61f94, since the original Jenkinsfile changes won't have worked in isolation and therefore need this change too.

We should also explain in the commit why this is necessary.

ChrisBAshton (commented):
Well done for persevering on this, @callumknights. The code looks good, my feedback is mainly around the git commit structure, which should be broken down into discrete self-contained changes. Let me know if you'd like to pair on the rebasing, as it can be daunting to get right!

test/imminence/imminence_api_pact_test.rb
callumknights: Sorted 👍

test/imminence/imminence_api_pact_test.rb
callumknights: 👍

test/imminence/imminence_api_pact_test.rb
callumknights: 👍

Jenkinsfile
callumknights: 👍

Rakefile
```diff
@@ -8,7 +8,7 @@ end
 
 Rake::TestTask.new("test") do |t|
   t.libs << "test"
-  t.test_files = FileList["test/**/*_test.rb"]
+  t.test_files = FileList["test/imminence/imminence_api_pact_test.rb"] # FileList["test/**/*_test.rb"]
```
ChrisBAshton, line 11: This change should be removed.

ChrisBAshton (commented):
This is looking great 👍

There's one last (fairly major) thing I'm struggling with, and that is we only have - as the commit suggests - a "First Pact test for Imminence". The intention was that we'd add more.

We're now exercising the API response for the [Places API](https://github.com/alphagov/imminence/blob/main/test/integration/places_api_test.rb). Whilst there are other things we could cover (such as the CSV/KML responses, and the other endpoints such as [searching places by slug & postcode](https://github.com/alphagov/imminence/blob/main/test/integration/places_api_test.rb#L270)), the actual API response looks the same as what we've already tested for, so they just seem like slightly different ways of calling the API. I think they're therefore quite low risk from a contract testing perspective, and as they're already covered by tests in Imminence, I'm happy to leave them out.

But I do think we need to write at least one more Pact test for the other Imminence API: "Areas API". Like the Places API, this has multiple ways of calling it ([Areas By Type](https://github.com/alphagov/imminence/blob/main/test/integration/areas_by_type_test.rb#L16-L37) and [Areas by Postcode](https://github.com/alphagov/imminence/blob/main/test/integration/areas_by_postcode_test.rb#L21-L36)), but I think we only really need to pick one of them for a Pact test. Hopefully now all the setup work is done, it should be a lot quicker to add these. What do you think?

Either way, I don't think it should hold up this PR - I think the new tests should be added in a follow-up PR. Let's get this one merged!

Rakefile
callumknights: Removed 👍

ChrisBAshton (commented):
LGTM 👍 I'd approve, but as per https://github.com/alphagov/imminence/pull/644#pullrequestreview-698122837, I'm not able to.

1pretz1 (approved):

PR #1067, opened 2021-05-11
Add Account API Feature: Save a page

huwd:
[Trello card](https://trello.com/c/OvsvoWsO/766-implement-an-api-to-save-a-page-unsave-a-page-check-if-a-page-is-saved-and-return-the-list-of-saved-pages)

Also relevant: [account-api#69](https://github.com/alphagov/account-api/pull/69)

## What

Add save page endpoints allowing an authenticated user to get an index of all their saved pages, as well ass add, delete or query a single saved page.

### DIFF
--- a/lib/gds_api/test_helpers/account_api.rb
+++ b/lib/gds_api/test_helpers/account_api.rb
@@ -302,6 +302,36 @@ module GdsApi
           **option,
         }
       end
+
+      ####################################
+      # DELETE /api/saved-pages/:page_path
+      ####################################
+      def stub_account_api_delete_saved_page(page_path:, **options)
+        stub_account_api_request(
+          :delete,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_status: 204,
+          **options,
+        )
+      end
+
+      def stub_account_api_delete_saved_page_does_not_exist(page_path:, **options)
+        stub_account_api_request(
+          :delete,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_status: 404,
+          **options,
+        )
+      end
+
+      def stub_account_api_delete_saved_page_unauthorised(page_path:, **options)
+        stub_account_api_request(
+          :delete,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_status: 401,
+          **options,
+        )
+      end
     end
   end
 end
--- a/lib/gds_api/test_helpers/account_api.rb
+++ b/lib/gds_api/test_helpers/account_api.rb
@@ -258,6 +258,50 @@ module GdsApi
           **options,
         )
       end
+
+      #################################
+      # PUT /api/saved_pages/:page_path
+      #################################
+      def stub_account_api_save_page(page_path:, **options)
+        stub_account_api_request(
+          :put,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_body: { saved_page: { page_path: page_path } },
+          **options,
+        )
+      end
+
+      def stub_account_api_save_page_already_exists(page_path:, **options)
+        stub_account_api_save_page(page_path: page_path, **options)
+      end
+
+      def stub_account_api_save_page_cannot_save_page(page_path:, **options)
+        stub_account_api_request(
+          :put,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_status: 422,
+          response_body: { **cannot_save_page_problem_detail({ page_path: page_path }) },
+          **options,
+        )
+      end
+
+      def stub_account_api_unauthorized_save_page(page_path:, **options)
+        stub_account_api_request(
+          :put,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_status: 401,
+          **options,
+        )
+      end
+
+      def cannot_save_page_problem_detail(option = {})
+        {
+          title: "Cannot save page",
+          detail: "Cannot save page with path #{option['page_path']}, check it is not blank, and is a well formatted url path.",
+          type: "https://github.com/alphagov/account-api/blob/main/docs/api.md#cannot-save-page",
+          **option,
+        }
+      end
     end
   end
 end
--- a/lib/gds_api/test_helpers/account_api.rb
+++ b/lib/gds_api/test_helpers/account_api.rb
@@ -207,6 +207,27 @@ module GdsApi
           stub_request(method, "#{ACCOUNT_API_ENDPOINT}#{path}").with(**with).to_return(**to_return)
         end
       end
+
+      ######################
+      # GET /api/saved_pages
+      ######################
+      def stub_account_api_returning_saved_pages(saved_pages: [], **options)
+        stub_account_api_request(
+          :get,
+          "/api/saved_pages",
+          response_body: { saved_pages: saved_pages },
+          **options,
+        )
+      end
+
+      def stub_account_api_unauthorized_get_saved_pages(**options)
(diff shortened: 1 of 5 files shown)


lib/gds_api/saved_page_api.rb
```diff
@@ -0,0 +1,47 @@
...
+# @api documented
+
+class GdsApi::SavedPageApi < GdsApi::AccountApi
+  # Look up all pages saved by a user in their Account
+  #
+  # @param [String] govuk_account_session Value of the session header
+  #
+  # @return [Hash] containing :saved_pages, an array of single saved page hashes
+  def get_saved_pages(govuk_account_session:)
+    get_json("#{endpoint}/api/saved-pages", auth_headers(govuk_account_session))
```
barrucadu, line 14: The endpoint is `/api/saved_pages`

lib/gds_api/saved_page_api.rb
```diff
@@ -0,0 +1,47 @@
+# Adapter for the Saved Page API
+# Authenticated with Account API
+#
+# @see https://github.com/alphagov/account-api
+# @api documented
+
+class GdsApi::SavedPageApi < GdsApi::AccountApi
```
barrucadu, line 7: I still find this a bit weird, as the subclassing means that `GdsApi::SavedPageApi` has a *superset* of the methods of the `GdsApi::AccountApi`.

If you can always replace `GdsApi::AccountApi` with `GdsApi::SavedPageApi`, why would you ever use `GdsApi::AccountApi`?

barrucadu: Actually, having realised that now, I think this should be a separate class *only if* the only methods it exposes are saved page methods.  So right now, I don't think this should be a separate class.

lib/gds_api/saved_page_api.rb
huwd: Yeah hang on, no point reviewing this yet until the account api bit is set.
There have been changes over there i need to reflect here too.

lib/gds_api/saved_page_api.rb
huwd: I guess I was thinking if the `AccountAPI` was the core API that covered authentication etc.
But then feature specific authenticated APIs were separate you could bring in a specific set methods for a task at hand.

So in that case we might have;
`AccountApi`
`EmailSubsriptionApi`
`SavedPageApi`

I'm not sure where attribute would live in there, they feel kind of integral to the AccountApi 🤔 

otherwise the AccountApi just starts getting huge as we bolt on more and more bits?

barrucadu: > otherwise the AccountApi just starts getting huge as we bolt on more and more bits?

But we don't have that problem *yet* - the pattern in gds-api-adapters at the moment is to have one class per app, and even much bigger APIs than ours (like publishing-api or email-alert-api) follow that pattern.

Splitting things up when we have only 12 endpoints feels like a premature optimisation not worth the inconsistency.

test/account_api_test.rb
```diff
@@ -287,4 +287,43 @@
...
     end
   end
+
+  describe "requesting all pages saved by a user" do
+    describe "the user is logged in" do
+      let(:given) { "there is a valid user session" }
+      let(:saved_pages) { [] }
+
+      before do
+        saved_page_api
```
barrucadu, line 297: This should be `account_api`, not `saved_page_api`

test/test_helpers/pact_helper.rb
```diff
@@ -3,6 +3,7 @@
 BANK_HOLIDAYS_API_PORT = 3003
 ACCOUNT_API_PORT = 3004
 LINK_CHECKER_API_PORT = 3005
+SAVED_PAGE_API_PORT = 3006
```
barrucadu, line 6: The changes to this file aren't needed, remove them

test/account_api_test.rb
```diff
@@ -287,4 +287,129 @@
...
+      let(:page_path) { "/guidance/some-govuk-guidance" }
+
+      before do
+        account_api
+          .given(given)
+          .upon_receiving("a PUT saved-page/:page_path request")
+          .with(
+            method: :put,
+            path: "/api/saved_pages/#{CGI.escape(page_path)}",
+            # headers: GdsApi::JsonClient.default_request_headers.merge(authenticated_headers),
```
huwd, line 375: This mystifies me, if I uncomment this line `bundle exec rake` starts exibiting failures saying that the expected header was not passed through 🤔

test/account_api_test.rb
```diff
@@ -287,4 +287,129 @@
...
+      let(:page_path) { "/guidance/some-govuk-guidance" }
+
+      before do
+        account_api
+          .given(given)
+          .upon_receiving("a DELETE saved-page/:page_path request")
+          .with(
+            method: :delete,
+            path: "/api/saved_pages/#{CGI.escape(page_path)}",
+            # headers: GdsApi::JsonClient.default_request_headers.merge(authenticated_headers),
```
huwd, line 405: ditto L375

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -207,6 +207,140 @@ def stub_account_api_request(method, path, with: {}, response_status: 200, respo
...
+      # GET /api/saved_pages
+      ######################
+      def stub_account_api_has_saved_pages(**options)
+        stub_account_api_request(
+          :get,
+          "/api/saved_pages",
+          response_body: { saved_pages: [{ page_path: "/foo" }, { page_path: "/bar" }] },
+          **options,
+        )
+      end
```
barrucadu, line 221: This should take the saved pages to return as a parameter, like the `values` param in

```ruby
      def stub_account_api_has_attributes(attributes: [], values: {}, **options)
        querystring = Rack::Utils.build_nested_query({ attributes: attributes }.compact)
        stub_account_api_request(
          :get,
          "/api/attributes?#{querystring}",
          response_body: { values: values },
          **options,
        )
      end
```

Then `stub_account_api_does_not_have_saved_pages` isn't needed any more

huwd: sure, i renamed it to `stub_account_api_returning_saved_pages`, as I thought otherwise the naming with `has` and returning something empty might be unexepected.

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -207,6 +207,131 @@ def stub_account_api_request(method, path, with: {}, response_status: 200, respo
           stub_request(method, "#{ACCOUNT_API_ENDPOINT}#{path}").with(**with).to_return(**to_return)
         end
       end
+
+      ######################
+      # GET /api/saved_pages
+      ######################
+      def stub_account_api_returning_saved_pages(saved_pages: {}, **options)
```
barrucadu, line 214: The default value should be a `[]`

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -207,6 +207,131 @@ def stub_account_api_request(method, path, with: {}, response_status: 200, respo
...
+      end
+
+      def stub_account_api_does_not_have_saved_page(page_path:, **options)
+        stub_account_api_request(
+          :get,
+          "/api/saved_pages/#{CGI.escape(page_path)}",
+          response_body: { saved_page: {} },
+          **options,
+        )
+      end
```
barrucadu, line 251: This returns a 404: https://github.com/alphagov/account-api/blob/main/app/controllers/saved_pages_controller.rb#L4

barrucadu (approved):

PR #1070, opened 2021-06-02
Log error messages when raising a timeout

bilbof:
This will preserve the error message - crucially
it will tell us whether a timeout is a connection
or read timeout, or a connection reset error.

This will be present in logs unless the timeout error
is rescued, and will help us debug timeout errors.


sihugh (approved):

sihugh:
RBENV_VERSION issue should be fixed by https://github.com/alphagov/govuk-jenkinslib/pull/87, so we can drop that commit if it gets merged ahead of this. Nice find!

PR #1077, opened 2021-06-14
Add methods to lookup local authority links by GSS Code

brucebolt:
We currently assign each local authority a slug. API requests have the GET parameter `authority_slug`.

This allows the use of an alternative data source (e.g. the Ordnance Survey Names API) that provides us with the GSS Code.

[Trello card](https://trello.com/c/2x6Eh2b3)


brucebolt:
This was opened as part of a spike.  Closing as the functionality has been implemented another way.

PR #1078, opened 2021-06-15
Permit bearer_token field to be JSON

bilbof:
This enables the bearer_token option for the JsonClient to be a JSON with the structure `'{"value": <string>}'`.

Trello: https://trello.com/c/6Uaayg1l/545-use-a-single-rotation-lambda-instead-of-duplicating-them-per-secret

### Motivation

We would like to store additional data with the secret body, to enable rotation of the bearer token with configuration for the rotation provided in the secret body.

E.g. the secret body may look like this:

```json
{ "api_user": "publisher@alphagov.co.uk", "value": "123", "application": "publishing-api" }
```

The `value` field would contain the bearer token string. The remaining fields would enable rotation by a Lambda.

This should not have an impact on the existing platform. I expect secrets stored in `govuk-secrets` to remain unchanged. This will only impact secrets stored in AWS SecretsManager for the new ECS-hosted platform.

### DIFF
--- a/lib/gds_api/json_client.rb
+++ b/lib/gds_api/json_client.rb
@@ -18,6 +18,7 @@ module GdsApi
 
       @logger = options[:logger] || NullLogger.instance
       @options = options
+      @bearer_token = parse_bearer_token(options[:bearer_token])
     end
 
     def self.default_request_headers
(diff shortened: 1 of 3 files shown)


bilbof:
This isn't necessary, since we can specify `json-key` in the ARN provided in the `valueFrom` field: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data-secrets.html#secrets-envvar

Thanks @sengi!

PR #1095, opened 2021-07-08
Add new areas/postcode pact test

callumknights:
An additional pact test for imminence, so we can test an endpoint that doesn't just return a place (as most of them do).

Trello: https://trello.com/c/clyNd0c5/2521-3-finish-adding-contract-tests-for-imminence


ChrisBAshton (approved):
LGTM 👍 nice one!

PR #1098, opened 2021-07-20
Delete account api oidc users

huwd:
A part of: [trello ticket](https://trello.com/c/b95YV8qC/912-%E2%9D%97-delete-users-in-account-api-when-theyre-deleted-in-account-manager)
See also: [account-api#154](https://github.com/alphagov/account-api/pull/154)

We are currently migrating features from our PaaS based
account-manager-prototype
to account-api.

Accounts API now acts as an additional store of user data. When a user
deletes their account through the account manager we require an
authenticated endpoint that will forward the request to account-api so
we can remove their data there too.

The API has a new delete user endpoint, these should be called via
the GDS API wrapper. This PR adds those methods here.

### DIFF
--- a/lib/gds_api/account_api.rb
+++ b/lib/gds_api/account_api.rb
@@ -43,6 +43,13 @@ class GdsApi::AccountApi < GdsApi::Base
     get_json("#{endpoint}/api/user", auth_headers(govuk_account_session))
   end
 
+  # Delete a users account
+  #
+  # @param [String] subject_identifier The identifier of the user, shared between the auth service and GOV.UK.
+  def delete_user_by_subject_identifier(subject_identifier:)
+    delete_json("#{endpoint}/api/oidc-users/#{subject_identifier}")
+  end
+
   # Update the user record with privileged information from the auth service.  Only the auth service will call this.
   #
   # @param [String] subject_identifier The identifier of the user, shared between the auth service and GOV.UK.
--- a/lib/gds_api/test_helpers/account_api.rb
+++ b/lib/gds_api/test_helpers/account_api.rb
@@ -82,6 +82,26 @@ module GdsApi
         )
       end
 
+      ############################################
+      # DELETE /api/oidc-users/:subject_identifier
+      ############################################
+
+      def stub_account_api_delete_user_by_subject_identifier(subject_identifier:)
+        stub_account_api_request(
+          :delete,
+          "/api/oidc-users/#{subject_identifier}",
+          response_status: 204,
+        )
+      end
+
+      def stub_account_api_delete_user_by_subject_identifier_does_not_exist(subject_identifier:)
+        stub_account_api_request(
+          :delete,
+          "/api/oidc-users/#{subject_identifier}",
+          response_status: 404,
+        )
+      end
+
       ###########################################
       # PATCH /api/oidc-users/:subject_identifier
       ###########################################
--- a/test/account_api_test.rb
+++ b/test/account_api_test.rb
@@ -108,6 +108,32 @@ describe GdsApi::AccountApi do
     end
   end
 
+  describe "#delete_user_by_subject_identifier" do
+    let(:subject_identifier) { "the-subject-identifier" }
+    let(:path) { "/api/oidc-users/#{subject_identifier}" }
+
+    it "responds with 204 No Content if the user existed and has been deleted" do
+      account_api
+        .given("there is a user with subject identifier '#{subject_identifier}'")
+        .upon_receiving("a delete-user request for '#{subject_identifier}'")
+        .with(method: :delete, path: path, headers: headers)
+        .will_respond_with(status: 204)
+
+      api_client.delete_user_by_subject_identifier(subject_identifier: subject_identifier)
+    end
+
+    it "responds with 404 if the user does not exist" do
+      account_api
+        .upon_receiving("a delete-user request")
+        .with(method: :delete, path: path, headers: headers)
+        .will_respond_with(status: 404)
+
+      assert_raises GdsApi::HTTPNotFound do
+        api_client.delete_user_by_subject_identifier(subject_identifier: subject_identifier)
+      end
+    end
+  end
+
   describe "the user is logged in" do
     let(:govuk_account_session) { "logged-in-user-session" }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Add `delete_user_by_subject_identifier` (for Account API)
+
 # 72.0.0
 
 - BREAKING: Remove `create_registration_state` method and helpers (for Account API)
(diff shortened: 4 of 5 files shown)


test/account_api_test.rb
```diff
@@ -308,6 +308,31 @@
         api_client.get_attributes_names(govuk_account_session: govuk_account_session, attributes: [attribute_name])
       end
     end
+
+    describe "#delete_user" do
+      let(:path) { "/api/user" }
+
+      it "responds with 204 OK if the user existed and has been deleted" do
+        binding.pry
```
bevanloon, line 316: Think this is causing your build to fail 😬

huwd: ah... i thought it was the pact tests waiting for t[his to get merged](https://github.com/alphagov/account-api/pull/154),
woops! thanks!

lib/gds_api/account_api.rb
```diff
@@ -43,6 +43,13 @@ def get_user(govuk_account_session:)
     get_json("#{endpoint}/api/user", auth_headers(govuk_account_session))
   end
 
+  # Delete a users account
+  #
+  # @param [String] govuk_account_session Value of the session header
+  def delete_user(govuk_account_session:)
+    delete_json("#{endpoint}/api/user", auth_headers(govuk_account_session))
+  end
```
barrucadu, line 51: Like `update_user_by_subject_identifier`, this should use `"#{endpoint}/api/oidc-users/#{subject_identifier}"` - maybe call it `delete_user_by_subject_identifier` to match?

barrucadu: In https://github.com/alphagov/account-api/pull/157 I've changed the code example to `delete_user_by_subject_identifier`

lib/gds_api/account_api.rb
```diff
@@ -43,6 +43,13 @@ def get_user(govuk_account_session:)
     get_json("#{endpoint}/api/user", auth_headers(govuk_account_session))
   end
 
+  # Delete a users account
+  #
+  # @param [String] subject_identifier The identifier of the user, shared between the auth service and GOV.UK.
+  def delete_user_by_subject_identifier(subject_identifier:)
+    delete_json("#{endpoint}/api/user/#{subject_identifier}")
```
barrucadu, line 50: It's `/aoi/oidc-users`

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -82,6 +82,37 @@ def stub_account_api_unauthorized_user_info(**options)
...
+      def stub_account_api_not_found_delete_account(subject_identifier:, **options)
+        stub_account_api_request(
+          :delete,
+          "/api/user/#{subject_identifier}",
+          response_status: 404,
+          **options,
+        )
+      end
+
+      def stub_account_api_unauthorized_delete_account(subject_identifier:, **options)
```
barrucadu, line 107: I think you can skip this case, as if you get a 401 you're using the wrong signon token

test/account_api_test.rb
```diff
@@ -108,6 +108,33 @@
     end
   end
 
+  describe "#delete_user_by_subject_identifier" do
+    let(:subject_identifier) { "the-subject-identifier" }
+    let(:path) { "/api/oidc-users/#{subject_identifier}" }
+
+    it "responds with 204 OK if the user existed and has been deleted" do
+      account_api
+        .given("there is a valid user session, and an existing user")
```
barrucadu, line 117: This is failing because there is no `provider_state` in account-api with this name.

One option would be to drop this `given`, as the subject identifier `user-id` [is guaranteed to exist](https://github.com/alphagov/account-api/blob/main/spec/service_consumers/pact_helper.rb#L37) and then use that instead of `the-subject-identifier`.

But it would probably be clearer to change this to

```ruby
account_api
  .given("there is a user with subject identifier '#{subject_identifier}'")
  .upon_receiving("a delete-user request for '#{subject_identifier}'")
  .with(method: :delete, path: path, headers: headers)
  .will_respond_with(status: 204)
```

And then add a `provider_state` to account-api called `there is a user with subject identifier 'subject-identifier'`

test/account_api_test.rb
```diff
@@ -108,6 +108,33 @@
...
+        .upon_receiving("a delete-user request")
+        .with(method: :delete, path: path, headers: headers_with_json_body)
+        .will_respond_with(status: 204, headers: json_response_headers)
+
+      api_client.delete_user_by_subject_identifier(subject_identifier: subject_identifier)
+    end
+
+    it "responds with 404 if the user does not exist" do
+      account_api
+        .given("there is a valid user session, and the user cannot be found")
```
barrucadu, line 127: This doesn't need a `given`, because (1) it doesn't need a valid user session, and (2) the default state is the user not existing.

It also doesn't need `headers_with_json_body` (only `headers`) because there is no response body.

test/account_api_test.rb
huwd: Something like this?
https://github.com/alphagov/account-api/pull/158

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -82,6 +82,28 @@ def stub_account_api_unauthorized_user_info(**options)
         )
       end
 
+      ########################
+      # DELETE /api/oidc-users
+      ########################
+
+      def stub_account_api_success_delete_account(subject_identifier:, **options)
```
barrucadu, line 89: To match the naming convention of other methods, this should be `stub_account_api_delete_user_by_subject_identifier`

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -82,6 +82,28 @@ def stub_account_api_unauthorized_user_info(**options)
...
+        stub_account_api_request(
+          :delete,
+          "/api/oidc-users/#{subject_identifier}",
+          response_status: 204,
+          **options,
+        )
+      end
+
+      def stub_account_api_not_found_delete_account(subject_identifier:, **options)
```
barrucadu, line 98: And this should be `stub_account_api_delete_user_by_subject_identifier_does_not_exist`

test/account_api_test.rb
```diff
@@ -108,6 +108,32 @@
     end
   end
 
+  describe "#delete_user_by_subject_identifier" do
+    let(:subject_identifier) { "the-subject-identifier" }
+    let(:path) { "/api/oidc-users/#{subject_identifier}" }
+
+    it "responds with 204 OK if the user existed and has been deleted" do
```
barrucadu, line 115: 204 is "No Content", not "OK"

lib/gds_api/test_helpers/account_api.rb
```diff
@@ -82,6 +82,28 @@ def stub_account_api_unauthorized_user_info(**options)
         )
       end
 
+      ########################
+      # DELETE /api/oidc-users
```
barrucadu, line 86: `DELETE /api/oidc-users/:subject_identifier`

barrucadu (commented):
Looks good, just needs a CHANGELOG entry and I'll approve

barrucadu (approved):

PR #1116, opened 2021-11-01
Add `match_user_by_email` for account-api

barrucadu:
See https://github.com/alphagov/account-api/pull/261

---

[Trello card](https://trello.com/c/5ZoujfPL/1115-implement-email-management-account-sniffer-logic)


huwd (commented):
Looks good, will approve when the other one has gone through.
If we decied rate limiting was right now there may be an additional status code, no?

huwd (approved):

PR #1121, opened 2021-12-07
Release version 76.0.0

alex9smith:
Includes [Remove `has_unconfirmed_email` parameter from Account API calls (#1120)](https://github.com/alphagov/gds-api-adapters/pull/1120)

[Trello](https://trello.com/c/SdkcF8EO/13-remove-hasunconfirmedemail-attribute)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 76.0.0
 * Remove `has_unconfirmed_email` parameter from Account API calls
 
 # 75.3.0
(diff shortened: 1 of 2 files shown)


barrucadu (commented):
This is technically a breaking change since you've removed a keyword parameter of a public function - nothing was using it (only DI call that endpoint, and they don't use gds-api-adapters), but still it should be a major version bump.

alex9smith:
Yes, good point. I'll update it

barrucadu (approved):

PR #1124, opened 2021-12-24
(WIP) Add send send_unpublish_message to email-alert-api

huwd:
[Trello](https://trello.com/c/mQNmwCo2/1140-send-users-an-email-when-a-single-page-is-unpublished)

WIP - notes and all so we remember what we were doing in the new year.

PR #1125, opened 2022-01-14
Add email-alert-api bulk_unsubscribe endpoint

huwd:
[Trello](https://trello.com/c/mQNmwCo2/1140-send-users-an-email-when-a-single-page-is-unpublished)

Add bulk_unsubscribe endpoint for email-alert-api

Replaces - https://github.com/alphagov/gds-api-adapters/pull/1124
Depends on - https://github.com/alphagov/email-alert-api/pull/1696
Depends on - https://github.com/alphagov/email-alert-api/pull/1694

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -247,6 +247,22 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     )
   end
 
+  # Unsubscribe all users for a subscription list
+  # optionally send a notification email explaining the reason
+  #
+  # @param [string]               subscriber_list_id    Identifier for the subscription list
+  # @param [string] (optional)    body                  Optional email body to send to alert users they are being unsubscribed
+  # @param [string] (optional)    sender_message_id     A UUID to prevent multiple emails for the same event
+  def bulk_unsubscribe(subscriber_list_id:, body: nil, sender_message_id: nil)
+    post_json(
+      "#{endpoint}/subscriber-lists/#{subscriber_list_id}/bulk-unsubscribe",
+      {
+        body: body,
+        sender_message_id: sender_message_id,
+      }.compact,
+    )
+  end
+
 private
 
   def nested_query_string(params)
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -408,6 +408,53 @@ module GdsApi
           .to_return(status: 404)
       end
 
+      def stub_email_alert_api_bulk_unsubscribe(subscriber_list_id:)
+        stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{subscriber_list_id}/bulk-unsubscribe")
+          .to_return(status: 202)
+      end
+
+      def stub_email_alert_api_bulk_unsubscribe_with_message(subscriber_list_id:, body:, sender_message_id:)
+        stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{subscriber_list_id}/bulk-unsubscribe")
+        .with(
+          body: {
+            body: body,
+            sender_message_id: sender_message_id,
+          }.to_json,
+        ).to_return(status: 202)
+      end
+
+      def stub_email_alert_api_bulk_unsubscribe_not_found(subscriber_list_id:)
+        stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{subscriber_list_id}/bulk-unsubscribe")
+          .to_return(status: 404)
+      end
+
+      def stub_email_alert_api_bulk_unsubscribe_not_found_with_message(subscriber_list_id:, body:, sender_message_id:)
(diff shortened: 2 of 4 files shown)


lib/gds_api/email_alert_api.rb
```diff
@@ -247,6 +247,24 @@ def send_subscription_verification_email(address:, frequency:, topic_id:)
...
 
+  # Unsubscribe all users for a subscription list
+  # optionally send a notification email explaining the reason
+  #
+  # @param [string]               slug                  SubscriptionList slug
+  # @param [string] (optional)    body                  Optional email body to send to alert users they are being unsubscribed
+  # @param [string] (optional)    sender_message_id     A UUID to prevent multiple emails for the same event
+  #                                                     generated as a v5 UUID comprising content_id + public_updated_at
+  #                                                     Email API requires this be provided if a body param is present.
+  def bulk_unsubscribe(slug:, body: nil, sender_message_id: nil)
```
barrucadu, line 258: I think call this `subscriber_list_id`, since we use "slug" in so many places and for different things through the publishing system, and also it's what the `subscribe` method uses.

lib/gds_api/email_alert_api.rb
```diff
@@ -247,6 +247,24 @@ def send_subscription_verification_email(address:, frequency:, topic_id:)
     )
   end
 
+  # Unsubscribe all users for a subscription list
+  # optionally send a notification email explaining the reason
+  #
+  # @param [string]               subscriber_list_id    SubscriptionList slug, for example a content_id for single page notifications
```
barrucadu, line 253: > for example a content_id for single page notifications

No, it's the ID of the subscriber list.

huwd: ah of course, sorry got that mixed up with the other one.

lib/gds_api/email_alert_api.rb
```diff
@@ -247,6 +247,24 @@ def send_subscription_verification_email(address:, frequency:, topic_id:)
     )
   end
 
+  # Unsubscribe all users for a subscription list
+  # optionally send a notification email explaining the reason
+  #
+  # @param [string]               subscriber_list_id    Identifier for the subscription list
+  # @param [string] (optional)    body                  Optional email body to send to alert users they are being unsubscribed
+  # @param [string] (optional)    sender_message_id     A UUID to prevent multiple emails for the same event
+  #                                                     generated as a v5 UUID comprising content_id + public_updated_at
```
barrucadu, line 256: I think leave out this line.

email-alert-api doesn't actually care how the UUID is generated, it's a detail which the user of this API needs to care about and this comment currently suggests that this endpoint will *only* work for content ID-based lists, which isn't true.

We talked about v5 UUIDs in the email-alert-api PR because we wanted to make sure there would be a feasible way to come up with a UUID for the events we currently care about (unpublishings which have both a content ID and a public timestamp, but no other useful identifying information), but another approach would work equally well so long as it's consistent (and other subscriber lists could be bulk-unsubscribed so long as the sender_message_id used doesn't conflict).

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -408,6 +408,53 @@ def stub_email_alert_api_does_not_have_subscriber_list_by_slug(slug:)
...
+            sender_message_id: sender_message_id,
+          }.to_json,
+        ).to_return(status: 409)
+      end
+
+      def stub_email_alert_api_bulk_unsubscribe_bad_request(subscriber_list_id:, body:)
+        stub_request(:post, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{subscriber_list_id}/bulk-unsubscribe")
+        .with(
+          body: { body: body }.to_json,
+        ).to_return(status: 400)
```
barrucadu, line 455: This is actually an unprocessable entity, which returns a 422

test/email_alert_api_test.rb
```diff
@@ -808,4 +808,63 @@
       end
     end
   end
+
+  describe "bulk_unsubscribe" do
+    let(:content_id) { SecureRandom.uuid }
```
barrucadu, line 813: `subscriber_list_id`

test/email_alert_api_test.rb
```diff
@@ -808,4 +808,63 @@
...
     end
   end
+
+  describe "bulk_unsubscribe" do
+    let(:subscriber_list_id) { SecureRandom.uuid }
+    let(:public_updated_at) { Time.now }
+    let(:body) { "email message goes here" }
+    let(:sender_message_id) { SecureRandom.uuid }
+
+    it "returns 204 with just a subscriber_list_id" do
```
barrucadu, line 818: `returns 202`

test/email_alert_api_test.rb
```diff
@@ -808,4 +808,63 @@
...
+      stub_email_alert_api_bulk_unsubscribe(subscriber_list_id: subscriber_list_id)
+      api_response = api_client.bulk_unsubscribe(subscriber_list_id: subscriber_list_id)
+      assert_equal(202, api_response.code)
+    end
+
+    it "returns 204 with a message" do
```
barrucadu, line 824: `returns 202`

PR #1131, opened 2022-01-21
Add `update_subscriber_list_details` (for Email Alert API)

huwd:
[Trello](https://trello.com/c/IkYegsOt/1225-update-subscriberlist-titles-on-major-minor-publishing-change)

Allow PATCH updates to subscriber list titles

See also: https://github.com/alphagov/email-alert-api/pull/1704

### DIFF
--- a/lib/gds_api/email_alert_api.rb
+++ b/lib/gds_api/email_alert_api.rb
@@ -267,6 +267,17 @@ class GdsApi::EmailAlertApi < GdsApi::Base
     )
   end
 
+  # Update subscriber list details, such as title
+  # @param [Hash]         params        A hash of detail paramaters that can be updated. For example title.
+  #                                     For allowed parameters see:
+  #                                     https://github.com/alphagov/email-alert-api/blob/main/docs/api.md#patch-subscriber-listsxxx
+  def update_subscriber_list_details(slug:, params: {})
+    patch_json(
+      "#{endpoint}/subscriber-lists/#{slug}",
+      params,
+    )
+  end
+
 private
 
   def nested_query_string(params)
--- a/lib/gds_api/test_helpers/email_alert_api.rb
+++ b/lib/gds_api/test_helpers/email_alert_api.rb
@@ -472,6 +472,27 @@ module GdsApi
         ).to_return(status: 422)
       end
 
+      def stub_update_subscriber_list_details(slug:, params:)
+        stub_request(:patch, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{slug}")
+        .with(body: params.to_json)
+        .to_return(
+          status: 200,
+          body: get_subscriber_list_response(params).to_json,
+        )
+      end
+
+      def stub_update_subscriber_list_details_not_found(slug:, params:)
+        stub_request(:patch, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{slug}")
+        .with(body: params.to_json)
+        .to_return(status: 404)
+      end
+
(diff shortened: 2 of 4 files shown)


test/email_alert_api_test.rb
```diff
@@ -909,4 +909,36 @@
...
+
+      assert_raises GdsApi::HTTPNotFound do
+        api_client.update_subscriber_list_details(slug: slug, params: params)
+      end
+    end
+
+    it "returns 400 when provided with no parameters" do
+      stub_update_subscriber_list_details_bad_request(slug: slug)
+
+      assert_raises GdsApi::HTTPUnprocessableEntity do
```
huwd, line 939: Changd this as we changed it in the reivew on the API

lib/gds_api/email_alert_api.rb
```diff
@@ -267,6 +267,17 @@ def bulk_unsubscribe(slug:, govuk_request_id: nil, body: nil, sender_message_id:
     )
   end
 
+  # Update subscriber list details, such as title
+  # @param [Hash]         params        A hash of detail paramaters that can be updated. For example title.
+  #                                     For allowed parameters see:
+  #                                     https://github.com/alphagov/email-alert-api/blob/main/docs/api.md#patch-subscriber-listsxxx
```
barrucadu, line 273: I weakly feel that you should take the actual parameters here.  It means we have to update gds-api-adapters when those change; but if it's ok in this case to just take an arbitrary hash and say "consult the API docs!" why don't all of the gds-api-adapters methods do that?

huwd: Yeah, I see what you are saying, I came down on the other side of this after a bit of thinking. Partly with one eye to the upcoming description addition.

I figured that there is very little parameter validation going on in the gem. If the API can return a status code telling me what's a valid / useful slug, then it's the APIs job to say what params it's permitting to update.

GDS API adapters is just to act as helpers to facilitate that right?
So as long as there are useful test helpers for what happens if the API rejects you with 422.

Only thing explicitly stating the params here would do is mean we have two places to update / more minor version bumps if we added other params? Can't think of a scenario where that'd help?

huwd: Perhaps especially as if you were sending params that weren't permitted to the api, they'll just be dropped, and it'll error if empty?

I basically don't care what else is in there as long as one valid param exists?

lib/gds_api/test_helpers/email_alert_api.rb
```diff
@@ -472,6 +472,27 @@ def stub_email_alert_api_bulk_unsubscribe_bad_request_with_message(slug:, govuk_
...
+        stub_request(:patch, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{slug}")
+        .with(body: params.to_json)
+        .to_return(status: 404)
+      end
+
+      def stub_update_subscriber_list_details_unprocessible_entity(slug:, params: {})
+        stub_request(:patch, "#{EMAIL_ALERT_API_ENDPOINT}/subscriber-lists/#{slug}")
+        .with(body: params.to_json)
+        .to_return(status: 422)
+      end
```
barrucadu, line 494: These should all have the same default `params` value.

huwd: should they?
For not found, you must have got past the 422 errors, so you will have needed to provide something as params on 484?
Same for a sucessful case.

the stub on 490 is a bit different, you get a 422 from an empty set of body params, but also from invalid ones. I figure as the dev using the stub is choosing to stick this in, they can choose if they want params or not?

isn't it better we return an `ArgumentError` for those first two cases if you don't give params?

barrucadu (approved):

PR #1132, opened 2022-01-24
Three improvements to the email-alert-api test helpers

barrucadu:
Firstly, adding `content_id` to the subscriber lists URL helper.

The way the test helpers handle subscriber lists is kind of weird.  We pass all the parameters---both the request and response ones undifferentiated---into a single `has_subscriber_list` method, which then uses an allowlist to pluck out the parameters which should go in the request URL, and uses the others to construct the response.  This is pretty strange, the only benefit I can see is that some parameters you'd otherwise have to specify twice, but the cost is that we have this list here which has to be kept up to date.  I think it would be better to remove the need for this entirely, but that can be a later change.

Secondly, adding `sender_message_id` and `govuk_request_id` to the bulk-unsubscribe bad-request helper.

These were left off because the only time the endpoint raises a bad request response is if you don't specify these.  But that makes it kind of difficult to actually use in the email-alert-service tests: I *could* blank out the `govuk_request_id` easily enough, but I can't blank out the `sender_message_id` - so in practice this helper can't actually be used.  So I've added those parameters in.

Thirdly, adding both with- and without-message versions of the bulk-unsubscribe helpers.

This is useful because, in cases where we don't care about what the message actually is (like the above case when we're testing a 422 response is gracefully handled), we can use the without-message variants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # unreleased
 
+* BREAKING: Add `sender_message_id` and `govuk_request_id` to bulk unsubscribe (bad request) test helper (for Email Alert API)
 * Add `content_id` to subscriber lists URL helper (for Email Alert API)
 
 # 78.1.0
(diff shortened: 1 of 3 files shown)


huwd (approved):
Great improvements, nice one

PR #1135, opened 2022-02-01
Upgrade Ruby to 2.6.9

ChrisBAshton:
Upgrades Ruby to 2.6.9. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.


ChrisBAshton:
Closing - [Ruby 2.6 is now EOL](https://www.ruby-lang.org/en/downloads/branches/), so we'll be looking to raise a PR to bump this to at least Ruby 2.7.

PR #1136, opened 2022-03-01
Add locations-api to adapter

catalinailie:


lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,53 @@
...
+    response = get_json("#{base_url}/postcode/#{CGI.escape postcode}.json")
+    location = Location.new(response) unless response.nil?
+    location.local_custodian_code
+  end
+
+  def coordinates_for_postcode(postcode)
+    response = get_json("#{base_url}/postcode/#{CGI.escape postcode}.json")
+    location = Location.new(response) unless response.nil?
+    {
+      "latitude" => location.lat,
```
catalinailie, line 15: I'm not sure if this is the format that we want to return.

ChrisBAshton: Putting a `puts response.body` after [this line](https://github.com/alphagov/locations-api/blob/ab071ebcd941089c8e7e63e89dc8f01e63144965/spec/requests/v1/locations_spec.rb#L33) in Locations API, I can retrieve some example JSON that Locations API actually returns:

```
[
  {"postcode":"E1 8QS","address":"1, WHITECHAPEL HIGH STREET, LONDON, E1 8QS","latitude":51.5144547,"longitude":-0.0729933,"local_custodian_code":5900},
  {"postcode":"E1 8QS","address":"2, WHITECHAPEL HIGH STREET, LONDON, E1 8QS","latitude":51.5144548,"longitude":-0.0729934,"local_custodian_code":5900}
]
```

So there are two things I think need changing here:

1. Locations API returns an array. A limitation of Mapit was that it only returned one result; we'll want to return all results, i.e. `[ { "latitude" => ... }, { "latitude" => ...} ]`.
2. Locations API already returns a very flat structure, so there's no need to define the `Location` helper class. We can just return what we need directly, i.e. `{ "latitude" => response[n]["latitude"], ... }` etc.

(It may seem overly complex but several buildings can share the same postcode, so they can and do have different lat/long coordinates and local custodian codes. It'll be up to the frontend applications that use GdsApiAdapters to decide what to do with the array of results).

ChrisBAshton (commented):
Well done for getting to grips with the interplay between GDS API Adapters and Locations API so quickly, @catalinailie!

I've left a few comments below.

I'll also get this in early: we do a _lot_ of rebasing at GDS! Basically, if something was incorrect in, or missing from, an initial commit, we'll go back into that commit and sort it out, rather than push up a 'fix' commit. It takes some getting used to, but the benefits to readability are immense! Read the [GDS Way](https://gds-way.cloudapps.digital/standards/source-code.html#branching-merging-conventions) for more details, and let me know if you'd like a hand your first couple of times rebasing. In the current PR, it looks like there should be 2 commits (rather than 5) in total: one for adding `locations_api` and another one for adding the new test case.

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,39 @@
...
+    unless response.nil?
+      response.each do |r|
+        locations << {
+          "latitude" => r["latitude"],
+          "longitude" => r["longitude"],
+        }
+      end
+    end
+
+    locations
```
ChrisBAshton, line 31: I'd convert the above to a `map` too.

test/locations_api_test.rb
```diff
@@ -0,0 +1,49 @@
+require "test_helper"
+require "gds_api/locations_api"
+require "gds_api/test_helpers/locations_api"
+
+describe GdsApi::LocationsApi do
+  include GdsApi::TestHelpers::LocationsApi
+
+  before do
+    @base_api_url = Plek.current.find("locations_api")
```
ChrisBAshton, line 9:
```suggestion
    @base_api_url = Plek.current.find("locations-api")
```

test/locations_api_test.rb
```diff
@@ -0,0 +1,49 @@
...
+      },
+      {
+        "latitude" => 51.5010097,
+        "longitude" => -0.1415871,
+        "local_custodian_code" => 5901,
+      },
+    ]
+  end
+
+  describe "postcodes" do
```
ChrisBAshton, line 25:
```suggestion
  describe "Locations API" do
```

ChrisBAshton (commented):
Thanks @catalinailie - this is looking great! 🎉  It's close to being ready for merge now.

We'll also need to update the [Changelog](https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md) and [bump the version](https://github.com/alphagov/gds-api-adapters/blob/1051159d9382100a0987863091476dad395814b0/lib/gds_api/version.rb). We use [semver](https://semver.org/), so this would be a minor bump (i.e. 79.1). That can be done in a new commit, as part of this PR.

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,39 @@
...
+    response = get_json("#{base_url}/locations?postcode=#{postcode}.json")
+
+    codes = []
+    unless response.nil?
+      response.each do |r|
+        codes << r["local_custodian_code"]
+      end
+    end
+
+    codes
```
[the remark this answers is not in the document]
catalinailie: I agree, that's much better. 👍

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 79.1.0
+
+* Add Locations API adapter
```
ChrisBAshton, line 3: Minor: I think we could make this more descriptive

```suggestion
* Add `local_custodian_code_for_postcode` and `coordinates_for_postcode` methods for Locations API
```

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "base"
+require_relative "exceptions"
+
+class GdsApi::LocationsApi < GdsApi::Base
+  def local_custodian_code_for_postcode(postcode)
```
ChrisBAshton, line 5: Ah, I've just noticed these two methods are missing RDoc style comments. Please have a look at [how Account API did this](https://github.com/alphagov/gds-api-adapters/blob/19f0f2c6015f2568a202f37aa205ea863c4f47ba/lib/gds_api/account_api.rb#L11-L16) and recreating that kind of comment here.

ChrisBAshton (commented):
Thanks @catalinailie - very close now. I've just spotted something around documentation (apologies I didn't catch it earlier!)

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,30 @@
...
+    return [] if response.nil?
+
+    response.map { |r| r["local_custodian_code"] }
+  end
+
+  # Get a list of coordinates for a postcode
+  #
+  # @param [String, nil] postcode The postcode for which the coordinates are requested
+  #
+  # @return [Array] A list o coordinates (two fields, "latitude" and "longitude") for a specific postcode
```
ChrisBAshton, line 22:
```suggestion
  # @return [Array] A list of coordinates (two fields, "latitude" and "longitude") for a specific postcode
```

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,30 @@
...
+  def coordinates_for_postcode(postcode)
+    response = get_json("#{endpoint}/locations?postcode=#{postcode}.json")
+
+    return [] if response.nil?
+
+    response.map { |r| { "latitude" => r["latitude"], "longitude" => r["longitude"] } }
```
ChrisBAshton, line 28: @brucebolt just thinking back to our handover discussion.  It makes sense for `local_custodian_code_for_postcode` to return multiple custodian codes, as a user's postcode could be split across multiple authorities, so we want to surface that.

But for `coordinates_for_postcode`, am I right in remembering that we wanted to return one set of coordinates, i.e. return the first result in the list, or do some work to approximate the average coordinates across all the results in the response? I'm afraid we didn't specify in the [Trello card](https://trello.com/c/wufIa6zM/2847-add-locations-api-to-gds-api-adapters-3), so just double checking 😅

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,30 @@
...
+  #
+  # @param [String, nil] postcode The postcode for which the custodian codes are requested
+  #
+  # @return [Array] All local custodian codes for a specific postcode
+  def local_custodian_code_for_postcode(postcode)
+    response = get_json("#{endpoint}/locations?postcode=#{postcode}.json")
+
+    return [] if response.nil?
+
+    response.map { |r| r["local_custodian_code"] }
```
ChrisBAshton, line 15: We may want to add a `.uniq` at the end of this, to filter out all duplicate custodian codes. (We'd need to add a test for this too - so can make this change in a separate commit).

Sorry I didn't catch this earlier!

It feels a little awkward introducing logic at the GDS API Adapters layer (rather than in Locations API directly) but I think it makes sense here as we only have the [one route](https://github.com/alphagov/locations-api/blob/593e7589c513743bccccd9193f00242bded2accd/config/routes.rb#L6) in Locations API (so no room for that level of granularity), and there's already a precedent for adding logic in GDS API Adapters e.g. in [content_store.rb](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/content_store.rb).

ChrisBAshton (commented):
Thanks @catalinailie. One more suggestion around the logic in `local_custodian_code_for_postcode`, which is down to a lack of detail on the Trello card - my apologies!

Hopefully this is at least a good practice PR for learning all about rebasing! 😄

lib/gds_api/locations_api.rb
brucebolt: As we didn't want to POST back the user's exact address when a user selects their full address (otherwise we get into PII collection issues), we'll have to use a single set of coordinates for all addresses in the postcode.  For postcodes where there is a single local authority, we'll have to do this anyway.

I don't believe the exact location is too important.  As far as I can see, we don't report the actual distance to users when doing imminence lookups, so an approximation should be suitable to order the list by nearest.

As for choosing the first or taking an average, I think an average (if we can do this) would be better as this should give us roughly the centre of the postcode (assuming the buildings are all evenly distributed).

catalinailie: Thanks @brucebolt for the context. I'll change the method to return the average.

ChrisBAshton: Thanks for clarifying, @brucebolt. I think what we return so far follows the [architecture plan (see step 5)](https://docs.google.com/document/d/1p29g1SQgi2obQnPPsl9amx7xmrJonyH4tRbVMAhXy-A/edit#heading=h.7v5d22d02otc), but it sounds like we'll need a small change.

We probably ought to do the averaging in Locations API (rather than GDS API Adapters), and could return that at the top level of the response, i.e. 1):

```diff
- [ { lat: 1, long: -2, address: "Foo" }, { lat: 2, long: -3, address: "Bar" } ]
+ { lat: 1.5, long: -2.5, [ { lat: 1, long: -2, address: "Foo" }, { lat: 2, long: -3, address: "Bar" } ] }
```

But I think you're saying we need to strip out the coordinates from each address? So it would be more like 2):

```diff
- [ { address: "Foo" }, { address: "Bar" } ]
+ { lat: 1.5, long: -2.5, [ { address: "Foo" }, { address: "Bar" } ] }
```

And if we're doing that, we could just repeat the top-level averages inside each result and keep the same array response structure 3):

```diff
-  { lat: 1.5, long: -2.5, [ { address: "Foo" }, { address: "Bar" } ] }
+ [ { lat: 1.5, long: -2.5, address: "Foo" }, { lat: 1.5, long: -2.5, address: "Bar" } ]
```

Can I check we're on the same page and you're suggesting option 3?

brucebolt: I wonder if we should go with option 1?

This would allow us to increase the scope of Locations API if needed in the future (e.g. we build a new service that requires coordinates for each property) but solves the immediate problem of needing to locate (roughly) the centre of a postcode.

ChrisBAshton: Sounds good to me, so long as there are no PII implications for keeping the coordinates in (I wouldn't have thought there's any increased risk, as we're already returning the full address in the Locations API response).

I'll write up a quick card now 👍

ChrisBAshton: https://trello.com/c/N46nNKYN/2867-return-average-lat-long-coordinates-in-locations-api

lib/gds_api/locations_api.rb
```diff
@@ -0,0 +1,28 @@
...
+class GdsApi::LocationsApi < GdsApi::Base
+  # Get a list of local custodian codes for a postcode
+  #
+  # @param [String, nil] postcode The postcode for which the custodian codes are requested
+  #
+  # @return [Array] All local custodian codes for a specific postcode
+  def local_custodian_code_for_postcode(postcode)
+    response = get_json("#{endpoint}/locations?postcode=#{postcode}.json")
+
+    return [] if response.nil?
```
ChrisBAshton, line 13: We should have a test for this condition (and also for `response.nil?` on line 26 of this file). There's already quite a lot going on in the first commit, so I'd add these `if`/`unless` guards, and their tests, in a second commit.

ChrisBAshton (approved):
Great work on this, @catalinailie 👍 I've left an optional (recommended) comment below, but feel free to merge.

ChrisBAshton (approved):
Nicely done 👏

PR #1137, opened 2022-03-28
Deprecate Email Alert API's create_message endpoint

sihugh:
This was created for the Brexit checker. I don't believe it is used by anything since the checker was retired, so intend to remove it.

https://trello.com/c/8x01txnU/202-brexit-checker-uses-email-alert-api-as-a-substitute-database

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Deprecate `create_message` for Email Alert API
+
 # 79.1.0
 
 * Add `local_custodian_code_for_postcode` and `coordinates_for_postcode` methods for Locations API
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Seems reasonable 👍 

I've a feeling Coronavirus team may be considering using this endpoint for some Coronavirus changes in early April. It's probably worth touching base with @Rosa-Fox to see if they have plans, but shouldn't really stop deprecation as there idea would be one-off I expect.

sihugh:
I'll happily revert this if it has enthusiastic users 👍

leenagupte:
Thanks for checking 😍. We (the Coronavirus team) will need to be able to send bulk emails, but I don't think this change interferes with that, so I think it's fine to deprecate it.

PR #1142, opened 2022-05-17
Write Pact tests for Locations API

catalinailie:
Trello card: https://trello.com/c/d5yUPZYA/2891-set-up-contract-tests-for-locations-api-3

Locations API needs to [meet the standards](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-128-continuous-deployment.md#check-api-contract-tests-pass-ie-adapters-work-with-their-apis) required for continuously deployment.
Contract tests have to be set up between Locations API and its API consumer, gds-api-adaptors.

### DIFF
--- a/test/locations_api_test.rb
+++ b/test/locations_api_test.rb
@@ -4,29 +4,30 @@ require "gds_api/test_helpers/locations_api"
 
 describe GdsApi::LocationsApi do
   include GdsApi::TestHelpers::LocationsApi
-
-  before do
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM ⭐ nice one!
Just needs the corresponding tests approved in Locations API before merge. https://docs.publishing.service.gov.uk/manual/pact-testing.html#adding-tests-for-a-new-app

PR #1145, opened 2022-06-08
Remove attribute types from account API

alex9smith:
We are simplifying account API and removing the `local` attribute type.

This means it will only store what were previously `cached` attributes
(those provided by the IDP), and therefore aren't allowed to be changed
by any GOV.UK app.

Remove the pact test that tries to set an attribute in preparation for
this.

---

[JIRA](https://govukverify.atlassian.net/jira/software/c/projects/GUA/boards/195?modal=detail&selectedIssue=GUA-122)

### DIFF
--- a/test/account_api_test.rb
+++ b/test/account_api_test.rb
@@ -279,7 +279,7 @@ describe GdsApi::AccountApi do
       let(:path) { "/api/attributes" }
 
       describe "#get_attributes" do
-        let(:attribute_name) { "local_attribute" }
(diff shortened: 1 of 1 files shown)


alex9smith:
You can see the pact tests passing against the updated branch of account API here: https://ci.integration.publishing.service.gov.uk/job/gds-api-adapters/job/GUA-122_remove_account_api_attribute_types/3/

huwd (approved):

PR #1146, opened 2022-06-08
Don’t require `govuk_request_id` in Email Alert API's #bulk_unsubscribe stubs

AgaDufrat:
The GOVUK-Request-Id HTTP header is set by Nginx on an initial request and
passed by the apps to the APIs that they call [[1]](https://github.com/alphagov/govuk-puppet/blob/05af1/modules/nginx/templates/proxy-vhost.conf#L7-L25). The gds-api-adapters handle
the header in the middleware [[2]](https://github.com/alphagov/gds-api-adapters/blob/2bb3a/lib/gds_api/railtie.rb#L5-L8).
It is only necessary to explicitly pass the header for out-of-band processes
via asynchronous workers (e.g. Sidekiq).

This still allows to pass it in as a parameter so it's not a breaking change
and also updates the documentation to more accurately describe when the param
is required.

This is only changing the test helpers because the methods doesn't require it already:
```
def bulk_unsubscribe(slug:, govuk_request_id: nil, body: nil, sender_message_id: nil)
```

https://trello.com/c/CQlLYnrF/1038-automate-unsubscribing-from-archived-specialist-topics-m

**Note**
There is a question if setting the header by passing it in the request body should be supported by the gds-api-adapters, given they read the headers in the middleware. There is precedent of the setting the onward GOVUK-Request-Id HTTP header by the apps.
Removing this option would be a breaking change so we may want to consider this as a follow up PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 80.0.1
+
+* Fix `bulk_unsubscribe` stubs so `govuk_request_id` is not a required parameter
+
 # 80.0.0
 
 * BREAKING: Remove support for publishing api import endpoint - a temporary endpoint that has been removed from publishing api. However, no apps are using this so it should not be breaking in practice.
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM (and I learnt about Object#tap, nice).

Tetrino (approved):

PR #1147, opened 2022-06-20
Remove support for the import endpoint

BeckaL:
This endpoint is being removed from publishing api (it was only ever intended to be temporary) so this PR also removes support for it in gds-api-adaptors.

Related publishing api change: https://github.com/alphagov/publishing-api/pull/2096

I've made this a major change as technically it's a breaking change, although no clients are using this so it should not actually break for anybody in practice.

The main question I have is around what the general release process is. I notice that some changes go in as unreleased, but I assume we want to release this one straightaway. Is there anything else I need to bear in mind given this library is widely used across gov.uk and there are some unreleased changes from before?

[Trello card](https://trello.com/c/FNtolpuM/13-enable-continuous-deployment-for-publishing-api)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 80.0.0
+
+* BREAKING: Remove support for publishing api import endpoint - a temporary endpoint that has been removed from publishing api. However, no apps are using this so it should not be breaking in practice.
+
 # 79.2.0
 
 Remove cookie and feedback consent attributes from Account API pact tests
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 80.0.0
+
+* BREAKING: Remove support for import endpoint - this was a temporary endpoint.
```
JonathanHallam, line 3: I'd maybe add a note here to say that no apps were using it, just to make it clearer that this update should be good for all apps

JonathanHallam (approved):
Left a small non blocking comment. The release process is pretty straightforward. Because you've changed the version, ruby gems will pick up on this change and pull it in. Once it's there dependabot will raise a load of PR's in all our apps and people will have to review them

PR #1148, opened 2022-06-21
Initial pact test setup for email-alert-api

KludgeKML:
Pact tests for anything in email-alert-api that is a) called by more than one app, or b) called by only one app, but that app is not owned by Interaction and Personalisation.

TODOs in the pact test file represent the tail of endpoints called by only a single app which is under Interaction and Personalisation ownership (the current owners of email-alert-api). These are to be added in a follow-up card 

https://trello.com/c/YZ1pEayS/1331-add-contract-tests-to-email-alert-api

Pact provider branch: https://github.com/alphagov/email-alert-api/tree/configure-pact-tests-provider

Provider PR: https://github.com/alphagov/email-alert-api/pull/1762

### DIFF
--- a/test/email-alert-api/email_alert_api_pact_test.rb
+++ b/test/email-alert-api/email_alert_api_pact_test.rb
@@ -1,0 +1,544 @@
+require "test_helper"
+require "gds_api/email_alert_api"
+require "gds_api/test_helpers/email_alert_api"
+
+describe GdsApi::EmailAlertApi do
+  include GdsApi::TestHelpers::EmailAlertApi
+  include PactTest
+
+  let(:api_client) { GdsApi::EmailAlertApi.new(email_alert_api_host) }
+
+  let(:example_subscriber) do
+    {
+      id: Pact.like(1),
+      address: "test@example.com",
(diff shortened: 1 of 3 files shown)


1pretz1 (commented):
Nice work digging into the various endpoints and pulling this together Keith, looks great! I also think it's a good idea to go wider than the CD remit and provide a comprehensive pact test suite 👍 

I think we're missing the [config to run the tests via Jenkins CI](https://github.com/alphagov/gds-api-adapters/blob/ddb49a487f5c8b5e28f74b81d98660fb2c02d98d/Jenkinsfile#L11-L63) which we'll need to add.

Secondly (a bit more subjective). I don't think commits 2 -> 5 add much and can probably be merged into commit 1. However, commit 1 is quite large and could probably be split up a bit as it includes a few distinct changes: config (pact_helper), pact tests themselves, future pact tests and a file rename / stub method updates. This isn't anything major but would've helped me review the work a tad easier! 🙂

1pretz1 (approved):
Looks good! We've agreed offline to enable Jenkins CI in another PR.

PR #1156, opened 2022-07-26
Add local_custodian_code and country_name as optional parameters for local links manager stubs

catalinailie:
These are needed for testing the switch from Mapit to Locations API in frontend controllers.

Trello card: https://trello.com/c/6TO6GDJM/2934-switch-frontend-to-use-locations-api-5

### DIFF
--- a/lib/gds_api/test_helpers/local_links_manager.rb
+++ b/lib/gds_api/test_helpers/local_links_manager.rb
@@ -98,13 +98,13 @@ module GdsApi
         parameters
       end
 
-      def stub_local_links_manager_has_a_local_authority(authority_slug, local_custodian_code: nil)
+      def stub_local_links_manager_has_a_local_authority(authority_slug, country_name: "England", local_custodian_code: nil)
         response = {
           "local_authorities" => [
             {
               "name" => authority_slug.capitalize,
               "homepage_url" => "http://#{authority_slug}.example.com",
-              "country_name" => "England",
+              "country_name" => country_name,
               "tier" => "unitary",
               "slug" => authority_slug,
             },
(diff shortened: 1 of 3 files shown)


lib/gds_api/test_helpers/local_links_manager.rb
```diff
@@ -98,13 +98,13 @@ def convert_to_query_string_params(parameters)
         parameters
       end
 
-      def stub_local_links_manager_has_a_local_authority(authority_slug, local_custodian_code: nil)
+      def stub_local_links_manager_has_a_local_authority(authority_slug, country_name: "England", local_custodian_code: nil)
```
ChrisBAshton, line 101: Commit message slightly inaccurate - suggests that `local_custodian_code` is being added to the above method, when it's being added to the `stub_local_links_manager_has_a_local_authority_without_homepage` method. Could either split these additions out into two commits, or tweak the message.

lib/gds_api/locations_api.rb
```diff
@@ -12,7 +12,8 @@ def local_custodian_code_for_postcode(postcode)
 
     return [] if response["results"].nil?
 
-    response["results"].map { |r| r["local_custodian_code"] }.uniq
+    # The local custodian code 7655 is Ordnance Survey added data, so it should be ignored
+    response["results"].map { |r| r["local_custodian_code"] }.reject { |code| code == 7655 }.uniq
```
ChrisBAshton, line 16: Ooh, this looks interesting. I don't really understand the reason behind this change - could you add more detail to the commit?

E.g.

- Where is this Ordnance Survey data being added?
  - For which postcode(s)?
  - Why is it being added? What does it mean?
  - Why do we need to ignore it?
- How did you discover it? Are there potentially other codes we'll need to filter out? (Might affect the implementation)

ChrisBAshton (commented):
I have one nitpick comment and one major comment below 👀 thanks!

lib/gds_api/locations_api.rb
catalinailie: Actually I was just looking over a pdf from ordnance survey. There is also the code 11 for HIGHWAYS ENGLAND. Now I'm not sure if this should be removed here, or in locations-api where we store the OSApi response 🤔

ChrisBAshton: Agreed, sounds like something best handled in Locations API itself.

ChrisBAshton (approved):

PR #1157, opened 2022-08-01
Update subs for locations_api and local_links_manager

catalinailie:
These changed are needed in testing the switch from `mapit` to `locations_api` in `frontend`. https://github.com/alphagov/frontend/pull/3298

### DIFF
--- a/test/local_links_manager_api_test.rb
+++ b/test/local_links_manager_api_test.rb
@@ -45,24 +45,25 @@ describe GdsApi::LocalLinksManager do
 
       it "returns the local authority details only if no link present" do
         stub_local_links_manager_has_no_link(
-          authority_slug: "blackburn",
-          lgsl: 2,
-          lgil: 4,
+          authority_slug: "westminster",
+          lgsl: 461,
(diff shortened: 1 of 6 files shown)


test/local_links_manager_api_test.rb
```diff
@@ -66,6 +66,30 @@
         assert_equal expected_response, response.to_hash
       end
 
+      it "returns the local authority details only if no link present for westminster" do
```
ChrisBAshton, line 69: Nitpick: I think we can delete this test - we're not really testing logic for Westminster AFAIK, we're just checking that the provided `snac` code comes back in the response.

We could achieve the same thing by updating the test at line 46 instead, swapping `blackburn` for `westminster` and passing the custom `snac` code 🤔

ChrisBAshton (approved):
Ah I hadn't clocked that my review comments would mean a change to gds-api-adapters! Sorry for creating extra work 😅 

Feel free to bump the version in this PR at the same time.

One non-blocking comment below.

PR #1158, opened 2022-08-03
Update local_links_manager stub for local_authority to return snac value

catalinailie:
`snac` is now returned in the `/local_authority` response (added [here](https://github.com/alphagov/local-links-manager/pull/939))


ChrisBAshton (approved):
LGTM 👍

PR #1159, opened 2022-08-18
Some TLC for this projects development gems

kevindew:
This PR started as a quick one to turn on dependabot for this project and then grew a little into resolving problems with a few of the gems.

It principally: bumps development gems and drops/replaces some unnecessary ones. It also turns on PACT_DO_NOT_TRACK on to stop the annoying alerts in test environment.

These changes only affect the development gems so these should all be utilised as part of a passing build.


gds-api-adapters.gemspec
```diff
@@ -21,24 +21,21 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency "minitest-around", "~> 0.5"
-  s.add_development_dependency "mocha", "~> 1.11"
-  s.add_development_dependency "pact", "~> 1.20"
-  s.add_development_dependency "pact_broker-client", "~> 1.14"
+  s.add_development_dependency "mocha", "~> 1.14"
+  s.add_development_dependency "pact", "~> 1.62"
+  s.add_development_dependency "pact_broker-client", "~> 1.65"
   s.add_development_dependency "pact-consumer-minitest", "~> 1.0"
-  s.add_development_dependency "pact-mock_service", "~> 3.6"
-  s.add_development_dependency "pry"
```
BeckaL, line 33: I expect you have a solid rationale for removing this, so probably more out of interest for me - isn't there a case for leaving pry in despite it not being used, to aid development?

BeckaL (approved):
Just a question from me about pry that I expect you have a solid reason for, otherwise all looks good to me, another nice review because of nice commit messages!

gds-api-adapters.gemspec
kevindew: I removed it because it's not referenced at all and there's no clear convention in gems for having a debugger or which one (some have byebug, some have pry-byebug etc). Not hugely opinionated on whether it should stay or go to be honest - sorry not much of a rationale 😄

BeckaL: We wouldn't expect it to be referenced though would we? I.e. just used temporarily for development. Unless there's another step to enabling it for development beyond just this? If not, I'd maybe be tempted to keep it in as I suspect someone will re-add it at some point (although noted no consensus). But also not strongly opinionated either way.

kevindew: Yup, that's reasonable. I'll switch it to byebug, as that's at least the one we reference in https://docs.publishing.service.gov.uk/manual/setting-up-new-rails-app.html

kevindew: Added in https://github.com/alphagov/gds-api-adapters/pull/1159/commits/0a6ed791ba7b7d774588eb26af1c5b595c586805

PR #1160, opened 2022-09-07
Remove email-alert-api#create_message

KludgeKML:
This api call has been marked as deprecated for 6 months, and it appears that there are no longer any calls to it in non-archived repos, so we should remove the code here (where it's easy to replace if it turns out there's a surprise user), then if nothing turns up in a couple of weeks we can remove the code supporting the call from email-alert-api itself.

### DIFF
--- a/test/test_helpers/email_alert_api_test.rb
+++ b/test/test_helpers/email_alert_api_test.rb
@@ -22,20 +22,6 @@ describe GdsApi::TestHelpers::EmailAlertApi do
     end
   end
 
-  describe "#assert_email_alert_api_message_created" do
-    before { stub_any_email_alert_api_call }
-
-    it "matches a post request with any empty attributes by default" do
-      email_alert_api.create_message("foo" => "bar")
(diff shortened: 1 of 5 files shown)


test/test_helpers/email_alert_api_test.rb
```diff
@@ -22,20 +22,6 @@
     end
   end
 
-  describe "#assert_email_alert_api_message_created" do
-    before { stub_any_email_alert_api_call }
-
-    it "matches a post request with any empty attributes by default" do
-      email_alert_api.create_message("foo" => "bar")
-      assert_email_alert_api_message_created
```
edwardkerry, line 30: Can we also remove this method declaration?
https://github.com/alphagov/gds-api-adapters/blob/54fed7b96da31a22b565965e03ac063a84a81cd3/lib/gds_api/test_helpers/email_alert_api.rb#L183

edwardkerry (approved):

PR #1165, opened 2022-10-03
Use TLS when fetching gem dependencies.

sengi:
This has been the default for new Gemfiles since a major security incident at rubygems.org in 2013[[1]].

[1]: https://venturebeat.com/business/rubygems-org-hacked-interrupting-heroku-services-and-putting-millions-of-sites-using-rails-at-risk/


KludgeKML (approved):
LGTM

PR #1167, opened 2022-10-03
Replace usages of Plek.current

kevindew:
This has been deprecated [1] since Plek 4.1 [2]. Instead we can use forego the need for a current method and call find directly.

[1]: https://github.com/alphagov/plek/pull/94
[2]: https://github.com/alphagov/plek/pull/92


KludgeKML (approved):
Yep, looks okay to me.

PR #1170, opened 2022-10-20
Use canonical name for `search-api` instead of legacy `search`.

sengi:
This is a breaking change because of the way the test helpers currently work, but I have PRs to fix all the occurrences of the legacy `search` name throughout alphagov ready to go once this is released, so it should be seamless.

Cleaning this up now means there's one less thing to trip over for Replatforming.


sengi:
Turns out this needs a Puppet change as well. alphagov/govuk-puppet#11870

PR #1174, opened 2022-11-01
Untangle pact tests from unit tests

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This PR organises the test files of this gem so that the pact tests and unit tests are separated, with the pact tests put in their own directory (`pacts`) with a suffix of pact_test.rb.

The motivation for doing this change was experimenting with GitHub Actions and the need to generate and publish pact files. I discovered that in some places pact tests were mixed into unit test files, others were distinct with a `*_pact_test.rb` so it was extremely hard to determine which ones were which.

Moving them to a distinct directory means that it'll be easy to have a rake test task that just runs them. e.g `rake pact_test`.

The pact tests are quite different to unit tests as these are mostly used to generate files that other applications use in contract tests and do very little asserting of gds_api_adapter behaviour.

This also does a few cleanups - it breaks up the 2000 line Publishing API pact test it into individual files per endpoint, it resolves various consistency issues.

This PR isn't as grim as the diff suggests I promise 😄 It's just moving code around and doesn't have logic changes. It's easiest reviewed commit by commit.


KludgeKML (approved):
This looks good to me, very nice bit of tidying. :)

PR #1182, opened 2023-01-04
Stop expecting internal Mongo document ID in return from imminence.

KludgeKML:
Imminence shouldn't be returning OID (the internal Mongo document ID) through the API endpoint. Frontend (currently the only consumer of this endpoint) doesn't appear to be using the OID. I think technically this isn't a breaking change, since it doesn't materially affect the API, even though it's a pact change.


1pretz1 (approved):
Nice catch!

PR #1196, opened 2023-05-23
v88.0: remove no-op "commit" option from Router.

sengi:
This option has had no effect for >2 years. See https://github.com/alphagov/router-api/pull/550

Strictly speaking this is a breaking change, but all of the affected call sites are in content_store so I'll just [fix them](https://github.com/alphagov/content-store/pull/1076) and then it'll be all done.


theseanything (approved):

sengi:
Rewrote the release note to make it a bit more precise.

PR #1202, opened 2023-07-14
Add govuk_request_id to logs when available

tunylund:
This pr aims to fix an issue where the gds-api-adapters logs output whenever an error has happened but the output is missing `govuk_request_id`

image

---

Whenever an error is raised by the api library, the output from the api client now contains govuk_request_id whenever its available.

gds-api-adapters by default sets up a header sniffer for govuk_request_id here: https://github.com/alphagov/gds-api-adapters/blob/main/lib/gds_api/railtie.rb#L7 which allows the code below to just add the header value into log lines.

I looked into if we could make govuk_app_config add the header to these log lines as it does something similar for the default Rails logger, but it looks like the most straightforward way is this.

### DIFF
--- a/lib/gds_api/json_client.rb
+++ b/lib/gds_api/json_client.rb
@@ -157,7 +157,7 @@ module GdsApi
     end
 
     def do_request(method, url, params = nil, additional_headers = {})
-      loggable = { request_uri: url, start_time: Time.now.to_f }
+      loggable = { request_uri: url, start_time: Time.now.to_f, govuk_request_id: GdsApi::GovukHeaders.headers[:govuk_request_id] }.compact
       start_logging = loggable.merge(action: "start")
       logger.debug start_logging.to_json
 
--- a/test/json_client_test.rb
+++ b/test/json_client_test.rb
@@ -613,6 +613,22 @@ class JsonClientTest < MiniTest::Spec
     assert_same client.logger, custom_logger
   end
 
+  def test_should_log_govuk_request_id_when_available
+    GdsApi::GovukHeaders.set_header(:govuk_request_id, "some-request-id")
+    custom_logger = mock
+    client = GdsApi::JsonClient.new(logger: custom_logger)
+    url = "http://www.example.com/timeout.json"
+    stub_request(:get, url).to_timeout
+
+    expected_string = "\"govuk_request_id\":\"some-request-id\""
+    custom_logger.expects(:debug).with(includes(expected_string))
+    custom_logger.expects(:error).with(includes(expected_string))
+
+    assert_raises do
+      client.get_json(url)
+    end
+  end
+
   def test_should_avoid_content_type_header_on_get_without_body
     url = "http://some.endpoint/some.json"
     stub_request(:any, url)


lib/gds_api/json_client.rb
```diff
@@ -157,7 +157,7 @@ def with_ssl_options(method_params)
     end
 
     def do_request(method, url, params = nil, additional_headers = {})
-      loggable = { request_uri: url, start_time: Time.now.to_f }
+      loggable = { request_uri: url, start_time: Time.now.to_f, govuk_request_id: GdsApi::GovukHeaders.headers[:govuk_request_id] }.compact
```
kevindew, line 160: Should we be including all of the headers in GdsApi::GovukHeaders with the logged item? e.g more like: GdsApi::GovukHeaders.headers.merge({ request_uri: url, start_time: Time.now.to_f })I know for the purposes of your particular task just govuk_request_id is of interest, but I'd thought if an app had gone to the effort of setting these headers it feels like it should be part of the debug output.

kevindew (commented):
Seems reasonable to include.

I assume you're aware and happy that it won't be a top level field in the resulting log (and be part of a dump of the message)

sengi (approved):

kevindew:
> I'd be somewhat cautious about logging all `GovukHeaders` headers rather than specific ones. I did a quick search for other headers that are inserted using that API: https://sourcegraph.com/search?q=repo%3Aalphagov+GovukHeaders&case=yes and right away I see `x_govuk_authenticated_user`, which sounds like something we might not necessarily want in info logs.
> 
> We also probably don't want to log new headers added in the future without anyone even necessarily knowing that that'll happen.

I get what you mean and would agree if it was regular info/warn logging, however this is for debug level logging so I'd expect pretty much everything at that granularity - nor would I expect people to be running that in prod by default.

sengi:
Ah ok, that's fair. From the PR description it doesn't look like debug logging though.

sengi:
The screenshot is from Logit.io in fact.

kevindew:
Ah you're right, it is only debug on the line in the preview the others are warn and error. Sorry my bad

As this uses the Rails logger, does this https://github.com/alphagov/govuk_app_config/blob/e5edc331726e1a0668c50eff53646d2b8534dc47/lib/govuk_app_config/govuk_json_logging.rb#L20-L22 not also add this in?

kevindew (approved):

kevindew:
> > Ah you're right, it is only debug on the line in the preview the others are warn and error. Sorry my bad
> > As this uses the Rails logger, does this https://github.com/alphagov/govuk_app_config/blob/e5edc331726e1a0668c50eff53646d2b8534dc47/lib/govuk_app_config/govuk_json_logging.rb#L20-L22 not also add this in?
> 
> Unfortunately not. `govuk_api_adapters` configures api-adapters to utilise logstasher logger, not Rails.logger. https://github.com/alphagov/govuk_app_config/blob/e5edc331726e1a0668c50eff53646d2b8534dc47/lib/govuk_app_config/govuk_json_logging.rb#L70
> 
> It would be nice if Rails.logger would be used, but in order to do that, we would have to implement the `JSON.parse` into the Rails.logger. As per the discussion in [this PR](https://github.com/alphagov/govuk_app_config/pull/300) - there was a desire to avoid JSON.parse in the formatter logic - so I decided to implement this pr.

Oh sure, I didn't even know we could use logstasher as just a regular logger .

Yeah seems reasonable about the JSON. Of course there had to be a second usage of JSON logging hidden away. If we find more maybe we'll have to revisit the JSON.parse but hopefully not as seems just more layers of complexity.

kevindew:
Sorry - closed by mistake

tunylund:
I merged this already, but just realized, if we would implement JSON.parse into Rails.logger, then we wouldn't need this pr at all.

kevindew:
> I merged this already, but just realized, if we would implement JSON.parse into Rails.logger, then we wouldn't need this pr at all.

Sure I can understand that though I still think it's best that that isn't the approach myself - particularly given that if a field is already used in logit with a different format the log will just get lost. It makes me feel quite uneasy having such free form log structure when the consequences of a naming collision is quite severe.

I imagine it'd be nicest if we had just a limited set of root level fields and, for examples like this where you want to log JSON, an ability for that to be logged under a namespace (i.e. for this gds_api_adapters) so that it wasn't just putting everything at root fields ready to collide.

sengi:
Yeah it's a real pain that ELK is so fussy about schemas; it'd be so much less of a hassle if it erred on the side of accepting whatever's thrown at it and just warning that it doesn't match the schema or whatever. But we've mostly just gotta live with that.

We could probably improve things a bit by trying to align with the [common schema](https://www.elastic.co/guide/en/ecs/current/index.html) (if you can't beat 'em, join 'em), but it's one of those things that could sink a lot of up-front effort for little real-world gain — so maybe we're better off just fixing these things as we encounter real-world use cases like we've been doing? 🤷

kevindew:
> Yeah it's a real pain that ELK is so fussy about schemas; it'd be so much less of a hassle if it erred on the side of accepting whatever's thrown at it and just warning that it doesn't match the schema or whatever. But we've mostly just gotta live with that.
> 
> We could probably improve things a bit by trying to align with the [common schema](https://www.elastic.co/guide/en/ecs/current/index.html) (if you can't beat 'em, join 'em), but it's one of those things that could sink a lot of up-front effort for little real-world gain — so maybe we're better off just fixing these things as we encounter real-world use cases like we've been doing? 🤷

Ooh I didn't know about that common schema - thanks for sharing. I guess we've got that by default applied, as looking that [error](https://www.elastic.co/guide/en/ecs/current/ecs-error.html) one in that explains where the schema appeared that nuked our field that was setting error as a string. Would agree on time vs gain - would certainly be nice and elegant but not really impact any person's workflow (and probably have a whole bunch of stuff that is hard to slot in).

The thing that bugs me about it is how hard it is it fix a case if something is wrong. Seems once you've got misindexed field type you've got to be prepared to throw the whole lot in the bin to change it 🙀

sengi:
The schemas are versioned so at least in theory it should never force you to delete old logs just to make a schema change, but it certainly drops stuff on the floor on ingestion if it doesn't perfectly match what it's expecting :(

PR #1203, opened 2023-07-18
Update data source for World Location methods

brucebolt:
These methods currently returns data retrieved from the Whitehall API, which is being deprecated.

Therefore updating the method to get the data from the World Content Item.

We need to transform the response to ensure the output remains as backwards compatible with previous versions of this gem as possible.  However a breaking change has been introduced, as pagination of the World Locations endpoint has been removed.

The following applications use the affected methods and will be updated in the following PRs (once this a new version of this gem has been released):
- https://github.com/alphagov/smart-answers/pull/6463
- https://github.com/alphagov/finder-frontend/pull/3113
- https://github.com/alphagov/contacts-admin/pull/1204

[Trello card](https://trello.com/c/4r5uSD2R)

### DIFF
--- a/lib/gds_api/worldwide.rb
+++ b/lib/gds_api/worldwide.rb
@@ -2,7 +2,7 @@ require_relative "base"
 
 class GdsApi::Worldwide < GdsApi::Base
   def world_locations
-    get_list("#{base_url}/world-locations")
+    all_world_locations
   end
 
   def world_location(location_slug)
--- a/lib/gds_api/worldwide.rb
+++ b/lib/gds_api/worldwide.rb
@@ -6,7 +6,13 @@ class GdsApi::Worldwide < GdsApi::Base
   end
 
   def world_location(location_slug)
-    get_json("#{base_url}/world-locations/#{location_slug}")
+    world_location = all_world_locations.find do |location|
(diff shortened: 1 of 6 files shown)


lib/gds_api/worldwide.rb
```diff
@@ -2,11 +2,17 @@
 
 class GdsApi::Worldwide < GdsApi::Base
   def world_locations
-    get_list("#{base_url}/world-locations")
+    all_world_locations
   end
 
   def world_location(location_slug)
-    get_json("#{base_url}/world-locations/#{location_slug}")
+    world_location = all_world_locations.find do |location|
```
CristinaRO, line 9: This might not be a real source of worry, but how efficient is it to load all world locations every time someone (or something) requests a single world location? Is there some kind of implicit caching?

test/pacts/worldwide_api_pact_test.rb
```diff
@@ -6,85 +6,7 @@
 
   let(:api_client) { GdsApi::Worldwide.new(whitehall_api_host) }
 
-  describe "#world_locations" do
```
CristinaRO, line 9: What replaces these tests?

CristinaRO (approved):

PR #1204, opened 2023-07-27
Change data source for `organisations_for_world_location` method

brucebolt:
This ends reliance on the Whitehall API by changing the data source to use Search API and Content Store.
    
The output is in an identical format, but is no longer paginated.

[Trello card](https://trello.com/c/jJ0SH0N3)


jkempster34 (approved):
This looks good to me. 

Out of curiosity, did you test to see how much longer the request takes now that we need to fetch multiple content items.

I'm also wondering if it's worth highlighting in the commit messages that we had to make multiple requests because thats the only sensible option at this time?

brucebolt:
> Out of curiosity, did you test to see how much longer the request takes now that we need to fetch multiple content items.

It takes around 3-4 times longer to make these requests.  However I noticed the only application to use this method (Smart Answers) caches the response anyway, so the number of requests should be very minimal.



> I'm also wondering if it's worth highlighting in the commit messages that we had to make multiple requests because thats the only sensible option at this time?

Nice idea.  I'll add that in.

PR #1206, opened 2023-07-31
Update casing of `MiniTest::Spec` and update dependencies

brucebolt:
This was incorrectly included as `MiniTest` when it should be `Minitest`, causing CI to fail, due to a removal of the compatibility layer in `minitest` 5.19.0.  Also requires `mocha` 2.1, as this in the earliest release to not require the compatibility layer.


kevindew (approved):

kevindew:
As I understand Mini Test recently dropped the casing. I wonder if we might need to specify the minimum Mini Test version in the gemspec or if the MiniTest change was backwards compatible (assuming usage of `Minitest`)

PR #1211, opened 2023-09-12
Add `media` method to Asset Manager

syed-ali-tw:
Asset Manager has a `media` route, which returns the actual asset for a given asset-manager-id and the filename.

Adding this method to gds-api-adapters so it can be used in the Frontend application.

[Trello Card](https://trello.com/c/LV4TnjlP/176-story-csv-previews-should-work-with-modern-urls)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+* Add `media` method to Asset Manager.
+
 # 91.0.0
 
 * BREAKING: Remove Account API `get_email_subscription` method and helpers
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 92.0.0
```
brucebolt, line 1: This change should probably be release `91.1.0`, since there's no breaking or incompatible changes being introduced.

Also [we recommend](https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html#versioning) that the version be bumped in a separate PR, to avoid any versioning conflicts if another release is made by someone else in the meantime.

brucebolt (approved):

PR #1231, opened 2024-01-30
Remove step to delete pact artifacts

brucebolt:
We are using a third-party plugin here, which requires access to a secret token and also appears to be flakey.

Therefore removing the step completely, as:
1. the artifact is tiny (~15KB in this application) and we have 50GB of storage across the organisation
2. GitHub delete old artifacts after 90 days, or we can configured a smaller amount

[Trello card](https://trello.com/c/ngFoCxYX)


brucebolt:
The Ruby test build is currently failing for the same reason as given in [this comment](https://github.com/alphagov/gds-api-adapters/pull/1230#issuecomment-1916549825).

jkempster34 (approved):

PR #1248, opened 2024-04-05
Added a new stub for attempting to delete a missing asset

ryanb-gds:
This is needed for testing how publishing applications react to attempting to delete a missing asset. At the moment some of them are not handling not found errors, meaning that the state of the publishing application remains out of sync with asset manager. This causes other logical issues such as being unable to discard a draft in Specialist Publisher because there is an attachment that can't be deleted.

This commit also adds some test coverage for other asset deletion stubs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+* Added stub for attempting to delete a missing asset from Asset Manager
+
 # 95.0.1
 * Update Pact specs to match the publishing-api [PR](https://github.com/alphagov/publishing-api/pull/2669)
 
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -92,6 +92,11 @@ module GdsApi
           .to_return(body: body.to_json, status: 200)
       end
 
+      def stub_asset_manager_delete_asset_missing(asset_id)
+        stub_request(:delete, "#{ASSET_MANAGER_ENDPOINT}/assets/#{asset_id}")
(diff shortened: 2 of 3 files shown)


test/test_helper.rb
```diff
@@ -17,6 +17,7 @@
 require "gds-api-adapters"
 require "govuk_schemas"
 require "climate_control"
+require "byebug"
```
richardTowers, line 20: 👀

ryanb-gds: Is there a better way to do this? I couldn't debug my test without it, thought others would probably like to be able to debug the tests as well

callumknights: Generally we don't commit things like this to main, some people have different local dev setups etc. As long as you can verify everything passes locally, remove anything you have added for local testing and commit. As long as CI passes then all is fine 👍

ryanb-gds: Done

richardTowers: Ah yeah, sorry - I assumed it was a mistake rather than a deliberate addition. I wouldn't object to it being added to testhelper deliberately I don't think - just seemed unrelated to the PR

brucebolt (commented):
Code looks good to me.  Could you add an entry to the `Unreleased` section of the `CHANGELOG`, since you'll need to release a new version of this gem (in another PR) to make the stub available to other applications.

callumknights:
I'm happy with the code here as well. I'm inclined to agree with you that this is probably a minor release, I initially misread the PR description and thought this was a bug fix but that doesn't appear to be the case.

callumknights (approved):

PR #1256, opened 2024-05-07
Publish pacts for branches other than main

hannako:
In https://github.com/alphagov/gds-api-adapters/pull/1188 we set a restriction to only run `rake:publish_pacts` on builds of the main branch. This was not intentional.

Once this change is merged, it should mean that we can test providers locally against branches of gds-api-adapters as per the [developer docs](https://docs.publishing.service.gov.uk/manual/pact-testing.html#changing-existing-pact-tests). E.g. `env PACT_CONSUMER_VERSION=branch-<branch-of-gds-api-adapters> bundle exec rake pact:verify`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       contents: read
       security-events: write
       actions: read
-  
+
   codeql-sast:
     name: CodeQL SAST scan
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
Nice!

kevindew (approved):
Looks good

PR #1273, opened 2024-06-05
Add Pact tests for Support API

deborahchua:
This follows the steps covered in the [Pact testing Developer docs](https://docs.publishing.service.gov.uk/manual/pact-testing.html) and has been tested locally.

Support API Pact provider branch: [add-pact-tests](https://github.com/alphagov/support-api/compare/add-pact-tests)

Provider PR: https://github.com/alphagov/support-api/pull/964

Trello card: https://trello.com/c/QBYirV6S/3494-add-pact-test-for-create-support-tickets-endpoint-5

### DIFF
--- a/test/pacts/support_api_pact_test.rb
+++ b/test/pacts/support_api_pact_test.rb
@@ -1,0 +1,67 @@
+require "test_helper"
+require "gds_api/support_api"
+
+describe "GdsApi::SupportApi pact tests" do
+  include PactTest
+
+  describe "#raise_support_ticket" do
+    let(:api_client) { GdsApi::SupportApi.new(support_api_host) }
+
+    it "responds with a 201 Success if the parameters provided are valid" do
+      support_api
+        .given("the parameters are valid")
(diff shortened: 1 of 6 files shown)


test/pacts/support_api_pact_test.rb
```diff
@@ -0,0 +1,69 @@
...
+        body: {
+          status: "error",
+        },
+      )
+
+      begin
+        api_client.raise_support_ticket(subject: "Ticket without body")
+      rescue GdsApi::HTTPUnprocessableEntity
+        # expected outcome
+      end
```
brucebolt, line 66: Would it be better to assert that the error was raised in this test?

brucebolt (commented):
Looks good to me.  Just left a small comment about one of the tests.

Also, could you add an entry to the 'Unreleased' section of the `CHANGELOG` file, so it's clear these changes exist when someone next releases the gem.

brucebolt (approved):

PR #1286, opened 2024-09-04
Remove redirect type from router api

theseanything:
Router API has dropped support for temporary redirects, as they aren't a used feature.


AgaDufrat (approved):
LGTM

PR #1290, opened 2024-10-30
Add GraphQL endpoint for Publishing API

brucebolt:
Publishing API now has a GraphQL endpoint. Adding this to the gds-api-adapters gem so other applications can make queries to it.

[Trello card](https://trello.com/c/jDF1nSQ0)


test/pacts/publishing_api/graphql_query_pact_test.rb
```diff
@@ -0,0 +1,42 @@
...
+        edition(basePath: "/my-document") {
+          ... on Edition {
+            title
+          }
+        }
+      }
+    QUERY
+
+    publishing_api
+      .given("a published content item exists with content_id: 19ad249e-7ac4-4aa4-8ab4-b6c5f381c043")
```
yndajas, line 21: **suggestion:** pertinent field

Per [my comment on the other PR](https://github.com/alphagov/publishing-api/pull/2950/files#r1822724361), it looks as though the base path is the meaningful field here for the Pact setup

test/test_helpers/publishing_api_test.rb
```diff
@@ -611,6 +611,30 @@
     end
   end
 
+  describe "#stub_publishing_api_graphql_query" do
```
yndajas, line 614: **thought:** testing test methods

Interesting that we're testing these test helper methods. I wonder how much this is just testing the library (webmock?)

test/test_helpers/publishing_api_test.rb
```diff
@@ -611,6 +611,30 @@
...
+    it "returns the given response" do
+      query = "some query"
+
+      stubbed_response = {
+        edition: {
+          title: "some title",
+        },
+      }
```
yndajas, line 622: **suggestion:** realistic response

Should this be nested in a hash under `data`?

yndajas (approved):
Nothing blocking here, just a few thoughts/questions

test/pacts/publishing_api/graphql_query_pact_test.rb
brucebolt: I've raised a PR to update the example and will also update it here.

test/test_helpers/publishing_api_test.rb
brucebolt: I did consider whether this test was needed, but concluded we're making a library of testing methods (with some logic, albeit very simple) available to other apps, so we should test them.

test/test_helpers/publishing_api_test.rb
brucebolt: Yes, it should be.  I'll update that.

PR #1292, opened 2024-11-13
Search API v2: Add adapter for autocomplete

csutter:
This brings #1282 back around now that we have decided not to expose Search API v2's autocomplete endpoint publicly, and need to access it from Finder Frontend.

This adds a method for the new autocomplete functionality to Search API v2's API adapter.

- Add `SearchApiV2#autocomplete`
- Remove redundant error scenario tests to avoid tests bloating up too much

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Add autocomplete endpoint for Search API v2 [PR](https://github.com/alphagov/gds-api-adapters/pull/1292)
+
 ## 97.2.0
 
 *  Add page and order args to get_content_by_embedded_document [PR](https://github.com/alphagov/gds-api-adapters/pull/1291)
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Looks good, though I imagine that this might box in for local dev where the public endpoints aren't accessible. Based on our previous chat about /api/search/autocomplete.json and /api/search.json, we'd find that a local dev could only have one of these. Is this a problem?

Also, only loosely related to this PR but a reminder before I forget. I guess autocomplete could push a lot more traffic through finder-frontend so we may want to up the number of pods.

csutter:
> Looks good, though I imagine that this might box in for local dev where the public endpoints aren't accessible. Based on our previous chat about /api/search/autocomplete.json and /api/search.json, we'd find that a local dev could only have one of these. Is this a problem?
So this should work in local dev if the developer has `search-api-v2` running, which is now less of a pain than it used to be – and if they don't, I reckon autocomplete is far enough off the critical path for local development that it'll be fine (worst outcome is you'll get some fetch errors in the console).


> Also, only loosely related to this PR but a reminder before I forget. I guess autocomplete could push a lot more traffic through finder-frontend so we may want to up the number of pods.
Agree, they're relatively basic requests but might block enough of the capacity that it's worth dialling up in anticipation 📈

PR #1310, opened 2024-12-16
Return GraphQL responses with underscore keys

brucebolt:
Publishing API returns GraphQL responses with camel cased keys. However we use underscored keys in our applications.

Therefore transforming the keys here, so the same does not need to be duplicated in every consuming application.

[Trello card](https://trello.com/c/RMkMvY8b)

### DIFF
--- a/gds-api-adapters.gemspec
+++ b/gds-api-adapters.gemspec
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 3.1"
   s.files        = Dir.glob("lib/**/*") + Dir.glob("test/fixtures/**/*") + %w[README.md Rakefile]
   s.require_path = "lib"
+  s.add_dependency "activesupport"
   s.add_dependency "addressable"
   s.add_dependency "link_header"
   s.add_dependency "null_logger"
--- a/lib/gds_api/publishing_api.rb
+++ b/lib/gds_api/publishing_api.rb
@@ -579,9 +579,11 @@ class GdsApi::PublishingApi < GdsApi::Base
   #
   # @param query [String]
   #
-  # @return [GdsApi::Response] A response with the result of the GraphQL query.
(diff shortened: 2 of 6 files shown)


gds-api-adapters.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 3.1"
   s.files        = Dir.glob("lib/**/*") + Dir.glob("test/fixtures/**/*") + %w[README.md Rakefile]
   s.require_path = "lib"
+  s.add_dependency "activesupport"
```
mike3985, line 18: Are there any users of this gem that don't already have ActiveSupport?

I do get that it's still a bit gross to do this even if that's currently the case, but if it's not the case I'd be a lot less keen on doing this 😆 

Maybe there are standalone gems that implement these methods?

lib/gds_api/publishing_api.rb
```diff
@@ -579,9 +579,11 @@ def get_schema(schema_name)
   #
   # @param query [String]
   #
-  # @return [GdsApi::Response] A response with the result of the GraphQL query.
+  # @return [GdsApi::Response] A response with the result of the GraphQL query whose keys are changed to be underscored.
   def graphql_query(query)
-    post_json("#{endpoint}/graphql", query:)
+    response = post_json("#{endpoint}/graphql", query:)
```
mike3985, line 584: Repeating what I said in directly to Bruce, for the rest of the team: I reckon we could do something with the `JsonClient`'s `create_response` block (https://github.com/alphagov/gds-api-adapters/blob/0ddad8bcd864c0f39a521988c0ff4d5f29cd6301/lib/gds_api/json_client.rb#L93-L94). It'd allow us to be as Publishing API- / GraphQL-specific as we wanted without forcing those details onto the rest of GDS API Adapters.

brucebolt:
Closing in favour of using snake cased keys for fields and arguments in Publishing API: https://github.com/alphagov/publishing-api/pull/3049.

PR #1316, opened 2024-12-19
(783) Add an API adapter for the signon users endpoint

pezholio:
Trello Card: https://trello.com/c/ifMWZl9n/783-add-api-adapters-for-signon

This also has all the necessary setup for Pact testing as well.


JonathanHallam (approved):
LGTM!

PR #1319, opened 2025-01-02
Update pact test to reflect that Publishing API publish endpoint is now idempotent

ryanb-gds:
The Publishing API publish endpoint used to return HTTP 409 Conflict error responses if a publishing app attempted to publish a content item which already had a published edition and no draft edition.

These errors are rarely very useful to the Publishing apps as they are usually caused by a duplicate request due to dependency resolution behaviour or Sidekiq's "at least once" guarantee.

We therefore updated Publishing API to return a 200 OK HTTP response is such cases, as the system is still in a perfectly valid state.

Relevant Publishing API PR: https://github.com/alphagov/publishing-api/pull/3068

### DIFF
--- a/test/pacts/publishing_api/publish_pact_test.rb
+++ b/test/pacts/publishing_api/publish_pact_test.rb
@@ -77,7 +77,7 @@ describe "GdsApi::PublishingApi#publish pact tests" do
     end
   end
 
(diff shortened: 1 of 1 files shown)


ryanb-gds:
Closing this because the relevant PR to Publishing API was also closed due to being a breaking change to an existing endpoint

PR #1343, opened 2025-05-08
Add support for unpublished editions in `graphql_content_item` method

brucebolt:
The `graphql_content_item` method is intended to transform Publishing API's GraphQL response to make it look like a content item returned from Content Store. This is so we can implement GraphQL whilst making as few changes to the frontend applications as possible.

As part of this, we need to handle unpublished editions correctly, by returning the content of the error in the response, so the frontend application can handle it.

[Trello card](https://trello.com/c/oD9MDupC)

### DIFF
--- a/lib/gds_api/publishing_api.rb
+++ b/lib/gds_api/publishing_api.rb
@@ -612,7 +612,14 @@ class GdsApi::PublishingApi < GdsApi::Base
   # @return [GdsApi::Response] A response with the result of the GraphQL query formatted like a Content Store content item.
   def graphql_content_item(query)
     create_response = proc do |r|
-      updated_body = JSON.parse(r.body).dig("data", "edition")
+      parsed_body = JSON.parse(r.body)
+
+      updated_body = if parsed_body["errors"] && (unpublished_error = parsed_body["errors"].find { |error| error["message"] == "Edition has been unpublished" })
+                       unpublished_error["extensions"]
+                     else
+                       parsed_body.dig("data", "edition")
+                     end
+
       updated_response = RestClient::Response.create(
         updated_body.to_json,
         r.net_http_res,
--- a/lib/gds_api/publishing_api.rb
+++ b/lib/gds_api/publishing_api.rb
@@ -615,6 +615,11 @@ class GdsApi::PublishingApi < GdsApi::Base
       parsed_body = JSON.parse(r.body)
 
       updated_body = if parsed_body["errors"] && (unpublished_error = parsed_body["errors"].find { |error| error["message"] == "Edition has been unpublished" })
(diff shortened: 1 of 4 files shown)


test/test_helpers/publishing_api_test.rb
```diff
@@ -665,6 +665,42 @@
...
+      stubbed_response = {
+        data: {
+          edition: nil,
+        },
+        errors: [
+          {
+            message: "Edition has been unpublished",
+            extensions: {
+              "document_type": "gone",
+              "schema_name": "gone,",
```
yndajas: **typo:** you've got a comma in the schema name value

lib/gds_api/publishing_api.rb
```diff
@@ -615,6 +615,8 @@ def graphql_content_item(query)
       parsed_body = JSON.parse(r.body)
 
       updated_body = if parsed_body["errors"] && (unpublished_error = parsed_body["errors"].find { |error| error["message"] == "Edition has been unpublished" })
+                       raise GdsApi::HTTPGone.new(410, nil, unpublished_error["extensions"]) if unpublished_error.dig("extensions", "schema_name") == "gone" && (unpublished_error.dig("extensions", "details").nil? || unpublished_error.dig("extensions", "details").values.compact.reject(&:empty?).empty?)
```
yndajas, line 618: **suggestion (non-blocking):** this is a really long line so it might help to put the `if` first and maybe break the conditions onto multiple lines

yndajas, line 618: **suggestion (non-blocking):** this looks similar to the equivalent `ContentItem` model code from Content Store, but it's not obvious to me what might lead details to be empty or not, and why that affects whether to raise an error. Perhaps worth documenting in the commit message if there's an easy explanation/example?

yndajas (approved):
Approving because the typo is meaningful but inconsequential and the other suggestions are non-blocking

test/test_helpers/publishing_api_test.rb
brucebolt: I'll remove that comma.

lib/gds_api/publishing_api.rb
brucebolt: That's a good idea, I'll make this change.

brucebolt: Looking at the [`GonePresenter` in Publishing API](https://github.com/alphagov/publishing-api/blob/32cd9658825fafa71921477a025d8c587a47443f/app/presenters/gone_presenter.rb#L54-L57), it looks as though the key will always be present with either a value or an empty string.

I'll leave the possibility of `nil` always being present, just in case I've missed something.

PR #1392, opened 2026-04-20
[PP-7588] Update Pact to reflect new base path schema definition

yndajas:
The absolute path definition used by base paths [is being updated](https://github.com/alphagov/publishing-api/pull/4018) to have multiple requirements via allOf. This test needs updating to reflect that change

I've tested the Publishing API branch locally against the Pacts (per [the guidance](https://docs.publishing.service.gov.uk/manual/pact-testing.html#changing-existing-pact-tests)) on this branch and everything passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 103.3.1
+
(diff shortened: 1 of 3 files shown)


AgaDufrat (approved):
LGTM

PR #1401, opened 2026-06-12
[PP-7732] Allow skipping schema validation in patch links

yndajas:
This option is being added to Publishing API to allow patch links requests to skip validation, which should enable removal of old now-invalid links when a document's type changes. More:
- https://github.com/alphagov/publishing-api/pull/4100
- https://github.com/alphagov/publishing-api/commit/8ee0b6a061e98e27c215f0054448538f10677f4d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 103.4.0
+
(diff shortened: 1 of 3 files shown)


brucebolt (approved):
Looks good to me.

Will hold off from merging until the Whitehall team have confirmed they are happy with the change.

yndajas:
Marked as draft while we await clarification from Whitehall on how the current code led to an error

PR #1402, opened 2026-06-12
Update search autocomplete tests

emmalowe:
We no longer raise internal errors in search-api-v2 if GCP times out or itself raises an internal error for an autocomplete request:
https://github.com/alphagov/search-api-v2/pull/677

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2146

### DIFF
--- a/test/search_api_v2_test.rb
+++ b/test/search_api_v2_test.rb
@@ -51,13 +51,8 @@ describe GdsApi::SearchApiV2 do
       stub_request(:get, /example.com\/autocomplete/).to_return(body: "[]")
     end
 
-    it "should raise an exception if the request is unsuccessful" do
-      stub_request(:get, /example.com\/autocomplete.json/).to_return(
-        status: [500, "Internal Server Error"],
-      )
-      assert_raises(GdsApi::HTTPServerError) do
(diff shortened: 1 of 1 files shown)


emmalowe:
@hannako I don't think we got round to deciding whether to move the [Internal Error handling](https://github.com/alphagov/search-api-v2/blob/main/app/controllers/application_controller.rb#L2) in `search-api-v2` from the `ApplicationController` to the `SearchesController` (since it's no longer needed for autocomplete). If we're leaving that as-is, I don't think it makes that much sense to update this test.

hannako (approved):

emmalowe:
> @hannako I don't think we got round to deciding whether to move the [Internal Error handling](https://github.com/alphagov/search-api-v2/blob/main/app/controllers/application_controller.rb#L2) in `search-api-v2` from the `ApplicationController` to the `SearchesController` (since it's no longer needed for autocomplete). If we're leaving that as-is, I don't think it makes that much sense to update this test.

We chatted about this yesterday, and decided to move the error handling in `search-api-v2`. PR for that here:
https://github.com/alphagov/search-api-v2/pull/691

PR #1406, opened 2026-06-29
Add RFC-192 compliant base path validator

KludgeKML:
Implements item 1 of [RFC-192](https://github.com/alphagov/govuk-rfcs/pull/192): "A validation method is added to [gds-api-adapters](https://github.com/alphagov/gds-api-adapters) which tests a base path passed to it against the new restriction. This becomes the source of truth."

Adds the new class GdsApi::Validators::BasePathValidator, which can be used to confirm whether a base path meets the criteria in RFC-192, and provides human-readable errors.

Usage:
```
   > good_validator = GdsApi::Validators::BasePathValidator.new("/government/news/new-validator-rfc-192")
   > good_validator.valid?
   true
   
   > bad_validator = GdsApi::Validators::BasePathValidator.new("/government/news/New-Validator_rfc-192")
   > bad_validator.valid?
   false
  > bad_validator.errors?
  ["Path contains characters that are not lowercase letters, numbers, -, ., or /"]
```

### DIFF
--- a/lib/gds_api/validators/base_path_validator.rb
+++ b/lib/gds_api/validators/base_path_validator.rb
@@ -1,0 +1,45 @@
+module GdsApi
+  module Validators
+    class BasePathValidator
+      MAX_PATH_LENGTH = 511
+
+      attr_reader :base_path
+
+      def initialize(base_path)
+        @base_path = base_path
+      end
+
+      def valid?
+        !(base_path.nil? || too_long? || potential_path_traversal? || ends_with_a_period? || invalid_chars?)
+      end
+
+      def errors
+        errors = []
+        errors << "Path cannot be nil" if base_path.nil?
+        errors << "Path cannot be longer than #{MAX_PATH_LENGTH} characters" if too_long?
+        errors << "Path contains runs of . and or / characters, which could be penetration attempts" if potential_path_traversal?
+        errors << "Path cannot end with a ." if ends_with_a_period?
+        errors << "Path contains characters that are not lowercase letters, numbers, -, ., or /" if invalid_chars?
+        errors
+      end
+
+    private
+
+      def too_long?
+        base_path.length > MAX_PATH_LENGTH
+      end
+
+      def potential_path_traversal?
+        base_path =~ /([\/.]{2,})/
+      end
+
+      def ends_with_a_period?
+        base_path[-1] == "."
+      end
+
+      def invalid_chars?
+        base_path !~ /^\/$|^(\/[a-z0-9.-]+)+$/
+      end
+    end
+  end
+end
--- a/test/validators/base_path_validator_test.rb
+++ b/test/validators/base_path_validator_test.rb
@@ -1,0 +1,79 @@
+require "test_helper"
+require "gds_api/validators/base_path_validator"
+
+describe GdsApi::Validators::BasePathValidator do
+  let(:valid_path_examples) do
+    [
+      "/",
+      "/government/topical-event/heat-wave-2026",
+      "/government/topical-event/heat-wave-2026.csv",
+      "#{'/0123456789' * 46}.json",
+    ]
+  end
+
+  let(:invalid_path_examples) do
+    [
+      "//",
+      "//govuk",
+      "/Government/topical-events",
+      "/government/topical_events",
+      "/government/miss-ca$h",
+      "/government/files,",
+      "/government/files:-more-government/",
+      "/values/../../secret-stuff",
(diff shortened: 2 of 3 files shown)


callumknights (dismissed):
Having had a good read of the RFC beforehand, as it passed me by. I'm happy to approve this 👍

test/validators/base_path_validator_test.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  describe "#errors" do
+    it "returns an empty array for valid paths" do
+      valid_path_examples.each do |base_path|
+        assert_equal([], GdsApi::Validators::BasePathValidator.new(base_path).errors, "#{base_path} should not return errors")
+      end
+    end
+
+    it "returns appropriate errors for valid paths" do
```
leenagupte, line 73: Could a test be returned for the specific type of error? e.g. one for too long, one for ends with a period, etc.

lib/gds_api/validators/base_path_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+        errors << "Path contains characters that are not lowercase letters, numbers, -, ., or /" unless valid_chars?
+        errors
+      end
+
+      def too_long?
+        base_path.length > MAX_PATH_LENGTH
+      end
+
+      def path_traversal?
+        (base_path =~ /([\/.]{2,})/).nil? ? false : true
```
leenagupte: I wonder if this clause should be the other way around? i.e it'd be easier to understand if the clause was checking if two `/` are `present?` rather than `nil`.
It'd mean having to change the logic above to `!path_traversal?`

I'm not convinced by the method name, but to be honest, can't think of a better one!

lib/gds_api/validators/base_path_validator.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def ends_with_a_period?
+        base_path[-1] == "."
+      end
+
+      def valid_chars?
+        (base_path =~ /^\/$|^(\/[a-z0-9.-]+)+$/).nil? ? false : true
```
leenagupte: I wonder if this could be a truthy check too and use `any?` to check if any of the characters are present and return true if they are.

leenagupte (commented):
I've left some inline comments, but their pretty minor.

On the whole this is looking good.

test/validators/base_path_validator_test.rb
KludgeKML: I'm not sure if I'm interpreting this correctly - is this not what it's doing? We have an array of different invalid paths, and we're checking specifically that they're returning the expected error that matches how that path is broken.

lib/gds_api/validators/base_path_validator.rb
KludgeKML: I've renamed this function invalid_chars? and made it use a != regex (returns false if the regex matches, true if it doesn't match, ie invalid characters are present). Simplifies the call slightly and means all the private methods are bad if they return true.

lib/gds_api/validators/base_path_validator.rb
KludgeKML: I've made it even more explicit `potential_path_traversal`, but also private so it shouldn't leak into other codebases accidentally. And I've taken out the nil test - we can't test present? because it isn't always available on a `nil`, but since we're not exposing this to other methods (it's now present) we don't need to be quite so picky that it's exactly true/false - truthy will do.

test/validators/base_path_validator_test.rb
leenagupte: Ah. I think I missed where / how that was happening. I'll take another look.

leenagupte: Sorry, I think I didn't stitch `invalid_path_examples` and `invalid_path_examples_errors` together when I first reviewed this.

lib/gds_api/validators/base_path_validator.rb
```diff
@@ -0,0 +1,45 @@
...
+      MAX_PATH_LENGTH = 511
+
+      attr_reader :base_path
+
+      def initialize(base_path)
+        @base_path = base_path
+      end
+
+      def valid?
+        !(base_path.nil? || too_long? || potential_path_traversal? || ends_with_a_period? || invalid_chars?)
```
leenagupte, line 13: The logic makes more sense. Thanks for updating.

leenagupte (approved):
Thanks for making the updates. The logic follows a bit better for me now. 👍

PR #1415, opened 2026-08-18
Add additional test helpers for asset manager

callumknights:
These are being added to support development of asset management endpoints in hmrc-manuals-api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 
-## Unreleased
+## 103.4.1
 
-* Add extra test helpers to Asset Manager ([PR]https://github.com/alphagov/gds-api-adapters/pull/1415)
+* Add extra test helpers to Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1415))
 
 ## 103.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Add extra test helpers to Asset Manager ([PR]https://github.com/alphagov/gds-api-adapters/pull/1415)
+
 ## 103.4.0
 
 * Improve RFC-192 validator method ([PR](https://github.com/alphagov/gds-api-adapters/pull/1411))
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -45,6 +45,23 @@ module GdsApi
           .to_return(body: response.to_json, status: 404)
       end
 
+      def stub_asset_manager_responds_forbidden(id)
+        response = {
+          "_response_info" => { "status" => "Forbidden. You don't have permission to access this resource." },
+        }
+
+        stub_request(:any, "#{ASSET_MANAGER_ENDPOINT}/assets/#{id}")
+          .to_return(body: response.to_json, status: 403)
+      end
+
+      def stub_asset_manager_responds_payload_too_large
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 413)
+      end
+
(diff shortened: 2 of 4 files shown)


lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -45,6 +45,27 @@ def stub_asset_manager_does_not_have_an_asset(id)
...
+      end
+
+      def stub_asset_manager_responds_unprocessable
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 422)
+      end
+
+      def stub_asset_manager_responds_not_acceptable
+        stub_request(:any, %r{\A#{ASSET_MANAGER_ENDPOINT}}).to_return(status: 406)
+      end
+
```
brucebolt: Could you add some tests for these new methods.

brucebolt (commented):
The test helpers themselves look good.

Can you also add an entry into the `CHANGELOG` file for this.

lib/gds_api/test_helpers/asset_manager.rb
callumknights: Done, I was clearly tired. Didn't notice the tests for the asset-manager helpers at all 🤦

callumknights:
For the `CHANGELOG`, is that the correct release process for this repo? Looking at the git history it seems to be encouraged to raise a separate PR with `CHANGELOG` additions as the same time as releasing a new version, and include all currently unreleased PR's that change?

brucebolt:
Normally we add an entry to the CHANGELOG under an 'Unreleased' heading in the same commit as the change that introduces it.

Then we either release a new version of the gem [in another commit](https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html#conventions) in the same PR (if wanting to release immediately) or in another PR when we want to group changes together.

callumknights:
Okay, thanks for clarifying. I'll add an `unreleased` in the commit, then do another commit for release as it would be good to get these straight out there and remove the remote gem path from my hmrc-manuals-api PR

test/test_helpers/asset_manager_test.rb
```diff
@@ -8,6 +8,7 @@
   let(:stub_asset_manager) do
     GdsApi::AssetManager.new(Plek.find("asset-manager"))
   end
+  let(:file_fixture) { load_fixture_file("hello.txt") }
```
brucebolt: Is there a reason these methods need a file?  The stub is for any request to Asset Manager, so you could just call the `asset` method instead of `update_asset`, then don't need to consider the asset at all in these tests.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 103.4.1
+
+* Add extra test helpers to Asset Manager ([PR]https://github.com/alphagov/gds-api-adapters/pull/1415)
```
brucebolt:
```suggestion
* Add extra test helpers to Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1415)
```

test/test_helpers/asset_manager_test.rb
callumknights: ~I suppose that would be fine, was stuck in the headspace of testing `update`, but you're right, this is more lightweight.~ I knew there was a reason I thought this was right. Both of these errors are (for our HMRC endpoint) for endpoints which expect a file (as well as the regenerate endpoint). However given your point that the stub is for any request, it's probably still right to make these nice and lightweight.

callumknights: Have updated 👍

CHANGELOG.md
callumknights: Sorted 👍

callumknights:
Superseded by https://github.com/alphagov/gds-api-adapters/pull/1416

PR #1416, opened 2026-08-24
[PP-7954] Add and update test helpers for Asset Manager

brucebolt:
Whilst implementing asset management in HMRC Manuals API, it was found that some new and updated test helpers have been required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 103.4.1
 
 * Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
   * `stub_asset_manager_create_asset_too_large`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   * `stub_asset_manager_create_asset_unprocessable`
   * `stub_asset_manager_request_to_asset_media`
   * `stub_asset_manager_request_to_get_asset`
+  * `stub_asset_manager_update_asset_forbidden`
+  * `stub_asset_manager_update_asset_not_found`
 * Rename `stub_asset_manager_receives_an_asset` test helper to `stub_asset_manager_create_asset`; the old name will be deprecated in a later release ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
 
 ## 103.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
   * `stub_asset_manager_create_asset_too_large`
+  * `stub_asset_manager_create_asset_unprocessable`
   * `stub_asset_manager_request_to_asset_media`
   * `stub_asset_manager_request_to_get_asset`
 * Rename `stub_asset_manager_receives_an_asset` test helper to `stub_asset_manager_create_asset`; the old name will be deprecated in a later release ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+* Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
+  * `stub_asset_manager_request_to_asset_media`
+  * `stub_asset_manager_request_to_get_asset`
+
 ## 103.4.0
 
 * Improve RFC-192 validator method ([PR](https://github.com/alphagov/gds-api-adapters/pull/1411))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   * `stub_asset_manager_update_asset_forbidden`
   * `stub_asset_manager_update_asset_not_found`
 * Rename `stub_asset_manager_receives_an_asset` test helper to `stub_asset_manager_create_asset`; the old name will be deprecated in a later release ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
+* Allow `stub_asset_manager_create_asset` test helper to return a body ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
 
 ## 103.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
   * `stub_asset_manager_request_to_asset_media`
   * `stub_asset_manager_request_to_get_asset`
+* Rename `stub_asset_manager_receives_an_asset` test helper to `stub_asset_manager_create_asset`; the old name will be deprecated in a later release ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
 
 ## 103.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
   * `stub_asset_manager_create_asset_too_large`
   * `stub_asset_manager_create_asset_unprocessable`
+  * `stub_asset_manager_get_asset_forbidden`
   * `stub_asset_manager_request_to_asset_media`
   * `stub_asset_manager_request_to_get_asset`
   * `stub_asset_manager_update_asset_forbidden`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
+  * `stub_asset_manager_create_asset_too_large`
   * `stub_asset_manager_request_to_asset_media`
   * `stub_asset_manager_request_to_get_asset`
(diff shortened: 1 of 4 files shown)


test/test_helpers/asset_manager_test.rb
```diff
@@ -99,4 +99,59 @@
...
+        stub_asset_manager_has_an_asset(asset_id, body)
+        response = stub_asset_manager.asset(asset_id)
+
+        assert_equal 200, response.code
+        assert_equal body[:key], response["key"]
+      end
+    end
+
+    describe "when the media endpoint is requested" do
+      it "returns the given attributes" do
```
yndajas: **question:** should this be "returns the file content"?

lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -85,6 +85,10 @@ def stub_asset_manager_receives_an_asset_too_large
         stub_request(:post, "#{ASSET_MANAGER_ENDPOINT}/assets").to_return(status: 413)
       end
 
+      def stub_asset_manager_receives_an_unprocessable_asset(body: "")
+        stub_request(:post, "#{ASSET_MANAGER_ENDPOINT}/assets").to_return(body:, status: 422)
```
yndajas, line 103: **question:** what's the use case for the body here?

lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -98,6 +98,14 @@ def stub_asset_manager_update_asset(asset_id, body = {})
           .to_return(body: body.to_json, status: 200)
       end
 
+      def stub_asset_manager_update_asset_forbidden(asset_id)
```
yndajas, line 115: **thought:** these update stubs make it clear that they're stubbing a PUT/update request, whereas the POST/create ones don't - might it be useful to make it clear on all the stubs? I'm not sure that create is an obvious enough default

test/test_helpers/asset_manager_test.rb
```diff
@@ -178,4 +178,30 @@
       end
     end
   end
+
+  describe "#stub_asset_manager_update_asset_forbidden" do
+    describe "when the endpoint is requested" do
```
yndajas: **suggestion:** "when the endpoint is requested _with the given ID_"?

lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -52,6 +52,10 @@ def stub_asset_manager_does_not_have_an_asset(id)
           .to_return(body: response.to_json, status: 404)
       end
 
+      def stub_asset_manager_asset_forbidden(id)
```
yndajas: **suggestion:** might be worth explaining the use case for a generic 403 stub after having created a more specific update/PUT one

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Addition of new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
```
yndajas: **nitpick:** I think the changelog entries typically begin with a verb

test/test_helpers/asset_manager_test.rb
brucebolt: Yes, it should be.  Will update.

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: Asset Manager returns an error message in the 422 response.  Have added this as we pass that error through to the client in HMRC Manuals API.

test/test_helpers/asset_manager_test.rb
brucebolt: Good addition.  I'll add this to add the relevant labels.

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: I've renamed it to `stub_asset_manager_get_asset_forbidden`, as this deals specifically with the endpoint to get an asset's metadata.

CHANGELOG.md
brucebolt: Yes, I'll update this.

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: I tried to keep the POST endpoints consistent with the existing helper `stub_asset_manager_receives_an_asset` and `stub_asset_manager_upload_failure` (which aren't ideally named, but would probably be a fair effort to rename across all usages now).

Maybe I could rename the new ones to start `stub_asset_manager_create_...`?

yndajas: Yeah I guess they're used beyond HMRC Manuals API. We could rename the original method then create an alias with the old name, or a full method under the original name with a deprecation warning. But probably not a now thing

```rb
def stub_asset_manager_create_asset
  # original implementation
end

def stub_asset_manager_receives_an_asset
  puts %(stub_asset_manager_receives_an_asset is deprecated and
will be removed in a future version of gds-api-adapters.
Replace calls to this method with stub_asset_manager_create_asset
)

  stub_asset_manager_create_asset
end
```

lib/gds_api/test_helpers/asset_manager.rb
yndajas: Should there be a test to check that it gets passed through to the response?

lib/gds_api/test_helpers/asset_manager.rb
yndajas: Should the method be `:get` rather than `:any` in that case?

callumknights:
Looking at the increased specificity of these helpers, this probably supersedes https://github.com/alphagov/gds-api-adapters/pull/1415, I'll close that if there are no objections

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: That's a really nice solution.  I've added this and a test for the deprecation.

Also added this to the CHANGELOG.

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: Ah, I had't spotted that.  Have updated it to `:get` now.

brucebolt:
Thanks for the review @yndajas.  I've addressed all the comments and also updated the CHANGELOG so it is added to incrementally in each commit.

brucebolt:
> Looking at the increased specificity of these helpers, this probably supersedes #1415, I'll close that if there are no objections

You may want to check if there's any extra helpers you'll need that aren't covered by this PR, e.g. `stub_asset_manager_create_asset_too_large` only stubs creating an asset, not updating an existing one.

callumknights:
Yeah, I was previously using the old methods pattern of just stubbing `:any` type of request, for the error cases at least, so I'll need to drop in extra helpers for each request type. I'll close that PR and raise another.

lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -81,6 +81,12 @@ def stub_asset_manager_receives_an_asset(response_url = {})
         end
       end
 
+      def stub_asset_manager_receives_an_asset(response_url = {})
+        warn "stub_asset_manager_receives_an_asset is deprecated and will be removed in a future version of gds-api-adapters. Replace calls to this method with stub_asset_manager_create_asset."
```
yndajas: **suggestion:** might be good to multi-line this given it'll probably be rendered in a terminal

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Add new test helpers for Asset Manager ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
   * `stub_asset_manager_request_to_asset_media`
   * `stub_asset_manager_request_to_get_asset`
+* Rename `stub_asset_manager_receives_an_asset` test helper to `stub_asset_manager_create_asset`; the old name will be deprecated in a later release ([PR](https://github.com/alphagov/gds-api-adapters/pull/1416))
```
yndajas, line 13: **thought:** do deprecations normally get a "DEPRECATED" at the start of their changelog entry?

lib/gds_api/test_helpers/asset_manager.rb
yndajas: I think this comment is still outstanding, but minor!

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: I initially had problems with indentation being output, but then remember to use a squiggly heredoc which solves that problem.  Will update.

CHANGELOG.md
brucebolt: I didn't add it this time, as we're not actually deprecating anything here. As a patch release, this update will likely be merged into apps automatically without a human reading this note anyway.

lib/gds_api/test_helpers/asset_manager.rb
brucebolt: I must've missed it.  Added a test now.

CHANGELOG.md
callumknights: This seems more like a deprecation warning to me rather than the thing itself. Soft deprecation maybe? Either way I agree with Bruce

yndajas: Oh that's interesting. My understanding is that deprecation is the stage before removal. E.g. you might see: "WARNING: some_method is deprecated and will be removed in a future version." It seems that it's used inconsistently though, e.g. [Swagger seem to talk about it as the removal stage](https://swagger.io/blog/best-practices-for-deprecating-apis), while other sources seem to suggest it's the warning stage:
- https://www.columbia.edu/cu/help/jdk/docs/guide/misc/deprecation/deprecation.html
- https://daj.medium.com/deprecate-the-most-misused-word-in-software-engineering-ef2d033fa9b0

yndajas (approved):

PR #1417, opened 2026-08-25
[PP-7892] Add restore asset test helpers

yndajas:
These will be used in HMRC Manuals API to stub successful and unsuccessful Asset Manager responses so that we can test the former deals with such responses properly


callumknights:
Is it worth breaking this into individual helper commits, with accompanying `CHANGELOG` additions, similar to #1416 and #1418

yndajas:
> Is it worth breaking this into individual helper commits, with accompanying `CHANGELOG` additions, similar to #1416 and #1418

Good shout - done!

callumknights (approved):

PR #1418, opened 2026-08-25
Asset manager test helpers pt2

callumknights:
Similar to https://github.com/alphagov/gds-api-adapters/pull/1416, some more test helpers for use with the new HMRC Manuals API endpoints

### DIFF
--- a/lib/gds_api/test_helpers/asset_manager.rb
+++ b/lib/gds_api/test_helpers/asset_manager.rb
@@ -120,6 +120,10 @@ module GdsApi
         stub_request(:put, "#{ASSET_MANAGER_ENDPOINT}/assets/#{asset_id}").to_return(status: 404)
       end
 
+      def stub_asset_manager_update_asset_too_large(asset_id)
+        stub_request(:put, "#{ASSET_MANAGER_ENDPOINT}/assets/#{asset_id}").to_return(status: 413)
+      end
+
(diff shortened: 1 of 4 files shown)


lib/gds_api/test_helpers/asset_manager.rb
```diff
@@ -116,6 +116,10 @@ def stub_asset_manager_update_asset_forbidden(asset_id)
         stub_request(:put, "#{ASSET_MANAGER_ENDPOINT}/assets/#{asset_id}").to_return(status: 403)
       end
 
+      def stub_asset_manager_update_asset_too_large(asset_id)
+        stub_request(:put, "#{ASSET_MANAGER_ENDPOINT}/assets/#{asset_id}").to_return(status: 413)
+      end
```
yndajas, line 125: **suggestion:** shall we keep the methods for any given endpoint in response code order?

yndajas (commented):
Just one nit, otherwise good to go

lib/gds_api/test_helpers/asset_manager.rb
callumknights: I honestly hadn't even noticed they were in response code order 🙈

callumknights: All sorted 👍

yndajas (approved):

PR #1419, opened 2026-08-26
[PP-7892] Remove restore conflict helper

yndajas:
Asset Manager doesn't ever raise a conflict error on this path. There was a question about this in the draft HMRC Manuals API docs, but I can't evidence of it being looked into: https://github.com/alphagov/hmrc-manuals-api/pull/1291/changes#r3371782630


callumknights (approved):
Nice find! 👍

PR #1421, opened 2026-09-08
Allow nil base_paths in validator

KludgeKML:
- Nil as a base_path corresponds to items that will not be published to content store, but should still be allowed in publishing-api. This means that we can either maintain a complicated list of elements in publishing api that shouldn't be validated, or we can remove the nil check in the validator. We've chosen the second, since it's easy to check nils externally to the validator in  situations where they absolutely must be avoided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


KludgeKML:
Related to https://github.com/alphagov/publishing-api/pull/4182

brucebolt (approved):

PR #1423, opened 2026-09-15
[PP-7716] Pact v2 compatibility

geek4good:
These are some necessary changes, because publishing-api is migrating its provider verification from pact-ruby v1 to v2, whose verifier has slight semantic differences. This has led to 17 interactions failing for the following reasons:

1. **Regex's match full strings in v2.** Pact v1 matched partially, so `matcher: /\S+/` accepted any message containing a non-whitespace character. The v2 matcher checks the whole string. This means `\S+` can never match a message containing spaces (e.g. "A lock-version conflict occurred."). Single-word message message still pass, but that's rather by coincidence than by design.
2. **The v2 verifier needs a declared Content-Type to parse an expected body as JSON.** In v2, a body without headers is treated as `text/plain` by default. So setting the `Content-Type => "application/json"` is now necessary for the verifier to correctly treat the response body as such.
3. **The Pact v2 verifier's HTTP client sends query params in alphabetical order.** While Pact v1's replay preserved the recorded order, v2 now orders the params alphabetically.

These changes make the provider work with Pact v1 and v2 verifiers.


callumknights (approved):
Thanks for this Lucas! I'm pleasantly surprised at the interoperability between Pact V1 and V2. I was expecting we would have to namespace gds-api-adapters pacts and gradually re-write to get the apps with interactions defined here to pass. This is much more lightweight. 👍

